PR #113, opened 2020-10-06
Show new node.js version

simpsoka:
new node.js version


zeke (approved):
I actually just added this file to the checklist in https://github.com/github/docs-internal/blob/main/contributing/node-versions.md but somehow skipped over it.

Thank you! 👍🏼

PR #117, opened 2020-10-06
update scripts/README.md

zeke:
This PR brings `script/README.md` up to date by running `script/update-readme.js`


JasonEtco (approved):
Looks good to me! I don't think CI will fail anymore, I merged in `main`. There are some scripts that don't have notes - I think that's okay, but an area we can improve later on.

janiceilene (approved):
👍

PR #184, opened 2020-10-07
Use the latest triage project board action

rachmari:
This update uses a version of the action that allows adding both issues and pull requests to the project board. This aligns with the action version used in the other triaging workflows.


zeke (approved):

haaami01 (commented):
let's hope it works with this

PR #215, opened 2020-10-07
correct the version of node required

jeffmcaffer:
### Why:
stated version of node was 14 when it's actually 12

### What's being changed:

fix the doc

### Check off the following:


zeke (approved):
Thanks for the PR. 👍🏼 

I made some additional changes to clarify exactly which Node.js versions are required or used in each environment.

contributing/node-versions.md
```diff
@@ -1,6 +1,8 @@
 # Node Versions
 
-The site currently runs on Node.js v14, the [Active LTS version](https://nodejs.org/en/about/releases/) from 2020-10-27 to 2021-10-26.
+In [development](contributing/development.md) enviroments this site will run on Node.js versions `12 - 14`.
+
+In [staging and production](contributing/deployments.md) environments this site runs on Node.js 14, the [Active LTS version](https://nodejs.org/en/about/releases/) from 2020-10-27 to 2021-10-26).
```
jeffmcaffer, line 5: it's a bit odd to say that Node 14 is the Active LTS when it's not. Perhaps just remove the qualification and say "this site runs on the latest in the Node.js 14 family"  (or Node.js 14.* as undoubtedly there will be a gap between a new 14.* coming out and us deploying).

contributing/development.md
```diff
@@ -8,7 +8,7 @@ This site is powered by Node.js! :sparkles: :turtle: :rocket: :sparkles:
 
 It runs on macOS, Windows, and Linux environments.
 
-You'll need **Node.js v14** to run the site. If you're using [`nodenv`](https://github.com/nodenv/nodenv), read the [`nodenv` docs](#nodenv) for instructions on switching Node.js versions. If you're not using `nodenv`, the best way to install Node.js is to [download the LTS installer from nodejs.org](https://nodejs.org).
+You'll need Node.js version `12 - 14` to run the site. To install Node.js, [download the "LTS" installer from nodejs.org](https://nodejs.org). If you're using [`nodenv`](https://github.com/nodenv/nodenv), read the [`nodenv` docs](#nodenv) for instructions on switching Node.js versions.
```
jeffmcaffer, line 11: consider using non-technical terms to talk about the versions. For example, for people not familiar with the Node versioning scheme, `12 - 14` could look like an actual version number. Whereas what I think you mean is "any stable version from 12 on".

contributing/node-versions.md
zeke: It's not the Active LTS now, but it will be in 3 weeks. Hence the `from 2020-10-27 to 2021-10-26.`

contributing/development.md
zeke: I chose the literal `12 - 14` string because it's explicit. I can change that to "Node.js versions 12 to 14`

We don't actually mean "any stable version from 12 on". If anyone tries to run the site on something greater than Node 14, it won't run.

zeke:
@jeffmcaffer I tweaked the wording based on your suggestions. Gonna ship this when it turns green.

Abhishek577270 (commented):
> 218<

PR #225, opened 2020-10-08
remove all-contributors suggestion from merge notification

### LINKED ISSUE #218: Random error in PR bots
Seems like there is a misconfiguration somewhere between the `github-actions` bot and the `allcontributors` bot 

In https://github.com/github/docs/pull/204 we see the github bot asking the allcontributors to add me as a docs contributor and then allcontributors not understanding what to do.

### LINKED ISSUE #220: automated "Thanks very much for contributing!" comment is confusing @all-contributors bot
The bots are confusing one another!  🤖 ⚡  🤖 

Screen Shot 2020-10-07 at 5 07 19 PM

cc @github/docs-engineering

zeke:
This PR removes the bit about using the all-contributors bot from the "you merged a PR" workflow.

Resolves #218
Resolves #220

cc @github/docs-engineering

### DIFF
--- a/.github/workflows/merged-notification.yml
+++ b/.github/workflows/merged-notification.yml
@@ -13,5 +13,7 @@ jobs:
             github.issues.createComment({
               ...context.repo,
               issue_number: context.payload.pull_request.number,
-              body: `Thanks very much for contributing! Your pull request has been merged 🎉 You should see your changes appear on the site in approximately 24 hours.`
+              body: `Thanks very much for contributing! Your pull request has been merged 🎉 You should see your changes appear on the site in approximately 24 hours.
+              
+              If you haven't already, you can add yourself to [the list of contributors](https://github.com/github/docs#contributors-) by following [these instructions](https://allcontributors.org/docs/en/bot/usage#commands). Thanks again! :sparkles:`
             })
--- a/.github/workflows/merged-notification.yml
+++ b/.github/workflows/merged-notification.yml
@@ -15,5 +15,5 @@ jobs:
               issue_number: context.payload.pull_request.number,
               body: `Thanks very much for contributing! Your pull request has been merged 🎉 You should see your changes appear on the site in approximately 24 hours.
               
-              If you haven't already, you can add yourself to [the list of contributors](https://github.com/github/docs#contributors-) by following [these instructions](https://allcontributors.org/docs/en/bot/usage#commands). Thanks again! :sparkles:`
+              If you haven't already, you can add yourself to [the list of contributors](https://github.com/github/docs#contributors-) by creating a new comment in this PR using [these instructions](https://allcontributors.org/docs/en/bot/usage#commands). Thanks again! :sparkles:`
             })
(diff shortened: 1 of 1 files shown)


JasonEtco (approved):
Thanks for fixing this @zeke 👏

.github/workflows/merged-notification.yml
```diff
@@ -13,11 +13,7 @@ jobs:
...
-              body: `Thanks very much for contributing! Your pull request has been merged 🎉 You should see your changes appear on the site in approximately 24 hours. To add your ✨ contribution to the README.md, create a new comment in this PR with:
-            
-            \`\`\`
-            @all-contributors please add @${context.payload.pull_request.user.login} for docs
-            \`\`\`
-            
-            If you want to, you can use the [emoji key](https://allcontributors.org/docs/en/emoji-key) to replace docs with a different contribution type.`
+              body: `Thanks very much for contributing! Your pull request has been merged 🎉 You should see your changes appear on the site in approximately 24 hours.
+              
+              If you haven't already, you can add yourself to [the list of contributors](https://github.com/github/docs#contributors-) by following [these instructions](https://allcontributors.org/docs/en/bot/usage#commands). Thanks again! :sparkles:`
```
janiceilene, line 18:
```suggestion
              If you haven't already, you can add yourself to [the list of contributors](https://github.com/github/docs#contributors-) by creating a new comment in this PR using [these instructions](https://allcontributors.org/docs/en/bot/usage#commands). Thanks again! :sparkles:`
```

janiceilene (approved):
🎉

jeffmcaffer:
@JasonEtco Thanks for the fix. I do wonder about removing the instructions though. Seems like there's a fair bit of friction without it. Users will have to follow the link, figure out the syntax, pick an emoji (do we care what they pick?) and then add the comment. They'll also need to remember to do that for each different type of contribution.

Is it possible to do one of
* automatically add the required comment
* keep something like the previous text but quote the bot command somehow so the allcontributors bot does not try to interpret it

JasonEtco:
Thanks for that feedback @jeffmcaffer!

> automatically add the required comment

We can make a best guess, but this gets a little nuanced when we want to tag the right kind of contribution (`code`, `docs`, `content`, `design`, etc). So automatically adding it would still require some manual effort, _or_ some up front work to make that guessing smarter based on changed files.

> keep something like the previous text but quote the bot command somehow so the allcontributors bot does not try to interpret it

That's what we tried to do, but it seems like the bot is too aggressive in its comment parsing 😅 any comment that mentioned the bot is parsed, and if it fails the bot comments anyway. If we were to write a comment without mentioning the bot, it wouldn't be a copy-pastable example. I'm interested in @janiceilene's thoughts here too - happy to keep experimenting with this!

jeffmcaffer:
I did a quick look at the project (https://github.com/all-contributors/all-contributors) and didn't see anything detailing how to include instructions in a PR (i.e., quoting). I wonder if they've encountered this scenario? (didn't see anything obvious in the issues).

Question, does GitHub have a way of quoting at mentions in its markup? Like if I want to say "mention me, @jeffmcaffer" in a comment without actually mentioning me, how would I do that?

PR #328, opened 2020-10-09
Remove the mention of HOME environment variable

dsame:
### Why:

The variable name clashes with unix shell HOME variable and confuses the users. The related issue is https://github.com/actions/virtual-environments/issues/1301

The variable is not set for all hosted images and has no means in the moment

### What's being changed:

I removed the mention of the variable from the documentation

### Check off the following:


al-cheb:
I think you should provide more details, because if i use the bash shell on Windows, $HOME variable is available.

maxim-lobanov:
cc: @alepauly @hross for review

A bit more context:
- Initial issue was "why we set this variable on Ubuntu and MacOS but not set on Windows system"
- This variable is not actually set by GitHub. It is set by Ubuntu and MacOS system. Since we don't set it from our side, it doesn't make much sense to describe it in doc because it can confuse. I could be wrong of course

thboop:
Windows has a home equivalent too (HOMEPATH), but we don't document that. Removing this makes sense, documenting all of the OS specific env variables is a duplication of effort/information and something that I'm not sure provides a lot of value to a user)

thboop (approved):

maxim-lobanov (approved):

content/actions/reference/environment-variables.md
```diff
@@ -39,7 +39,6 @@ We strongly recommend that actions use environment variables to access the files
 | Environment variable | Description |
 | ---------------------|------------ |
 | `CI` | Always set to `true`. |
```
rachmari, line 41: We've got another similar PR where someone is suggesting to add the Windows-specific variable rather than removing the `HOME` variable. Curious what your thoughts are there?

I can see removing the OS-specific variables, but it might be helpful as we do describe that the directory is where user data is stored.

```suggestion
| `CI` | Always set to `true`. |
| `HOME` | (On Linux and macOS runners) The path to the {% data variables.product.prodname_dotcom %} home directory used to store user data. For example, `/github/home`. |
| `HOMEPATH` | (On Windows runners) The path to the {% data variables.product.prodname_dotcom %} home directory used to store user data. For example, `c:\Users\runneradmin`. |
```

rachmari (commented):
@dsame I just wanted to raise a related change that was created in another pull request. Curious what your thoughts are there?

content/actions/reference/environment-variables.md
dsame: My strongly opinion is we must not mess user home dir related and OS specific environment variable with the github specific variable. 

In fact "The path to the GitHub home directory used to store user data" has nothing to do with "home dir to store user settings"

As a result, the first step is to remove the mentioning of UNIX-specific environment variable HOME and the second step is to add something like "GITHUB_HOME" or "GITHUB_DATA" or "GITGHUB_WORKDIR" which might point to the same directory as HOME or to another.

maxim-lobanov: Personally, I agree with Sergey and @thboop mentioned the similar point https://github.com/github/docs/pull/328#issuecomment-707112891
We should not mess variables that are set by GitHub and variables are set by system. At least, because there are a ton of other system variables that we don't want to describe here.

> but it might be helpful as we do describe that the directory is where user data is stored.

Actually, it is already described `GITHUB_WORKSPACE`. 
```
The GitHub workspace directory path. The workspace directory is a copy of your repository if your workflow uses the actions/checkout action. If you don't use the actions/checkout action, the directory will be empty. For example, /home/runner/work/my-repo-name/my-repo-name.
```
I guess @dsame meant the same under `GITHUB_HOME`

dsame: @rachmari with the GITHUB_WORKSPACE already there's no sense to mention HOME variable at all. It works for windows users but confuses *nix guys because of messing with shell HOME variable.

rachmari: Sounds good. I'll close out the other PR that is suggesting to add the `HOMEPATH` variable and I'll get this update shipped.

rachmari (approved):
LGTM! 🚢 I'll get this merged right away.

w-bonelli:
Hi, although this PR removed it from the default variables table, I noticed the docs [still describe `$HOME` as a default environment variable](https://github.com/github/docs/blob/main/content/actions/learn-github-actions/environment-variables.md#naming-conventions-for-environment-variables) in the naming conventions section, which initially gave me the impression that it will be set by the runner on all platforms. Maybe it is worth removing mention of `$HOME` from the naming conventions section as well, or clarifying that it is Unix-specific? Maybe I was the only one confused by this :)

cmwilson21:
@w-bonelli - Thanks for dropping in and leaving this comment. Would you mind turning it into a new issue? That way we can triage it for our wonderful writers to check into.

Thanks! 💖

PR #340, opened 2020-10-09
It can be misunderstood in human language.

crucifyer:
It can be misunderstood as meaning human language, so I think it would be good to clarify.
I want to prevent non-technical topics from appearing on the topics/featured page.
thank you.

### Why:
The reason for the modification is not to mean that the code is related to Korean,
This is because it is intended to be used to inform the person who wrote the code that they can speak Korean.

### Check off the following:


janiceilene:
👋 @crucifyer Thanks for opening this pull request! Since you can add both programming languages and spoken languages to the topics, we should keep this phrasing general.

image

PR #362, opened 2020-10-09
use ed25519 keys instead of rsa keys

timball:
RSA keys are outdated. This PR updates the document to suggest using ed25519 which, at this time, defaults to a 16 round key deviation function. It may be useful to suggest more rounds for the KDF, but that's for a different PR. 

--timball

### Why:
This Change is to encourage people to use a more sound cryptographic algorithm to generate their ssh keys. 

Currently, the docs recommend 'ssh-keygen' and the RSA algorithm with 4096 bit keys. This is better than the `ssh-keygen` default of 3072 bit keys. [Also, this paper suggests that about 1 in thousand RSA moduli were not cryptographically sound](https://eprint.iacr.org/2012/064.pdf). Last but not least, the RSA algorithm at 1024 bits (or less) is considered unsafe by NIST. We should not encourage the use of [an old algorithm](https://en.wikipedia.org/wiki/RSA_(cryptosystem)) that is often used in an unsafe way.

This change proposes that Github recommend using the more modern, and safer, ed25519 key algorithm rather than RSA. 

### What's being changed:
The original ssh-keygen flags `-t rsa -b 4096` specifying the use of the rsa algorithm with a key 4096 bits long are replaced with `-t ed25519` which specifies using the ed25519 algorithm. 

### Check off the following:


janiceilene:
Thanks for opening a PR 👍 Please fill out the pull request template and we'll get this up for review!

janiceilene:
👋 @timball The pull request template is the information in the original comment at the top of this pull request (the `Why?` and `What's changed?`). You can see all of the text in the [pull-request.md](https://raw.githubusercontent.com/github/docs/main/.github/PULL_REQUEST_TEMPLATE.md). It helps us to triage and choose the right reviewer for each PR that's opened. Thanks!

timball:
> Thanks for opening a PR 👍 Please fill out the pull request template and we'll get this up for review!

Okay! filled it out and editted the request template! 

--timball

janiceilene:
Thanks @timball I'll get this triaged for review!

hubwriter:
@timball - Many thanks. The change looks like a good one. I just want to run this by our security folks here at GitHub then I'll merge this. The same thing crops up elsewhere in the docs, so we'll get those changed too. Thanks again. 👍

ptoomey3:
The reason for the docs as they are is because, at least at the time they were last updated, having the default docs use `ed25519` would have resulted in a non-trivial number of customers receiving an error message, since `ed25519` was new enough that a reasonably large portion of our customers were not using an operating system crypto lib that supported them by default. Do we know the stats on that today? If there is still a long-tail of folks that can't use `ed25519`, I wouldn't be opposed to updating the defaults and adding a "if that doesn't work..use RSA 4096 like this ..." section.

ptoomey3:
Ha..I just checked the repo for docs before we open-sourced it and I had already done dug around a bit and assessed that it was 👍 to update the docs to ed25519. The issue just hadn't gotten prioritized to make the actual docs change yet. 👍 from me.

content/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
```diff
@@ -19,7 +19,7 @@ If you don't want to reenter your passphrase every time you use your SSH key, yo
 {% data reusables.command_line.open_the_multi_os_terminal %}
 2. Paste the text below, substituting in your {% data variables.product.product_name %} email address.
   ```shell
-  $ ssh-keygen -t rsa -b 4096 -C "<em>your_email@example.com</em>"
+  $ ssh-keygen -t ed25519 -C "<em>your_email@example.com</em>"
   ```
```
hubwriter, line 23: I've been advised that this command won't be supported for some GitHub users. I think we should give them the RSA command as a fallback. I suggest adding the following note immediately after the ed25519 command, indented as part of this step of the procedure:

<pre>
  {% note %}

  **Note:** If you are using a legacy system that does not support the Ed25519 algorithm, use:
  ```shell
   $ ssh-keygen -t rsa -b 4096 -C "<em>your_email@example.com</em>"
  ```

  {% endnote %}

</pre>

hubwriter:
I've raised an issue to get another couple of topics updated likewise: https://github.com/github/docs/issues/876

content/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
timball: having a fallback is a great idea ... totally support this. The goal would be not to rely on defaults. 

--timball

hubwriter (approved):
@timball - Thanks. As you agreed to my suggested change (adding the fallback note), I've gone ahead and changed this PR accordingly so that it can be merged. Thanks again for raising this.

hubwriter:
This PR is ready to merge.
We're currently not merging into this repo but we'll get this merged in a few days.

Issue #442, opened 2020-10-12
Analytics


asadovtair:
Analytics

asadovtair:
@asadovtair ``

asadovtair:
_Sent from my Mi MIX 3 5G using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_asadovtair@gmail.com

asadovtair:
![](https://i.imgur.com/JUuzFOO.png)


_Sent from my Mi MIX 3 5G using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_asadovtair@gmail.com

Issue #210, opened 2020-10-07
Improve documentation on variables and reusables


casals:
### What article on docs.github.com is affected?

The specific `README.md` files for both variables and reusables; `CONTRIBUTING.md`

### What part(s) of the article would you like to see updated?

As they are now, these files are a little bit dry - it's not something that's intuitively usable by first-time contributors. I suggest both of these files could be updated to look like the one for [liquid tabs](https://github.com/github/docs/blob/main/lib/liquid-tags/README.md), with references, usage examples, etc. 

Also - adding a brief section on each of these (liquid tabs, variables, reusables, etc) would be extremely helpful.

## Content design plan

- [ ] Open for an OS contributor: For specific `README.md` files, I think they should be as complete as possible: definitions, examples, everything that a content contributor could learn about what's being referenced (code intrincacies = negative scope).
- [ ] @janiceilene will make these changes: As for `CONTRIBUTING.md`: IMHO it should be a complete quick start. Let's take, for instance, the last 24h: I was learning about variables, reusables, and callout tags as the PRs were reviewed. I mean, the technical detailes are there if I follow the md links, but the text doesn't give me any reasons to do so. And even if I do it, there's no clear benefit for the newcoming contributor - it reads and feels like technical documentation (which is technically not wrong, but not necessarily helpful to a newcomer as such). 
  - [ ] Structure the CONTRIBUTING.md file more like a quick start guide - "welcome, here's the basic stuff you need to know, here's why you need to know it, here's where you can learn more about each of them" - paced, with simple examples. Perhaps something like [this](https://github.com/MovingBlocks/Terasology/wiki/How-to-Work-on-a-PR-Efficiently).

janiceilene:
@casals Those two docs could definitely be better! Any thoughts about what kinds of information you think would be helpful there?

casals:
Sure - for specific `README.md` files, I think they should be as complete as possible: definitions, examples, everything that a content contributor could learn about what's being referenced (code intrincacies = negative scope).

As for `CONTRIBUTING.md`: IMHO it should be a complete quick start. Let's take, for instance, the last 24h: I was learning about variables, reusables, and callout tags as the PRs were reviewed. I mean, the technical detailes are there if I follow the md links, but the text doesn't give me any reasons to do so. And even if I do it, there's no clear benefit for the newcoming contributor - it reads and feels like technical documentation (which is technically not wrong, but not necessarily helpful to a newcomer as such). 

[...]

PR #444, opened 2020-10-12
Configure referrerPolicyo

asadovtair:
Set the default referrer-policy header to strict-origin-when-cross-origin

### Check off the following:
#210 #441 #442 #443 #328

### DIFF
--- a/middleware/index.js
+++ b/middleware/index.js
@@ -23,6 +23,7 @@ module.exports = function (app) {
   app.use(require('./loaderio-verification'))
   app.use(require('./cors'))
   app.use(require('./csp'))
+  app.use(require('./referrer-policy'))
   app.use(require('helmet')())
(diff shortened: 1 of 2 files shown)


JasonEtco:
👋 thank you for contributing! Can you elaborate on why this change is needed? What does having the `referrer-policy` header do for us, and is there an open issue that explains this? Thanks!

asadovtair:
_Sent from my Mi MIX 3 5G using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

heiskr:
It looks like `helmet()` already calls this referrer policy middleware, though the default it sets to is `no-referrer`.

https://helmetjs.github.io/

I think if we were to change this property, I'd prefer the configuration option over calling the same middleware twice.

PR #482, opened 2020-10-13
Actions: Add section about group/ungroup commands

Simran-B:
### Why:

Missing in Actions documentation, but available here:
https://github.com/actions/toolkit/blob/main/docs/commands.md#group-and-ungroup-log-lines

### What's being changed:

- Add new section about grouping log lines
- Improve table markup
- Change headline level of _Multline strings_ example from h4 to h5 (make Example a child section of Multiline string section)

---

image

---

Changed headline level:

image

---

### Check off the following:
- [ ] ~All of the tests are passing.~
- [x] I have reviewed my changes ~in staging~ locally.


Simran-B:
Hm, should I retake the screenshot at 150% or 200% browser scale? That would make it look crisper on high density screens.

janiceilene:
Thanks so much @Simran-B! I'll get this triaged for review 🎉

Simran-B:
The screen shot looks much better on my phone now, but it's displayed quite large on desktop. Is there an existing solution to adjust the size based on native pixel size or something along this line?

https://docs-482--actions-group-comman.herokuapp.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#grouping-log-lines

janiceilene:
Sorry about the delay @Simran-B! We've got quite a few things going on that our small team is taking care of, but this is definitely not stale and we'll get it reviewed 💛

content/actions/reference/workflow-commands-for-github-actions.md
```diff
@@ -158,6 +158,25 @@ Creates an error message and prints the message to the log. You can optionally p
...
 ```
 
+### Grouping log lines
+
+```
+::group::{title}
+::endgroup::
+```
+
+Emitting a `group` with a title will instruct the logs to create a collapsible region up to the next `endgroup` command.
```
rachmari, line 168: It looks like the description from https://github.com/actions/toolkit/blob/main/docs/commands.md#group-and-ungroup-log-lines. Maybe we can expand that description just a bit.

```suggestion
Creates a collapsible group in the log. To create a group, use the `group` command and specify a `title`. Anything you print to the log between the `group` and `endgroup` commands is nested inside a collapsible group in the log.
```

rachmari (commented):
@Simran-B thanks for catching this and updating the docs! ✨ I just had one suggestion to change the language to match the tone style guide for the docs. Once that is updated, I can get this shipped for you! 🚢

content/actions/reference/workflow-commands-for-github-actions.md
Simran-B: I like this change. Maybe change the last bit to 

> nested inside an expandable entry in the log.

My idea is to use something less repetitive here and also to provide a slightly different description of what this actually is so that the user has a better chance of understanding it. Or would this make it more confusing than helpful in the end, if it switches between "collapsible group" and "expandable entry"?

Simran-B: Actually, "collapsible group" and "expandable group" are different. The former starts expanded, while the latter can be collapsed. The default state in the Web UI is collapsed, which would make it an expandable group (if you watch the log live, then there is actually a moment where it's expanded, but it gets collapsed a bit later)

rachmari: This suggestion sounds great and I agree it's less repetitive and more accurate with expandable rather than collapsible! I'll go ahead and commit this!

rachmari (approved):
@Simran-B thank you so much for cleaning up the table and adding the group command! This looks fantastic. 🙌 I'm going to ship this right away. 🚀 🚢

PR #531, opened 2020-10-14
Add more details about workflow_run

programmer04:
### Why:

Documentation about `workflow_run` does not contain very important info about its special behavior:

- run only on default branch regardless of whether it is triggered by a workflow run on other branches or not

- has to exist on the default branch

This behavior was verified by me and described in [community support](https://github.community/t/on-workflow-run-does-not-work-for-me/128833) - kudos to @BrightRan 🙂 

### What's being changed:

New info added

### Check off the following:


janiceilene:
Thanks for opening a PR @programmer04! I'll get this triaged and up for review ⚡

janiceilene:
Sorry about the delay @programmer04! This is definitely not stale 🙃 we're a small team working through lots of amazing Hacktoberfest PRs. Thank you for your patience 💖

content/actions/reference/events-that-trigger-workflows.md
```diff
@@ -655,7 +655,7 @@ on:
 
 {% data reusables.webhooks.workflow_run_desc %}
```
rachmari, line 656: It looks like this section is missing the table that the other event sections have. We should add that rather than including the information in the description:

```suggestion
{% data reusables.webhooks.workflow_run_desc %}

| Webhook event payload | Activity types | `GITHUB_SHA` | `GITHUB_REF` |
| --------------------- | -------------- | ------------ | -------------|	
| [`workflow_run`](/webhooks/event-payloads/#workflow_run) | - n/a | Last commit on default branch | Default branch |	
```

content/actions/reference/events-that-trigger-workflows.md
```diff
@@ -655,7 +655,7 @@ on:
...
 
-If you need to filter branches from this event, you can use `branches` or `branches-ignore`.
+If you need to filter branches from this event, you can use `branches` or `branches-ignore`. Workflow has to exist on the default repository branch and will only run on it, regardless of whether it is triggered by a workflow run on other branches or not.
```
rachmari, line 658: After adding the table above, I think we can ✂️ the addition:

```suggestion
If you need to filter branches from this event, you can use `branches` or `branches-ignore`.
```

rachmari (commented):
@programmer04 thank you for the docs update!! 🙌 Your change actually uncovered a larger bug in this section of the docs. This section doesn't include the table that describes which commit and reference are used when running the workflow. Instead of adding the information to the description, I've suggested adding the missing table.

BrightRan:
Hi guys,
The docs seems still is not updated.

programmer04:
Thanks, @rachmari for your suggestions I've applied them. 🙂 

Moreover, maybe you can pass a feature request to the team in GH taking care of `workflow_run`... Having a `workflow_run` which works in the described way would be useful too. 

On branch `feature-1` from some event, `workflow-A` is triggered. Next dependant `workflow-B` (which is triggered by successful finishing `workflow-A` by using `workflow_run`) instead of running on the default branch (current default and the only one behavior) can be specified to be run on a current branch in this case `feature-1`

rachmari (approved):
@programmer04 thank you so much for getting this updated! 🙌 This looks great. I'll get it merged right away! 🚢 

I've also created an internal customer feedback issue to track your request:

> Moreover, maybe you can pass a feature request to the team in GH taking care of workflow_run... Having a workflow_run which works in the way described below in my opinion would be useful too.

> On branch feature-1 from some event, workflow-A is triggered. Next dependant workflow-B (which is triggered by successful finishing workflow-A by using workflow_run) instead of running on the default branch (current default and the only one behavior) can be specified to be run on a current branch in this case feature-1

PR #541, opened 2020-10-15
chore: Add Markdownlint checking and CI

nschonni:
### Why:
Linting the markdown will  ensure that common rendering issues are caught and that the files are consistently formatted.

### What's being changed:

- Add a markdownlit config with all the current rules that are failing turned off
- Add a CI job to run whenever markdownfiles or the config are changed
- Update the nested code fences inside ordered list that had incorrect indenting. I would have skipped that to later, but it was crashing the parser
- Updated the translation files for the same reason in a separate commit, but they could be undone and ignored
- Probably easier to review with the "Hide whitespace changes" first, before looking at the rich-diffs

### Check off the following:


nschonni:
@chiedo I noticed you changed all the action version to shas. Is that something that I need to do, or would that happen if this is accepted afterwards?

chiedo:
👋🏿  @nschonni thanks for this PR! Yes the sha changes will automatically apply. They should be present once you click the "Update branch" button.

We're running a little behind on things but we'll review this soon!

nschonni:
@chiedo no worries, I see you have a bunch of hacktoberfest spam to deal with right now 😆 

I hit the update, but since this is proposing a new Action file, I figured it might need some manual intervention

chiedo:
Got it! We'll take a look.

nschonni:
@chiedo ha, figured it out based on the CI failure. I copied the values from the `pa11y.yml`

content/admin/installation/installing-github-enterprise-server-on-google-cloud-platform.md
```diff
@@ -27,7 +27,7 @@ Before launching {% data variables.product.product_location_enterprise %} on Goo
 {% data variables.product.prodname_ghe_server %} is supported on the following Google Compute Engine (GCE) machine types. For more information, see [the Google Cloud Platform machine types article](https://cloud.google.com/compute/docs/machine-types).
 {% if currentVersion != "free-pro-team@latest" %}
 | High-memory   |
-  ------------- |
+| ------------- |
```
nschonni, line 30: This is the only "functional" change, as it broke on both the English and translated documents

nschonni (commented):
Rebased. If reviewing with the whitespace diff off, I commented on the only text change on the English doc that required changes in the translated docs because the table format issue there got mangled in the translation parser

chiedo:
@rachmari / @janiceilene  would you be willing to take this one over? This would impact the writers much more heavily than engineering.

From an engineering perspective, the change I would suggest is that the `translations` directory is exempt from the linter. Because everything in that directory comes from Crowdin.

nschonni:
@chiedo I can ignore the `translations` folder, or setup a separate job that you could ignore if it's failing, but I have seen Crowdin mangle formatting sometimes. EX: the tables in this PR when it misparsed the orignal english version

chiedo:
Thanks @nschonni ! I understand why you'd suggest that. But changes in the `translations` directory gets overridden regularly by a service called Crowdin. So anything issues we fixed would return with the next sync.

nschonni:
@chiedo yup, I get not wanting to take PRs directly on the content because Crowdin overwrites this (I was involved in setting it up for Node.js), but checking the content coming from Crowdin makes sure the content isn't malformed when it comes back from there.

The original parsing issue that made me need to modify the MD files in this PR has been fixed upstream, so if a release comes out there soon, I'll pull out those content changes to a new PR.

I'm thinking that this should be split into 2 jobs with 2 different path filters to handle the translations separately though. That way the second translations-only job could fail/be ignored while the translations are addressed in Crowdin, and not show a failure on other MD changes

nschonni:
@chiedo OK, split the job so there is different filtering based off of the translated vs. regular content. This way you can catch if Crowdin pushes broken markdown, but those failures won't show up on PRs for non-translated content

nschonni:
I can also rebase out the translated file changes, but the second job will crash till they are addressed

chiedo:
Got it thanks! We'll wait and see what @janiceilene and the writers think. This one will be their call as it would affect their workflows.

nschonni:
@janiceilene a new version of the Markdownlint library was release to address the parsing issue I ran into with the nested code fences, so I've rebased out those commits. The content changes are still over in #675, so this PR can just concentrate on the CI/CLI addition

janiceilene:
Thanks so much @nschonni! We're a little backed up right now, but we'll take a look at this as soon as we can 💛

nschonni:
Just flipped to Draft because it was pointed out on another PR that `lint` is a required job, and I think the path filtering needs to change to use fkirc/skip-duplicate-actions instead of the built-in `paths`

nschonni:
@JamesMGreene copied over the filtering setup from the js-lint job

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - uses: actions/cache@d1255ad9362389eac595a9ae406b8e8cb3331f16
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
+
+      - name: Install Dependencies
+        if: steps.cache.outputs.cache-hit != 'true'
```
JamesMGreene, line 40: I think the previous step needs an `id` of "cache" for this to work. 🤔

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: npm ci
+
+      - name: Run Markdownlint
+        run: |
+          echo "::add-matcher::.github/workflows/markdownlint-problem-matcher.json"
```
JamesMGreene, line 45: What is this line doing?

.github/workflows/markdownlint.yml
```diff
@@ -0,0 +1,46 @@
...
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@36feb0d8d062137530c2e00bd278d138fe191289
+        with:
+          cancel_others: 'false'
+          github_token: ${{ github.token }}
+          paths: '["**/*.md", "!translations/**", "package*.json", ".markdownlint.json", ".github/workflows/markdownlint.yml", ".github/workflows/markdownlint-problem-matcher.json"]'
+          paths_ignore: '["translations/**"]'
+  lint:
```
JamesMGreene, line 25: Can we rename this to something other than "lint" to avoid confusion with the JavaScript-linting job? Maybe `mdlint` or `markdown-lint`? 🤷‍♂️

JamesMGreene (commented):
Very cool! I had looked into markdownlint before but we hadn't found time to invest into it yet.

I left a few comments but I'm OOO today so I'll defer a more detailed review to someone else (or my future self). ☺️

.github/workflows/markdownlint.yml
JamesMGreene: Perhaps this would work as the default ID might match the Action's name? Not sure. Either way, I'd personally like to see an explicit `id` added when referencing it later. 🤷‍♂️

.github/workflows/markdownlint.yml
nschonni: This registers the problem matcher file so that PRs get in-line comments on the file tab if the markdownlint job puts any issues to the jobs console

.github/workflows/markdownlint.yml
nschonni: OK, I'll see about adding it, but I think this was just copied from the existing Lint job

.github/workflows/markdownlint.yml
nschonni: Figured it was better to use the same job name so it stays a required check

nschonni:
@JamesMGreene gave this a rebase for the package-lock.json conflict if you want to remove the "stale" label

bojannovakovic13 (approved):

chiedo:
Thanks again for another contribution but we're not sure when we can prioritize this. But we've got this PR documented for the future if we choose to move forward with this approach!

DavidAnson:
@chiedo @janiceilene @JamesMGreene Are you able to share the concerns or limitations that led to rejecting this PR? If there are changes to the markdownlint toolset that would make it more broadly applicable, I'd love to consider them. Thanks!

chiedo:
@DavidAnson thanks for your offer! On our end, time, focus and prioritization prevented us from doing so. But if we chose to get all the potentially impacted parties together in the future to get this in place we'll chime back in here!

DavidAnson:
It's a deal, thanks! 😀

PR #558, opened 2020-10-15
Ensure that the Actions Skipper never errors out

chiedo:
Make sure that if the duplicate actions checker fails, it still shows a successful check. As the Actions skipper is non-blocking.

### Check off the following:


JasonEtco (approved):
Ah nice, we should have caught this in earlier reviews 🤦

PR #563, opened 2020-10-15
Fix the autoupdatebranch Action

chiedo:
When moving to shas for more security, it looks like we broke the Docker action.

This should get it working again. Ideally we move to a Sha somehow in the future but I'm unsure how to do that considering the unique format of this Action being served from Docker rather than GitHub.

Example of the Action breaking: https://github.com/github/docs/actions/runs/309155706/workflow

### Check off the following:


JamesMGreene:
Is it maybe that the Docker registry digest SHAs don't match the Git commit SHAs...? 🤷🏻‍♂️ 
https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier

I am admittedly not well-versed in Docker. 😕

JamesMGreene (approved):
Approving this as a revert to get things functioning again. 👍🏻

PR #610, opened 2020-10-16
A quick experiment to get allow React components in Markdown pages and throughout the site

chiedo:
This is an experiment to get React components working in markdown pages. Super rough and a proof of concept.

With the goal being to make our codebase more extensible by giving us access to the entire React ecosystem.

Read about how it works [here](https://github.com/github/docs/edit/experiment-with-react-and-mdx/react/README.md?pr=%2Fgithub%2Fdocs%2Fpull%2F610)

## Todos

- [x] Get static React Components working
- [x] Get server side rendering working
- [x] Get Dynamic React Components working
- [x] Get client side hydration working 
- [x] confirm the passing of props works 

## Todos if we move forward with this experiment

- [x] get tests working again
- [x] Use the production react script for react in production instead of the React development script that's currently in-use in the header

## Worth considering in the future

- Consider caching server side rendering of react components in Redis. eg. Rendering the components is not free. eg. the Code tab below was 30ms to render. Although this may not be needed as we already cache page responses

## Check off the following:


chiedo:
@JasonEtco would you be down to 🍐  on this at some point? server side rendering is working but I need some help with the webpack and logic to get the react components hydrated client side.

chiedo:
Thinking we need to do something like this (pseudocode) at the bottom of every component in `./react/*`

This should make sure that when in the browser every component gets hydrated 
```
if(window.type){
   const componentContainers = document.querySelector('.unique-class-name-of-component');

   for  (const index of componentContainers){

        componentContainer = componentContainers[index];
        ReactDOM.hydrate(React.createElement(ComponentName), componentContainer);

}

```

And then part two would be figuring out how to get the react files that work on the server working in the browser. Assuming webpack would be needed for that .

chiedo:
Did the first part of what I think I will work here: https://github.com/github/docs/pull/610/commits/4d9a6c70a83f5c71de2fae3c2e8ff8d275e21003

Next need to figure out webpack

chiedo:
Got it working!

Jeff-GiJoe:
Sorry for the late reply.  How do I review this? 😅

webpack.config.js
```diff
@@ -12,9 +13,19 @@ module.exports = {
   },
   module: {
     rules: [
+      {
+        test: /\.js?$/,
```
townofdon, line 17: Should this be `test: /\.js$/,`? (note the omitted question mark)

react/CoolTable.js
```diff
@@ -0,0 +1,40 @@
...
+    <MUIDataTable
+      title={"Employee List"}
+      data={data}
+      columns={columns}
+      options={options}
+    />
+  )
+}
+
+if (typeof window === 'undefined') {
```
sarahs, line 30: Same question as above.

content/github/getting-started-with-github/access-permissions-on-github.md
```diff
@@ -17,16 +18,42 @@ A repository owned by a user account has two permission levels: the *repository
 
 ### Organization accounts
 
+#### Code Block Component
+<!--react-->
+<CodeBlock language="javascript"
```
sarahs, line 23: Looks like the React component adds languages, so we should be able to drop this in the Markdown?
```suggestion
<CodeBlock
```

react/CoolTable.js
```diff
@@ -0,0 +1,40 @@
...
+ ["John Walsh", "Test Corp", "Hartford", "CT"],
+ ["Bob Herm", "Test Corp", "Tampa", "FL"],
+ ["James Houston", "Test Corp", "Dallas", "TX"],
+];
+
+const options = {
+  filterType: 'checkbox',
+};
+
+const CoolTable = function () {
```
sarahs, line 19: I know this is just an example, but should the columns and data be able to be passed here via props? Thinking we should probably not add this to the codebase unless it's a working component.

react/CodeBlock.js
```diff
@@ -0,0 +1,81 @@
+const React = require('react')
+const ReactDOM = require('react-dom')
+const { Prism } = require('react-syntax-highlighter')
```
sarahs, line 3: We use http://ghub.io/rehype-highlight (via `docs/render-content` via [hubdown](https://github.com/electron/hubdown#plugins)) for syntax highlighting everywhere else. If we use Prism here, we're probably going to get different highlighting?

react/CodeBlock.js
```diff
@@ -0,0 +1,81 @@
...
+const { dark } = require('react-syntax-highlighter/dist/cjs/styles/prism')
+const { Tab, Tabs, TabList, TabPanel } = require('react-tabs')
```
sarahs, line 5: Can we use the Primer component here instead? https://primer.style/components/TabNav

sarahs (commented):
@chiedo This is really cool! The examples work nicely. :zap:

I did a quick review (since this is still in progress) and left a few early thoughts. I didn't dig into any of the Babel stuff.

One note: this adds a lot of dependencies to the docs app. I wonder if we can shift any of that dependency burden to [Primer React components](https://primer.style/components/)?

lib/react/engine.js
```diff
@@ -0,0 +1,23 @@
...
+const CodeEditor = require('../../dist/react/CodeEditor')
+/* eslint-enable */
+
+const renderReact = async componentStr => {
+  const componentName = componentStr.match(/[^<]([a-zA-Z])*/gm)[0]
+  const jsx = `<div className="react-component-${componentName}">\n${componentStr}\n</div>`
+  const component = transform(jsx)
+
+  // eslint-disable-next-line
+  return renderToString(eval(component))
```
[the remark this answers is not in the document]
chiedo: It's eval out of the box. But something like this would probably be safer. 

https://www.npmjs.com/package/safe-eval

server.js
```diff
@@ -5,9 +5,27 @@ require('./lib/feature-flags')
...
+for (const index in tree.children) {
+  const file = tree.children[index]
+  if (file.type === 'file') {
+    if (!fs.existsSync(path.join('dist', 'react'))) {
+      fs.mkdirSync(path.join('dist', 'react'), { recursive: true })
+    }
+    const content = transform(fs.readFileSync(file.path, 'utf8'))
+    fs.writeFileSync(path.join('dist', file.path), content)
+  }
+}
```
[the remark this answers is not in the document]
chiedo: Yep! Did so here. 

https://github.com/github/docs/pull/610/commits/548853fb2973b61d3c3f39526507f0ceb41e07b1

With that being said, I'm not convinced this is the way we should go about it. It was very much a, "solve the hard problems first, not the time consuming ones" approach

react/CodeEditor.js
```diff
@@ -0,0 +1,41 @@
...
+        onChange={onChange}
+        name={`code-editor-${hash({ language: props.language, code: props.code })}`}
+        editorProps={{ $blockScrolling: true }}
+        defaultValue={props.code}
+      />
+    </div>
+  )
+}
+
+if (typeof window === 'undefined') {
```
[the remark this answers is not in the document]
chiedo: Nope! Just me being lazy while hacking away. Patched.

webpack.config.js
chiedo: Yes it should! Patched.

lib/page.js
```diff
@@ -119,9 +121,30 @@ class Page {
...
 
-    const markdown = this.mapTopic
+    let markdown = this.mapTopic
       ? getMapTopicContent(this, context.pages, context.redirects)
       : this.markdown
+    
+    // If the article is interactive parse the React!
+    if (this.interactive) {
+      // Search for the react code comments to find the react components
+      const reactComponents = markdown.match(/<!--react-->(.*?)<!--end-react-->/gms)
```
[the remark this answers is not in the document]
chiedo: Removed!

lib/page.js
```diff
@@ -119,9 +121,30 @@ class Page {
...
+
+      // Render each of the react components in the markdown
```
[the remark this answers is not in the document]
chiedo: Thank you! Switched to your recommendation. Although components can be inline, at least based on how I implemented it so I just took your Promise.all logic.

lib/react/engine.js
```diff
@@ -0,0 +1,23 @@
...
+// These all need to be here even though eslint doesn't think so
+/* eslint-disable */
+const React = require('react')
+const CodeBlock = require('../../dist/react/CodeBlock')
+const CoolTable = require('../../dist/react/CoolTable')
+const CodeEditor = require('../../dist/react/CodeEditor')
+/* eslint-enable */
+
+const renderReact = async componentStr => {
+  const componentName = componentStr.match(/[^<]([a-zA-Z])*/gm)[0]
```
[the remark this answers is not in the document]
chiedo: Accepted!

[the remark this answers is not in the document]
chiedo: And added comments.

react/CodeBlock.js
chiedo: Yeah we could! My work was solely to do a proof of concept to make sure we could do whatever we want. I'm not actually recommended the use of any of these specific react components. Just making sure we can render any react component and have it be functional.

react/CodeBlock.js
chiedo: Same answer as https://github.com/github/docs/pull/610#discussion_r508502786

react/CoolTable.js
chiedo: Yep haha. I decided not to continue with that Component example. But I removed it from the codebase! And just to be clear, I'm not suggesting we add any of this to the codebase. My whole goal was a proof of concept. We can figure out what to do with all this stuff tomorrow after the demo!

chiedo:
Another idea I had was, if we wanted an even cleaner syntax, we could replace this syntax:

<!—react—>
<Component />
<!—end-react—>

With this syntax:

```react
<Component />
```

Just an idea... we can have syntax that’s less obnoxious if we want to

lib/react/engine.js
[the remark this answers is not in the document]
chiedo: Couldn't help it! Replaced with `new Function`.

Safe now!

nschonni:
Would it also change the extension to MDX for those files? I'm guessing it's following this style https://mdxjs.com/

chiedo:
@nschonni I didn't consider that but that would make a lot of sense if we were using MDX. I initially planned to use MDX but then abandoned it and built the functionality from the ground up. So no mdx extension since we're not using MDX and that would require we make lots of updates to our parsing. Rather we just use frontmatter `interactive: true`

chiedo:
@heiskr how would you feel about trying to get tests working on this this week so we can merge the work. Super low risk / high value in my opinion if we get the tests passing.

heiskr:
I think this pull request works as a proof of concept, but to get it production ready we might need a fresh branch so we're only including the MVP for the first deployment.

chiedo:
Works for me!

lib/page.js
```diff
@@ -133,10 +134,31 @@ class Page {
...
+    if (this.interactive) {
+      // Search for the react code comments to find the react components
+      const reactComponents = markdown.match(/<!--react-->(.*?)<!--end-react-->/gs)
+
+      // Render each of the react components in the markdown
+      await Promise.all(reactComponents.map(async (reactComponent) => {
+        let componentStr = reactComponent
+
+        // Remove the React comment indicators
+        componentStr = componentStr.replace('<!--react-->\n', '').replace('<!--react-->', '')
```
heiskr, line 151: Possible to use a RegExp with replace, like 
```suggestion
        componentStr = componentStr.replace(/<!--react-->\n?/g, '')
```

lib/react/engine.js
```diff
@@ -0,0 +1,57 @@
+const { renderToString } = require('react-dom/server')
+const { transform } = require('./babel')
+const React = require('react')
+const fs = require('fs')
+const path = require('path')
+const dirTree = require('directory-tree')
```
heiskr, line 6: We also have a `walk-sync` module, but I don't know how different these are

heiskr (approved):
🚀 
I removed the specific components (CodeBlock and CodeEditor) so this PR is more likely to get shipped, we can always pull them out of the version history of the branch when we want to add them later.

lib/page.js
chiedo: Can you confirm this works @heiskr and if it does can you commit it? I get confused by the first replace having a new line character and the next one not. I personally like coding verbose. But that's a personal preference.

lib/react/engine.js
chiedo: You want to try and switch to it @heiskr  so we don't have to add another package?

chiedo:
> I removed the specific components (CodeBlock and CodeEditor) so this PR is more likely to get shipped, we can always pull them out of the version history of the branch when we want to add them later.

Before removing those components, did you confirm that the test suite would still pass when a React component was embedded in a markdown file?

If so, is this ready to 🚢 ? 🙏🏿

heiskr:
> Before removing those components, did you confirm that the test suite would still pass when a React component was embedded in a markdown file?

The components would need their own tests, and we don't have a pattern for testing React components yet, but I don't think that all has to happen in one pull request. Given that its passing and we can add React components later, I think that's sufficient for the first pull request.

> Also, I added usage documentation here, can you give that a 👀 ✔️

I've looked at it, and I fixed a few typos but nothing major. Looks good to me.

lib/page.js
heiskr: I think its fine as is, just an FYI more than anything.

lib/react/engine.js
heiskr: It looks like directory-tree and walk-sync have different output. walk-sync creates a flat array of just the file names, where directory-tree has a nested structure. I think there might be consequences to the difference as this code seems to look at just the first level. I think it's okay to ship directory-tree for now, and we can always change it later if we find another pattern we prefer.

swampthing1984:
/

PR #650, opened 2020-10-19
Update events-that-trigger-workflows.md

cmcaine:
### Why:

The wrong field is `required` in the `workflow_dispatch` event example.

The text says:

> This example defines the `name` and `home` inputs and prints them using the `github.event.inputs.name` and `github.event.inputs.home` contexts. If a `name` isn't provided, the default value 'Mona the Octocat' is printed.

But `name` is required and `location` is not required and has no default.

### What:

Make `name` optional and `location` required.

### Check off the following:

I have not done any of these, this is a trivial change and your CI should catch issues :)


drumenbassik (approved):

drumenbassik (approved):

lucascosti:
👋 @cmcaine Thanks a lot for this PR!

You are correct that the description doesn't really match the example. However, I think we can change it so that the behavior is clearer, rather than switching the required values for the fields. I'll submit a new commit to your PR so it hopefully makes more sense!

lucascosti:
Ok, change made in https://github.com/github/docs/pull/650/commits/6bfe64d723f1df182a54c6a01fa9b4d6864dc275

You can see a preview here: https://docs-650--patch-1.herokuapp.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-workflow-configuration

@cmcaine: Let me know if this makes sense for you as well, and I'll merge it in 🙂

lucascosti (approved):

cmcaine:
That looks like an improvement :)

but line 114 setting the default name should maybe be removed if a name is required? Or maybe default & required is fine?

lucascosti:
> That looks like an improvement :)
> 
> but line 114 setting the default name should maybe be removed if a name is required? Or maybe default & required is fine?

Yep, default & required is fine 🙂. For the UI, this results in the field being pre-populated with the default value. e.g.:

image

I'll get this merged in!

cmcaine:
But what is the difference between `default` and `default + required`?

In both cases it is impossible to submit an empty string, because it will be replaced by the default, right? So what does `required` add?

On a field without a default, `required` does something more interesting.

lucascosti:
> But what is the difference between `default` and `default + required`?
> 
> In both cases it is impossible to submit an empty string, because it will be replaced by the default, right? So what does `required` add?
> 
> On a field without a default, `required` does something more interesting.

* `default` only: Prepopulates the UI with the default value. You can submit an empty string in the UI, but if you do, the empty value will be replaced with the default value in the workflow run execution.
* `default + required`: Prepopulates the UI with the default value. You can't submit empty string in the UI; something must be provided.

PR #677, opened 2020-10-20
check repo is docs-internal in both jobs

rachmari:
I wasn't checking that the repo is docs-internal in the second job. That is causing a few failures in workflows in the public repo.


heiskr (approved):
🚢

.github/workflows/first-responder-docs-content.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
   first-responder-remove-pr:
     name: Remove PR from FR project board
-    if: (github.event.label.name == 'docs-content-fr' && github.event.action == 'unlabeled') || github.event.action == 'closed'
+    if: github.repository == 'github/docs-internal' && (github.event.label.name == 'docs-content-fr' && github.event.action == 'unlabeled') || github.event.action == 'closed'
```
jamiefontenot, line 56: repeat 16 as 18 for deposit as its state collection

jamiefontenot (commented):
check git up external
x

PR #722, opened 2020-10-22
Changed the term master to main

pr0PM:
Since we are updating it everywhere :)

### What's being changed: Glossary 

The glossary section has been updated in all the relevant places that came under my eyes while reading it.

### Check off the following:


janiceilene:
Thanks so much for the PR @pr0PM! I'll get this triaged for review 🌟

data/glossaries/external.yml
```diff
@@ -395,10 +395,10 @@
     Markdown and uses a particular form of Markdown called GitHub Flavored Markdown. See [GitHub Flavored Markdown Spec](https://github.github.com/gfm/) or [Getting started with writing and formatting on GitHub](/articles/getting-started-with-writing-and-formatting-on-github).
 - term: Markup
   description: A system for annotating and formatting a document.
-- term: master
+- term: main
   description: >-
     The default development branch. Whenever you create a Git repository, a
-    branch named "master" is created, and becomes the active branch. In most
+    branch named "main" is created, and becomes the active branch. In most
```
hubwriter, line 401: > Whenever you create a Git repository, a branch named "master" is created

Unfortunately this isn't true. If you create a directory, go into it, and do `git init` to create a new repository, in most cases you'll find yourself in a branch called `master` - unless you have configured a git template that changes this to something else, like `main`.

I'd suggest we simplify the definition of `main` here to just:

> Typically, the default branch for repositories on GitHub is called `main`.

hubwriter (changes requested):
Thanks for pushing this pull request. We should definitely update the glossary to use `main`. I've suggested a change to the definition for `main`.

janiceilene:
@pr0PM If you want to make the changes suggested in this [comment](https://github.com/github/docs/pull/722#discussion_r512120574) and let us know when they're done, we can get this PR merged! Thank you 💝

data/glossaries/external.yml
pr0PM: So shall I make a commit with proposal to add separate definition of `main` in the glossary as suggested by you, keeping the definition of `master` intact since it's the current standard in Git?

hubwriter: Good point. I think you're right: we shouldn't do away with `master` in the glossary all together. Could you commit another change that alters the definition for `main` and adds an entry for `master`. I think we should simplify the definition of `master` and indicate that it's not a term used at GitHub. How about:

> The default branch in many Git repositories. By default, when you create a new Git repository on the command line a branch called `master` is created. Many tools now use an alternative name for the default branch. For example, when you create a new repository on GitHub the default branch is called `main`.

pr0PM: done

hubwriter (approved):
@pr0PM - thanks. I'll merge this now.

PR #762, opened 2020-10-23
Update managing-complex-workflows.md

### LINKED ISSUE #761: Secrets example in Managing complex workflows doesn't work as expected

https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#storing-secrets

The _Storing secrets_ example uses different cases for storing and for using the defined environment variable and so doesn't work as expected.

### Additional information

I am making a pull request to address the issue (#762).

gsrohde:
Environment variable names are apparently case-sensitive since without this change, the `example-command` (I tried it with `echo`) is fed the empty string.  (On the other hand, environment variable names are case-_insensitive_ in one respect: Trying to define two environment variables whose names differ only in case yields a "duplicate definition" error.)  This relates to issue #761.

### Why:

The example does not work as expected.

### Check off the following:


janiceilene:
Thanks for opening a PR @gsrohde! I'll get this triaged for review 🌟

ha1dyansyah:
761

lucascosti:
Thanks a lot for this, @gsrohde! You are correct for most *nix OSs, but it _might_  depend on the OS and shell of the runner executing the job.

This example doesn't specify a `runs-on:` key so it's ambiguous. I'll add a slight edit to your PR to add the key with an `ubuntu-latest` runner so it's correct for Linux runners (by far the most popular OS in Actions) and it shows the case of the environment variable should be the same.

lucascosti (approved):

PR #768, opened 2020-10-23
chore: Only run freeze reminder on internal repo

nschonni:
### Why:

Shouldn't affect forks or the public repo, but I'm guessing this will start triggering double notifications running on the public and private repo on the same cron schedule. The `env.FREEZE` check could probably be hoisted here too

### Check off the following:


heiskr (approved):
Thanks! 🚢

JamesMGreene:
I appreciate the critical eye and the contribution! 🙇🏻‍♂️ 💝 

FWIW, I intentionally left this out when I created it to ensure we remembered to unfreeze both repos when the time came. We can try it this way and hopefully it works out as well. 😅

nschonni:
@JamesMGreene fair enough, I figured it would more be noise in the slack channel, but feel free to revert as I can see it might catch only clearing the variable in one repo :)

PR #836, opened 2020-10-27
Clarify artifact upload/download mechanism

sveniu:
### Why:
Doc link: https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts

The article seems to confuse the term "file", "artifact" and "directory" in certain places, which can be very confusing to the reader.

This formulation is problematic since, by not qualifying "name", the reader may think it refers back to "file", while it indeed refers back to "upload" (aka "artifact"). In other words, it may be understood as if the `math-homework.txt` file changes name to `homework`.

This appears to be an error, in that the `homework` directory doesn't exist. Using `download-artifact@v2` with a named artifact, will place the artifact contents in the current working directory. Only without a named artifact will it create separate directories for each downloaded artifact. See also this compatibility note from `download-artifact`'s README.md:

> When using download-artifact@v1, a directory denoted by the name of the artifact would be created if the path input was not provided. All of the contents would be downloaded to this directory.

The examples use v2, however.

This uses the arguably ambiguous "upload" term in lieu of "artifact". It may be confusing to the reader to discern between uploading of files vs uploading of artifacts.

### What's being changed:

The problems above are addressed, like so:
* Clarify the difference between file(s), directories and artifacts.
* Remove "places the file in a directory named ..". This seems to confuse directory with artifact name. (Alternatively, I speculate that it may needlessly document how file storage is organized on the server side.)
* Remove erroneous directory "homework" in step 2 of the example.


janiceilene:
Thanks so much for opening a PR and the thorough explanation of your changes! I'll get this triaged for review ✨

amein0:
O

rachmari (approved):
@sveniu this looks perfect. Thanks for clarifying this content in the docs! 🙌 I'll get this merged right away. 🚢 🚀

PR #862, opened 2020-10-28
chore: Pull JS Linting to separate Action

nschonni:
### Why:

Reduce number of jobs, so the JS linting is only run if JS files are changed. I left the duplicate job run in the test-translations.yml, but it could also probably be removed if the branch filters are updated here

### What's being changed:

A separate linting job that uses the pinned version of `standard` rather than a floating version from `npx` run only when JS files are changed. Relocated the previous `--fix` version of the script to a separate target

### Check off the following:

### DIFF
--- a/.github/workflows/js-lint.yml
+++ b/.github/workflows/js-lint.yml
@@ -1,0 +1,59 @@
+name: Lint JS
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches-ignore:
+      - translations
+
+jobs:
+  see_if_should_skip:
+    runs-on: ubuntu-latest
+
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@36feb0d8d062137530c2e00bd278d138fe191289
+        with:
+          cancel_others: 'false'
+          github_token: ${{ github.token }}
+          paths: '["**/*.js","package*.json",".github/workflows/js-lint.yml"]'
+
+  lint:
+    runs-on: ubuntu-latest
+    needs: see_if_should_skip
+    if: ${{ needs.see_if_should_skip.outputs.should_skip != 'true' }}
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675
+
+      - name: Setup node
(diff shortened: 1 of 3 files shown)


janiceilene:
Thanks @nschonni! I'll get this triaged for review ⚡

heiskr (approved):
👏🏼 This looks like a solid improvement.

JamesMGreene (changes requested):
I love what this change achieves! 😍 

However, I believe this _also_ implies that we will be unable to make the linting job be a **required check** for merging into the `main` branch, which is less desirable. 😕 

I'll confirm with the Actions support folks internally but I'm going to block this PR for now by marking my review as "Request changes". 🛑

nschonni:
No problem, if this way doesn't work, I can try and re-work it with fkirc/skip-duplicate-actions

nschonni:
@JamesMGreene figured out a way to test it by removing the trigger on the config change, and it looks like it will stall with the "Required".

image

I'll try with the other action, and do the same for https://github.com/github/docs/pull/541

nschonni:
@JamesMGreene I can't see if the now "Skipped" required check will block the commit. Can you let me know, and I can in-line the skip check flow into the `lint` job instead

JamesMGreene (approved):
Approving to check if it meets "Required Checks"... uhh... requirements. 😅

nschonni:
Oops, looks like the branch merge triggered the `lint` check. I'll rebase back to the first commit and re-submit the second one to see if we can get back to the "Skipped" required check

nschonni:
@JamesMGreene OK, it's back to the "Skipped" required check again

.github/workflows/js-lint.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@36feb0d8d062137530c2e00bd278d138fe191289
+        with:
+          cancel_others: 'false'
+          github_token: ${{ github.token }}
+          paths: '["**/*.js","package*.json",".github/workflows/js-lint.yml"]'
```
JamesMGreene, line 24: Let's add some spaces here for legibility. ✨ 

```suggestion
          paths: '["**/*.js", "package*.json", ".github/workflows/js-lint.yml"]'
```

JamesMGreene (approved):
Good news! ☝🏻  It actually seems to be happy with this. 🎉 

image

That's so awesome, @nschonni. It really cleans the test workflow up nicely. 😍

PR #915, opened 2020-10-29
chore: Only trigger first responder for internal

nschonni:
### Why:

Got failure notifications for this job when I [closed a PR](https://github.com/github/docs/runs/1328880168?check_suite_focus=true). This appears to be for internal use only, but the job filter wasn't quite working. If this is supposed to actually run on the public repo, then the target would need to change to `pull_request_target` so it can access the SECRET value, but the previous changes indicate it's only supposed to be internal

### What's being changed:

Ensure that the first repository filter condition isn't overridden by all "closed" event triggers.

### Check off the following:


heiskr (approved):
🙇🏼 Thanks!

PR #934, opened 2020-10-30
Rename Enterprise Administrator to Enterprise Server (actual product) for easier identification

jccwm:
### Why:

Looking for something for GHES, several of the Field Services engineers and I noted that Enterprise Server was misnamed as "Administrator" - named for the user, not the product.

### What's being changed:

Screen Shot 2020-10-30 at 1 02 47 PM

Screen Shot 2020-10-30 at 12 57 40 PM

### DIFF
--- a/content/admin/index.md
+++ b/content/admin/index.md
@@ -1,5 +1,5 @@
 ---
-title: Enterprise Administrators
+title: Enterprise Server
 redirect_from:
   - /enterprise/admin/hidden/migrating-from-github-fi/
   - /enterprise/admin


Suya159357:
3DxYARuZFpZsPEpWee1RSjuPMFktj2sy7y

janiceilene:
👋 @jccwm That set of docs is specifically focused on Enterprise Server Admin, it's not all of the docs for endusers on GHES. For those docs, you can navigate to almost any page and use the picker in the top right corner.

image

So, from https://docs.github.com/en/enterprise-server@2.22/github you can navigate to all of the end user docs for 2.22. Feel free to ping me if I can answer any other questions or if a walk through of anything would be helpful!

PR #945, opened 2020-10-31
Reorder sections

tmthyln:
The sections should be reordered; this order makes more sense for how the section titles read, as well presenting first the more likely scenario for beginners (or even if they made more commits, the title makes it clear they should check the next section).

### Why:

The new section order makes more sense in reading. Now readers will first see the section "Removing a file added in the most recent unpushed commit" (which is probably more common and actually has an in-depth answer on this page) and then see the "Removing a file that was added in an earlier commit" (so now the word "earlier" makes sense, if the first section didn't apply.

### What's being changed:

Just the ordering of the 2 sections is changed. The content is identical.

### Check off the following:


tmthyln (commented):
I don't think there should be any issues; it's a relatively minor change but will make the page a little more clear.

janiceilene:
Thanks so much for opening a PR @tmthyln! I'll get this triaged for review 💖

rachmari (approved):
@tmthyln this updated ordering makes sense to me! Thank you so much for taking the time to contribute to the docs. 🙏 💚 I'll get this merge right away! 🚢

PR #1114, opened 2020-11-06
GHAE Enterprise Support Page updates

briceholland:
Some tweaks to the verbiage around what is provided with Premium Support and GHAE.

Instead of the "Government notification box" indicating the TSAM will inform them of the difference, I propose a secondary article, similar to how we've setup Premium Support, with something like:
For more information, see "About GitHub AE Support for Commercial and Government."
This new page should house all of the roles/responsibilities for each of the two respective services in a table format that's clear and defines the support offering better.

Also, while reviewing, this seems like it should be designated with "Cloud" to denote GHES and GHE**C** better:
https://docs.github.com/en/enterprise-server@2.22/admin/enterprise-support/about-github-premium-support-for-github-enterprise
to "About GitHub Premium Support for GitHub Enterprise Cloud"

So we ultimately have:
About GitHub Premium Support for GitHub Enterprise Server
About GitHub Premium Support for GitHub Enterprise Cloud
About GitHub AE Support for Commercial and Government

The navigation would simply show the Commercial/Government Cloud page under "About GitHub Enterprise Support" in the navigation so the customers can see the differences in the two and we can better explain what we provide. Both of the "About Premium Support" pages expand on a lot of good information about Premium Support.

### Why:

- GitHub AE differences better defined and match the cadence we set with customers
- TSAMs called out more effectively on the bullet points, because it's a BIG selling point (Premium doesn't guarantee a TSAM, but Premium Plus and GitHub AE does)
- Premium content not really offered by default for GHAE, however quarterly support reviews are.
- We manage the environment in GHAE altogether, whereas in Premium Support we offer Managed Admin hours.

### What's being changed:

https://docs.github.com/en/github-ae@latest/admin/enterprise-support/about-github-enterprise-support

### Check off the following:


briceholland:
Still going through the recently added pages... just need to update the Severity variables and potentially draft the new page if I feel confident enough. :)

content/admin/enterprise-support/about-github-enterprise-support.md
```diff
@@ -28,9 +28,16 @@ In addition to all of the benefits of {% data variables.contact.enterprise_suppo
...
   - Phone support 24 hours per day, 7 days per week
   - A{% if currentVersion == "github-ae@latest" %}n enhanced{% endif %} Service Level Agreement (SLA) {% if enterpriseServerVersions contains currentVersion %}with guaranteed initial response times{% endif %}
-  - Access to premium content{% if enterpriseServerVersions contains currentVersion %}
-  - Scheduled health checks{% endif %}
-  - Managed services
+{% if currentVersion == "github-ae@latest" %}
+  - An assigned Technical Service Account Manager
+  - Quarterly support reviews
+  - Managed Admin services
+{% if enterpriseServerVersions contains currentVersion %}
```
nschonni, line 35:
```suggestion
{% else if enterpriseServerVersions contains currentVersion %}
```

megbird:
👋  @briceholland thanks for bringing all of this into a PR! Let's move this work to an internal PR to work on with some other changes 😄

megbird (approved):
These changes look good to me! The only question I have is if "Managed Admin hours" and "Managed Admin services" are different offerings? If not, we can simplify the versioning a bit, but I trust that they are and this is correct 😄

PR #1132, opened 2020-11-06
Update workflow-syntax-for-github-actions.md

### LINKED ISSUE #825: Fix miscellaneous errors in GitHub Actions / Reference / Workflow syntax

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions

1. Under the section `jobs.<job_id>.steps.run` in the subsection [Exit codes and error action preference](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference): Under the bullet point for `cmd`, the second item starts out "cmd.exe will exit with the error level of the last program it executed, and it will and return the error code to the runner."  The second "and" is extraneous and should be deleted.

2. The last sentence in the section [`jobs.<job_id>.steps.with.entrypoint`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepswithentrypoint) reads "The entrypoint keyword is meant to use with Docker container actions, but you can also use it with JavaScript actions that don't define any inputs."  "to use" should be changed to "to be used".

3. This is a general comment about the _syntax_ of the syntax article: Since, for the most part, the YAML used in workflow files can be considered a form of JSON, it indeed makes some sense to use JavaScript-like syntax to refer to the parts of the YAML file.  For example, `jobs.<job_id>.timeout-minutes` clearly refers to the value corresponding to the `timeout-minutes` key in the hash value corresponding to the `<job_id>` key in the hash value corresponding to the `jobs` key, and this is precisely how you would refer to it in JavaScript.  Where it gets confusing is in notations such as `jobs.<job_id>.steps.timeout-minutes`.  This is because the value corresponding to the `steps` key is not another hash but rather an array (of steps).  Using the dot notation obscures the fact that in this case, `timeout-minutes` applies to a particular step and doesn't apply to the set of steps as a whole.  It would be clearer to write something like `jobs.<job_id>.steps[n].timeout-minutes` to make clear that `timeout-minutes` applies only to some particular step `n` and not to all steps.  The `[n]` here stands in for any step in the sequence analogously to the way `<job_id>` stands in for some particular key in the `jobs` hash.

JBryceOrtiz:
issue #825.  Under the section jobs.<job_id>.steps.run in the subsection Exit codes and error action preference: Under the bullet point for cmd, the second item starts out "cmd.exe will exit with the error level of the last program it executed, and it will and return the error code to the runner." The second "and" was deleted because it was extraneous.

The last sentence in the section jobs.<job_id>.steps.with.entrypoint read "The entrypoint keyword is meant to use with Docker container actions, but you can also use it with JavaScript actions that don't define any inputs." "to use" was changed to "to be used".

### Why:
Under the section jobs.<job_id>.steps.run in the subsection Exit codes and error action preference: Under the bullet point for cmd, the second item starts out "cmd.exe will exit with the error level of the last program it executed, and it will and return the error code to the runner." The second "and" was extraneous and was deleted. 

The last sentence in the section jobs.<job_id>.steps.with.entrypoint reads "The entrypoint keyword is meant to use with Docker container actions, but you can also use it with JavaScript actions that don't define any inputs." "to use" was changed to "to be used".

### What's being changed:

Under the section jobs.<job_id>.steps.run in the subsection Exit codes and error action preference: Under the bullet point for cmd, the second item starts out "cmd.exe will exit with the error level of the last program it executed, and it will and return the error code to the runner." The second "and" was deleted.

The last sentence in the section jobs.<job_id>.steps.with.entrypoint reads "The entrypoint keyword is meant to use with Docker container actions, but you can also use it with JavaScript actions that don't define any inputs." "to use" was changed to "to be used".

### Check off the following:
- [ X] All of the tests are passing.
- [ X] I have reviewed my changes in staging.
- [ X] For content changes, I have reviewed the [localization checklist](https://github.com/github/docs/blob/main/contributing/localization-checklist.md)
- [ X] For content changes, I have reviewed the [Content style guide for GitHub Docs](https://github.com/github/docs/blob/main/contributing/content-style-guide.md).


janiceilene:
Thanks so much for opening a PR @JBryceOrtiz 🙌 I'll get this triaged for review!

lucascosti (approved):
Thanks for this fix, @JBryceOrtiz! I'll merge this in when the tests pass 👍

PR #1189, opened 2020-11-09
SSH file to "id_ed25519.pub"

servel333:
### Why:

It looks like the filename was changed in the "generating a new ssh key" document, and I figure this should be updated as well.

Ref: https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

### What's being changed:

```diff
- pbcopy &lt; ~/.ssh/id_rsa.pub
- Copies the contents of the id_rsa.pub file to your clipboard
+ pbcopy &lt; ~/.ssh/id_ed25519.pub
+ Copies the contents of the id_ed25519.pub file to your clipboard
```

### Check off the following:


janiceilene:
This is not stale! Sorry about missing this one @servel333 💛 I'll get it triaged for review.

shati-patel (approved):
Thank you for the changes @servel333! 🎉 

PS: The "Files changed" tab shows 0, but this is because an identical change was merged just before. Your contribution is still 100% valid!

servel333:
It's worth noting that the instructions mention using both an RSA and the other one. These two docs may need further updates in order to resolve the issue of different key names on different systems, as well as addressing legacy keys that users already use.

shati-patel:
> It's worth noting that the instructions mention using both an RSA and the other one. These two docs may need further updates in order to resolve the issue of different key names on different systems, as well as addressing legacy keys that users already use.

Thank you for pointing that out 👍🏽 There have been quite a few changes to these docs lately, so we're slowly getting them into a better shape. Feel free to open an issue if you see any immediate problems and places that might be confusing!

Thanks again for taking the time to help us improve the docs, we appreciate the help 😊

PR #1115, opened 2020-11-06
chore: flatten ifs for fkirc/skip-duplicate-actions

nschonni:
### Why:

Each of the sub-steps except on that only applies when the previous steps fails are skipped if the "should skip" job evaluates to `true`. It was figured out with the js-lint job that this "Skipped" jobs still allow merging of the "Required" jobs.

### What's being changed:

Hoist the `if` condition for the test jobs to the root, and remove the duplicate checks `if`s on the sub-steps

### Check off the following:
- [x] All of the tests are passing.
- [x] I have reviewed my changes in staging.
- [x] For content changes, I have reviewed the [localization checklist](https://github.com/github/docs/blob/main/contributing/localization-checklist.md)
- [x] For content changes, I have reviewed the [Content style guide for GitHub Docs](https://github.com/github/docs/blob/main/contributing/content-style-guide.md).

PR #1388, opened 2020-11-16
Revert "chore: flatten ifs for fkirc/skip-duplicate-actions"

chiedo:
This reverts commit d9d17364e334fb33974836a44aea8fe7702fc381 ([this PR](https://github.com/github/docs/pull/1115)).

Over the past few days we've been running into this stall with our pull requests.

Screen Shot 2020-11-16 at 9 08 36 AM

I can't explain why but I think it's because of [this PR](https://github.com/github/docs/pull/1115) so I want to revert it and see what happens.

This is what the test suite looks like now with the fix on this PR which seems to suggest that this does fix the problem. Furthermore, the test suite on this PR is running while it isn't on others that were opened roughly 15 minutes before this PR.

Screen Shot 2020-11-16 at 9 14 09 AM

### Check off the following:

/cc @nschonni @github/docs-engineering


JasonEtco (approved):

nschonni:
No worries, maybe something to flag internally.
I guess before it always waits on the `needs`, but the job level `if` is somehow causing it to delay reaction instead of starting the job then querying for the `if` on the steps

Issue #253, opened 2020-10-08
Shell not indicated in some examples, missing alternative shell solutions


Simran-B:
### What article on docs.github.com is affected?

For instance, [Actions - Workflow commands: Setting an environment variable](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) does not say that the examples are bash specific. They won't work as-is in cmd or powershell on Windows.

### What part(s) of the article would you like to see updated?

- Tell the user what shell an example is for if it's not compatible with all [standard terminals](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell)
- Ideally provide examples for other shells as well

### Additional information

In [Workflow syntax for GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-running-a-script-using-bash) there is actually a great example of how it should be:

![Examples which indicate intended shell](https://user-images.githubusercontent.com/7819991/95436524-353c9e00-0954-11eb-9019-7f315db758ba.png)

Simran-B:
On second thought, I think it would make more sense to have tabs similar to the platform specific content, but render the tabs for each example group instead of once at the top of the page:

![image](https://user-images.githubusercontent.com/7819991/95475591-a9903500-0986-11eb-9d0d-ecd0ae236f97.png)

Changing the tab should be reflected by all other shell-specific examples. I'll try to implement a prototype. I think it will require nested Liquid tags? Let me know if you can think of another way.

```
### About workflow commands

{% shelltabs %}

{% bash %}

`echo "{name}={value}" >> $GITHUB_ENV`

{% endbash %}

{% pwsh %}

`echo "{name}={value}" | Out-File -Append -FilePath $env:GITHUB_ENV -Encoding utf8`

{% endpwsh %}

{% cmd %}

`chcp 65001 > nul & echo {name}={value} >> %GITHUB_ENV%`

{% endcmd %}

{% endshelltabs %}
```

Simran-B:
Hard-coding it to shells might actually allow this to be implemented without nesting tags, but I'm in favor of a generic tab widget solution. I believe that it will require a group name to correctly update related example groups, and to avoid switching tabs of unrelated groups with an equally named tab. Something along:

```
{% tabs "shell" %}
{% tab "Bash" %} ... {% endtab %}
[...]

PR #1606, opened 2020-11-25
Tab widget (prototype)

Simran-B:
See https://github.com/github/docs/issues/253

`npm run dev` to start server, then open `http://localhost:4000/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions` in browser.


himanshu211raj:
@Simran-B i didn't understand what you mean to say? As i am a beginner I don't know much about open source.

Simran-B:
@himanshu211raj Do you know JavaScript? If yes, then you can help me here. If no, then you can still help to add more examples (for example PowerShell scripts where currently only Bash examples are given). Or do you want to help with something else?

himanshu211raj:
> @himanshu211raj Do you know JavaScript? If yes, then you can help me here. If no, then you can still help to add more examples (for example PowerShell scripts where currently only Bash examples are given). Or do you want to help with something else?

Where i can help in adding more examples and in which other domains i will make contributions and how?

Simran-B:
@himanshu211raj The "Setting an environment variable" example only works in a Bash shell for instance:

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

You could add additional examples for PowerShell Core and Windows `cmd`. Here is an example how it should look:

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-running-a-script-using-bash

If you are not familiar with PowerShell and `cmd`, then try to find something else that can be improved. You can find general information on how to contribute here:

https://github.com/github/docs/blob/main/CONTRIBUTING.md#contributing-to-this-repository-

janiceilene:
Thanks so much for opening a PR @Simran-B! I'll get this triaged for review 💖

Simran-B:
@janiceilene Thanks, but this PR is still in draft state, i.e. too early for a review. I wanted to share my work in progress code so that others have a chance to help. I would appreciate early feedback nonetheless.

There are a couple of issues at the moment:
- Tab groups aren't independent and tabs are currently switched based on name regardless of the group they belong to
- The tab content isn't initialized (hidden) as desired
- Clicking tabs lets the content jump (undesired scrolling), except for the first tabs instance (does preventDefault() help perhaps?)
- Markdown support in tab titles is a bit hacky
- Extended markdown might not work as tab content, but haven't tried

janiceilene:
Thanks for the clarification @Simran-B! I moved it over to where the engineering team can take a look with that additional context 💖

Simran-B:
Notes to self:
- Implemented compensation for content shift when switching tabs that keeps the on-screen distance between the clicked tab and the top of the viewport constant.
- Add group information to content, not just tab header, for easier selection?
- Use classes or data attributes to indicate tab metadata? (browser support, performance)
- Allow deep linking with anchors
  - Tab link should be copyable
  - Anchor link should activate the tab on load
  - Clicking tabs should potentially set hash in address bar (add to window.history), despite preventDefault()
- A11y? (tab index, ARIA roles, cursor support?)
- Should nested tabs or too many tabs (how many?) be disallowed explicitly?
- Should content between tab group tag and first tab tag raise an error?
- Adjust CSS class for tabs based on number of tabs? `<nav class="UnderlineNav my-3">` (`my-<amount>`)
- Do extended markdown features only work inside of blocks which derive from the extended markdown tag?
- Should probably render initial tab state into HTML
- Remember preferred tab per group, using local storage?
- Any frontmatter customization options that would be useful? Like `defaultPlatform`, but how to do that (esp. in frontmatter schema) if groups and tabs are defined per page?
- Tests to potentially add:
  - Tabs are rendered, only one tab active
  - Correct activation of initial tab for each group and the corresponding content
  - No jumping caused by switching tabs where above tab content would normally push the content down

Legionsixxx9 (approved):
@Simran-B

Simran-B:
Still experimenting with this locally, making it cleaner and more robust.

Note to self: check what causes the unwanted Markdown processing of the HTML template and find a way to avoid it.

tashaemery (approved):

crichID:
> Still experimenting with this locally, making it cleaner and more robust.

@Simran-B thank you for continuing to experiment with this! I'm following along and looking forward to using this new widget in the docs.

lib/liquid-tags/tabs.js
```diff
@@ -0,0 +1,121 @@
...
+
+  async render (context) {
+
+    // TODO: push a dummy as first node, then update it here? (but from which callback?)
+    // TODO: 4+ leading spaces in tabs.html = markdown code block, how to avoid? Use '{% include tabs %}' as inline template?
+    const src = await this.template.engine.fileSystem.readTemplateFile(this.name())
+    this.tabs.forEach(async tab => {
+      // Could use textOnly mode to strip out markup, but slugger does that and more, see pushBlock()
+      //await renderContent(tab.title, null, { textOnly: true })
+      tab.title = (await renderContent(tab.title, null)).slice(3, -4) // HACK: remove surrounding <p>
```
vanessayuenn, line 111: 👍🏼 on slugger usage here! It's a good idea to support markdown in the tab header, but I think if we do, we should also support parsing variables/reusables here as well to be consistent. `context` will need to be passed in for the variables/reusables to be rendered properly. We could rearrange the order of calling these methods a bit:
```suggestion
      tab.title = slugger.slug(await renderContent(tab.title, context))
```

vanessayuenn (commented):
@Simran-B A general-usage tab component has been on our radar for a while, so thank you for making it happen! This is looking great so far! 

In terms of implementation, I know this is still early stage, but I am curious if you have considered a nested block approach, instead of the current liquid tag within a liquid block approach? I think this will result in modular and more maintainable code. From a usability perspective, a nested liquid block usage might be more user-friendly as well:

```
{% tabs "shell" %}

  {% tab "Bash" %}
    bash stuff
  {% endtab %}

  {% tab "PowerShell" %}
    powershell stuff
  {% endtab %}

{% endtabs %}
```

I am happy to elaborate some more if it's helpful. Please let me know how I can support you to get this to the finish line!

Simran-B:
Hi @vanessayuenn, thanks for the feedback. I missed your comment among all the notifications (and there are frequent spam comments in this repo unfortunately). Sorry for getting back so late.

> I am curious if you have considered a nested block approach, instead of the current liquid tag within a liquid block approach?

I did, but the markup looks somewhat cleaner with tags and it (mostly) solves a particular problem: What to do with content between tab blocks? For example:

```
{% tabs "shell" %}

-- What to do with this text? --

  {% tab "Bash" %}
    bash stuff
  {% endtab %}

-- And this? --

  {% tab "PowerShell" %}
    powershell stuff
  {% endtab %}

-- And this? --

{% endtabs %}
```

With tabs as tags, only one corner case remains:

```
{% tabs "shell" %}

-- What to do with this text? --

  {% tab "Bash" %}
    bash stuff

  {% tab "PowerShell" %}
    powershell stuff

{% endtabs %}
```

From a usability point of view and especially with larger amounts of tabbed content, I can see how tab blocks would be better however. I'll see if I can intercept the callbacks for the content to throw an error if there's something outside of tab blocks.

Simran-B:
@vanessayuenn Regarding nested blocks, are there any existing examples?

Registering both `tabs` and `tab` doesn't seem right, as the latter is only valid within the former. All state should be stored in the Tabs class, but I'm not sure how one would do that with two independent block types (wild guess: use the parent environment?). Maybe the current implementation could be kept, but adjusted to also expect `endtab` in unknownTag? Or is there a way to register the tab block within the tabs block only?

If there's no built-in functionality for nested block handling, then I'm not sure how to progress. The parser code is pretty daunting.

luca11221:
> @Simran-B A general-usage tab component has been on our radar for a while, so thank you for making it happen! This is looking great so far!
> 
> In terms of implementation, I know this is still early stage, but I am curious if you have considered a nested block approach, instead of the current liquid tag within a liquid block approach? I think this will result in modular and more maintainable code. From a usability perspective, a nested liquid block usage might be more user-friendly as well:
> 
> ```
> {% tabs "shell" %}
> 
>   {% tab "Bash" %}
>     bash stuff
>   {% endtab %}
> 
>   {% tab "PowerShell" %}
>     powershell stuff
>   {% endtab %}
> 
> {% endtabs %}
> ```
> 
> I am happy to elaborate some more if it's helpful. Please let me know how I can support you to get this to the finish line!



> This PR is stale because it has been open 7 days with no activity and will be automatically closed in 3 days. To keep this PR open, update the PR by adding a comment or pushing a commit.

heiskr:
Closing for now, feel free to reopen if you'd like to continue this effort.

chiedo:
@Simran-B sorry we've taken so long to get back to you on this. 🙇🏿  We're not going to move forward in this direction but we appreciate your contribution! Thanks for the time and effort you invested in this PR.

Simran-B:
@chiedo No problem. I'd still like to know if it's a general decision against tabs and whether you have an alternative in mind. I'll probably work on this further as an exercise when I find the time, as I need such a widget myself (for Jekyll, but I'm more comfortable with JavaScript than Ruby, so this is great for experimentation especially because the Liquid renderer is basically a port).

chiedo:
That sounds good @Simran-B! @myarb can speak towards the product vision on tabs vs not having tabs. But from an engineering perspective I do know that we'll be working on changes that would influence how this solution was implemented.

myarb:
👋🏼  @Simran-B we aren't currently planning to expand our tabs in docs, but this is helpful feedback for future docs changes. If you're still interested in working on the problem stated in https://github.com/github/docs/issues/253, adding Windows-specific examples in areas where they're missing in a PR would be a welcome addition as Lucas mentioned in this [comment](https://github.com/github/docs/issues/253#issuecomment-745037613) 💜  Thanks so much for your contributions and passion for GitHub Docs!

Issue #1722, opened 2020-12-01
jumping content


Arhell:
### What is the current behavior?

![1](https://user-images.githubusercontent.com/26163841/100774165-78723500-340a-11eb-8657-55a324a2e732.jpg)
on open & close

### What changes are you suggesting?

change some css class

### Additional information
I will do it soon

PR #1732, opened 2020-12-01
fix dropdown on mobile

Arhell:
### Why:
fix issue https://github.com/github/docs/issues/1722

### What's being changed:

Content won't jump


Arhell:
cc @chiedo

chiedo:
@Arhell can you explain the issue you're describing one more time? I'm not sure I understand. Also can you include before and after screenshots (maybe animated gifs if that will help)? Thanks!

Arhell:
@chiedo 
Of course, when on mobile you click the dropdown then the content jumps
normal state

Before

content jumped down
after fix

the content is in place and the dropdown works correctly as it does in large resolutions without interfering with the content

chiedo:
Thank you! That makes perfect sense. Thanks for the fix!

chiedo (approved):

Issue #1093, opened 2020-11-05
Use real examples in API code examples


jidanni:
Please change
```
    "last_read_at":"last_read_at"
```
to
```
   "last_read_at":"2014-11-07T22:01:45Z"
```

They occur in
$ w3m -dump https://docs.github.com/en/free-pro-team@latest/rest/reference/activity | grep -n last_read_at.*:.*last_read_at
```
973:  -d '{"last_read_at":"last_read_at"}'
978:  last_read_at: 'last_read_at'
1524:  -d '{"last_read_at":"last_read_at"}'
1531:  last_read_at: 'last_read_at'
```

However they are not accessible from
https://github.com/github/docs/edit/main/content/rest/reference/activity.md
for me to edit. Nor does Search find what file they are in. Therefore
please make the edit for me. Thanks.

janiceilene:
Thanks for opening an issue @jidanni! I believe that information is autogenerated 🤔 but I'll triage this for someone with more knowledge on our rest documentation 💛

lucascosti:
👋 @jidanni.

Our API documentation is autogenerated from the API code, which is why you can't find it in the source in this docs repo.

> Please change
> 
> ```
>     "last_read_at":"last_read_at"
> ```

I assume you're referring to the code examples? e.g [here](https://docs.github.com/en/free-pro-team@latest/rest/reference/activity#mark-notifications-as-read--code-samples)

These are also autogenerated and at the moment just have placeholder values. We do have plans to generate these with proper examples rather than the placeholders you see.

I'll rename this issue and tag in docs engineering too 🙂

jidanni:
> I assume you're referring to...

Yes.

I don't see why impossible/non-working code sample are used, when [directly below it](https://docs.github.com/en/free-pro-team@latest/rest/reference/activity#get-a-thread--code-samples) real/working ones are used.

lucascosti:
> I don't see why impossible/non-working code sample are used, when [directly below it](https://docs.github.com/en/free-pro-team@latest/rest/reference/activity#get-a-thread--code-samples) real/working ones are used.

For integers in examples, it always uses `42` as the number in the code samples. For strings, our generator just defaults to just using the name of the parameter as the example string.

We have plans to improve this to use real examples generated from our schema, so stay tuned! 🙂

umutduyum:
GET /repos/{owner}/{repo}/actions/artifacts

Aggrord09:
Nice

[...]

PR #1803, opened 2020-12-04
Actions: env context cannot be used in job.<job_id>.if

Simran-B:
### Why:

The following raises the error `Unrecognized named-value: ‘env’`:

```yaml
env:
  VAR: foo
job:
  <job_id>:
    if: env.VAR == 'foo'
```

However, according to the [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#env) it should be possible (at least that is one interpretation):

> A map of environment variables that are available **to all jobs and steps in the workflow**.

The following statement about the [env context](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#env-context) conflicts with the above:

> You can only use the env context in the value of the with and name keys, or **in a step’s if conditional**.

It can apparently not be used in a job's if conditional.

Also see: https://github.community/t/how-to-use-workflow-env-variables-in-job-ifs/147046

### What's being changed:

Change the former statement to "the steps of all jobs" to exclude the jobs themselves. However, should the env context be usable at job level in another key but `if` then this change will also cause confusion - is it usable in any of the job level keys such as `jobs.<job_id>.name`, or does the second statement refer to the name key of steps only?

### Check off the following:


Simran-B:
Not sure if this statement should be revised as well:

> You can also set environment variables that are only available to a job or step.

janiceilene:
Thanks for opening a PR @Simran-B! I'll get this triaged for review ✨

janiceilene:
Not stale! Thanks for your patience @Simran-B 💖

fatdagreat78 (approved):
#1093

skedwards88 (approved):
🚀 @Simran-B Thanks for clarifying this wording! Edit: I made one small suggestion. Let me know what you think, then I'll get this merged for you.

content/actions/reference/workflow-syntax-for-github-actions.md
```diff
@@ -187,7 +187,7 @@ For more information about cron syntax, see "[Events that trigger workflows](/ac
 
 ### `env`
 
-A `map` of environment variables that are available to all jobs and steps in the workflow. You can also set environment variables that are only available to a job or step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps.env`](#jobsjob_idstepsenv).
+A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to a job or step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps.env`](#jobsjob_idstepsenv).
```
skedwards88, line 190: What do you think of changing the wording in the next sentence as well to have this same clarification 

```suggestion
A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps.env`](#jobsjob_idstepsenv).
```

Simran-B:
@skedwards88 Thanks, I rebased the PR and applied your suggestion.

I took a bit of time to test where the env context is actually available, and there's definitely more to change in the docs:

> You can only use the env context in the value of the `with` and `name` keys, or in a step's `if` conditional. ([link](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#env-context))

This sentence is ambiguous. One interpretation is that the env context is available for all keys called `with` or `name`, as well as a step's `if` key, i.e. the following paths:
- `name`
- `jobs.<job_id>.name`
- `jobs.<job_id>.steps[*].name`
- `jobs.<job_id>.steps[*].with`
- `jobs.<job_id>.steps[*].if`

It's not available for the former two in fact, so the correct interpretation is that the env context is available for the aforementioned keys **in steps only**. This should be made clearer.

There's also a slight problem with `jobs.<job_id>.steps[*].with`: you don't use `${{ env.VAR }}` as value under that key like `with: ${{ env.VAR }}`, but rather as value for a mapping under this path, for instance:

```yaml
with:
  some-name: ${{ env.VAR }}
```

I find this problematic in particular because there are two special keys, `jobs.<job_id>.steps[*].with.args` and `jobs.<job_id>.steps[*].with.entrypoint`. I haven't tried to use the env context under them yet, but maybe it doesn't work for them. If the statement about the `with` key actually refers to nested values, then it would include them too, which might be incorrect.

I also found out that there are at least two more keys in which the env context can be used:
- `jobs.<job_id>.steps[*].run`
- `jobs.<job_id>.steps[*].env` (actually as nested value again, as part of a mapping)

```yaml
    steps:
      - name: env context in run
        run: echo ${{ env.VAR }} # works

      - name: env context in action
        uses: actions/github-script@v3
        env:
          my_var: ${{ env.VAR }} # works
        with:
          script: |
            console.log('process.env.my_var: ' + process.env.my_var)
```

It can definitely not be used for:
- `jobs.<job_id>.steps[*].id`
- `jobs.<job_id>.steps[*].uses` (cannot be dynamic)

I still have to test the following:
- jobs.<job_id>.steps[*].with.args
- jobs.<job_id>.steps[*].with.entrypoint
- jobs.<job_id>.steps[*].continue-on-error
- jobs.<job_id>.steps[*].timeout-minutes

Simran-B:
The env context is also available for:
- `jobs.<job_id>.steps[*].with.args`
- `jobs.<job_id>.steps[*].with.entrypoint`
- `jobs.<job_id>.steps[*].continue-on-error` (needs to be cast to boolean)
- `jobs.<job_id>.steps[*].timeout-minutes` (needs to be cast to number)

Simran-B:
@skedwards88 Please take another look. I made it clearer that the env context is only available in steps and added a list of step keys which support it.

content/actions/reference/context-and-expression-syntax-for-github-actions.md
```diff
@@ -109,13 +109,22 @@ The `env` context contains environment variables that have been set in a workflo
...
+
+- `if`
+- `name`
+- `run`
+- `with` (including `with.args` and `with.entrypoint`)
+- `env`
+- `continue-on-error` (required conversion from string to boolean)
+- `timeout-minutes` (requires conversion from string to number)
+
+ For more information on the step syntax, see "[Workflow syntax for {% data variables.product.prodname_actions %}](/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idsteps)."
```
skedwards88, line 122: Thanks for testing this! I really respect the work you put into validating this. ✨ 

If new keys are added in the future, it seems that the default behavior is to allow it to use the `env` context. To make it easier to keep the docs up-to-date, I recommend documenting the keys that won't accept the `env` context instead of the keys that do accept it. To do that, you could add this to line 110, after "The `env` context syntax allows you to use the value of an environment variable in your workflow file.":

> You can use the `env` context in the value of any key of a step except for the `id`, `uses`, or `env` keys. For more information on the step syntax, see "[Workflow syntax for {% data variables.product.prodname_actions %}](/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idsteps)."

And then can remove lines 112-122.

What do you think?

Simran-B: Yeah, it's probably better this way around from a maintenance perspective and to keep it succinct. Adjusted the PR as suggested, with some slight modifications.

This is my test workflow BTW:

```yaml
env:
  JOB_NAME: Name defined at job level
  JOB_OS: windows-latest
  JOB_ID: foo
  JOB_ACTION: actions/github-script@v3
  JOB_TIMEOUT: 1
  JOB_BOOL: true
  JOB_WITH: |
    { "script": "console.log('with block supplied as JSON string')" }
  JOB_STEP: |
    {"name": "env as step", "run": "echo Entire step defined as string"}
  JOB_ENTRY: /custom.sh
  JOB_ARGS: FirstArgument SecondArgument?

#name: ${{ env.JOB_NAME }} # error
name: Env context

on:
  push:
    branches:
      - main

jobs:
  job-level:
    #name: ${{ env.JOB_NAME }} # error
    name: Job 1
    #runs-on: ${{ env.JOB_NAME }} # error
    runs-on: ubuntu-latest
    steps:
      - name: env context in run
        run: echo ${{ env.JOB_NAME }} # works
      - name: env var in action
        uses: actions/github-script@v3
        with:
          script: |
            console.log('process.env.JOB_OS: ' + process.env.JOB_OS)
          # works, set via job level env
      - name: env context in action
        uses: actions/github-script@v3
        env:
          my_os: ${{ env.JOB_OS }} # works
        with:
          script: |
            console.log('process.env.my_os: ' + process.env.my_os)
      - name: env in id (set output)
        #id: ${{ env.JOB_ID }} # error
        id: foo
        run: |
          echo "::set-output name=result::bar"
      - name: env in id (read output)
        run: |
          echo "result = ${{ steps[env.JOB_ID].outputs.result }}"
      - name: env in uses
        #uses: ${{ env.JOB_ACTION }} # error
        uses: actions/github-script@v3
        with:
          script: console.log("${{ env.JOB_ACTION }}") # works
      - name: env in with
        uses: actions/github-script@v3
        with: ${{ fromJSON(env.JOB_WITH) }} # works
      #- ${{ fromJSON(env.JOB_STEP) }} # error
      - name: env in with.args
        uses: actions/hello-world-docker-action@v1
        with:
          entrypoint: /bin/echo
          args: ${{ env.JOB_ARGS }} # works
      - name: env in with.entrypoint
        uses: actions/hello-world-docker-action@v1
        continue-on-error: true
        with:
          who-to-greet: 'Mona the Octocat'
          entrypoint: ${{ env.JOB_ENTRY }} # works
      - name: env in timeout
        continue-on-error: true
        timeout-minutes: ${{ fromJSON(env.JOB_TIMEOUT) }} # works
        run: |
          echo sleeping for 2 minutes
          sleep 120
          echo done
      - name: env in continue
        continue-on-error: ${{ !!env.JOB_BOOL }} # works
        run: |
          exit 1
```

skedwards88 (approved):
🚀 Awesome work! I will get this merged down for you.

Issue #136, opened 2020-05-20
Default the configuration page for self-hosted runners to Linux


jenschelkopf:
### Issue overview

The 'Configuring the self-hosted runner application as a service' page has tabs for specific instructions for Mac, Windows, and Linux.

Currently, the default tab is Mac.

Let's change the default to Linux, as this is the most common for our users.

![image](https://user-images.githubusercontent.com/1865328/82494939-f5ae0c00-9aaf-11ea-93a0-80dd439160ac.png)

### Products affected by this issue

_to be completed by the core product docs team first responder_

- [x] Dotcom (all products)
  - [ ] GitHub Free
  - [ ] GitHub Pro
  - [ ] GitHub Team
- [ ] GitHub Enterprise Cloud
- [ ] GitHub Enterprise Server (Versions: )
- [ ] API
- [ ] Desktop

_*Note:* The doctocat who triages this issue may invite you to open a PR to address it. Doing so is absolutely not required, though it's helpful for a speedy fix! Someone from our team will work with you to take your changes across the finish line._

/cc @github/product-docs-core

Simran-B:
> it looks like Linux does appear first in the content

What determines the order of the tabs? If it's not the content, and if it's also not sorted alphabetically (it's Mac, Windows, Linux in above screenshot), then how does the current tab order come about?

zeke:
> What determines the order of the tabs? 

Here's the HTML for the tabs: https://github.com/github/docs/blob/5395b9371056e7c8eb9b649e5c1de4de80613155/includes/article.html#L31:L40

Here's the JavaScript file that dynamically updates the tabs and the displayed content on page load: https://github.com/github/docs/blob/main/javascripts/display-platform-specific-content.js

Here's the file where tests should be added for any changes to the existing platform-switching behavior: https://github.com/github/docs/blob/main/tests/browser/browser.js (Note that the platform-switching feature was added before we added this Puppeteer headless browser testing capability, so there are no existing browser tests for the current platform-switching functionality)

I hope that helps, @Simran-B!

Simran-B:
Thanks @zeke for all the information! I created a PR: #1917

> Currently, the default tab is Mac.

This isn't the case (anymore?), the default tab is based on the user's operating system, which falls back to Mac if it's none of the supported platforms.

jenschelkopf:
IMO, the tab should not default to the user's OS. The user's OS is not related to the user's runner type. Linux is by far the most common runner type, so we should default to Linux always.

[...]

PR #1917, opened 2020-12-09
Add support for page.defaultPlatform frontmatter

Simran-B:
### Why:

[Configuring the self-hosted runner application as a service](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service) should default to Linux instead of the user OS / macOS, see #136

### What's being changed:

- Support for optional `defaultPlatform` frontmatter
- Set `defaultPlatform: linux` for above page
- Add some browser and unit tests for platform switcher
- Fix `npm run browser-test` for Windows

### Check off the following:
- [x] All of the (relevant) tests are passing.
- [x] I have reviewed my changes in staging. (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**) - https://docs-1917--default-platform.herokuapp.com/en/free-pro-team@latest/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service
- [x] Add more tests (should not have platform switcher, actually shows/hides platform-specific content)

Questions:
- Allow `all` as defaultPlatform value?
- Is `<nav data-default-platform="{{ page.defaultPlatform }}">` the way to go to get info from frontmatter to client-side JS?


janiceilene:
Thanks so much for opening a PR for this update @Simran-B! I'll get this triaged for the engineering team to take a look 👀

lib/frontmatter.js
```diff
@@ -100,6 +100,11 @@ const schema = {
     },
     interactive: {
       type: 'boolean'
+    },
+    // Platform-specific content preference
+    defaultPlatform: {
+      type: 'string',
+      enum: ['mac', 'windows', 'linux'] // TODO: add 'all'?
```
heiskr, line 107: How would 'all' work?

tests/browser/browser.js
```diff
@@ -168,3 +168,45 @@ async function getLocationObject (page) {
   })
   return JSON.parse(location)
 }
+
+describe('platform specific content', () => {
```
heiskr, line 172: I think these tests would be faster under content or rendering tests, using Cheerio instead of puppeteer.

Simran-B: The tests aren't that slow compared to other ones, but I can take a look if I can port this over.

lib/frontmatter.js
Simran-B: I have that question too. I'm not sure how it is intended to work, it's also not used anywhere, but described here: https://github.com/github/docs/blob/main/contributing/content-markup-reference.md#operating-system-tags

JasonEtco: I think we can omit it because:

* The equivalent could be to just not include `defaultPlatform` in the frontmatter, and existing behavior will prevail
* We don't... actually... use `{% all %}` anywhere in the codebase 🙃

Simran-B: Yeah, I think it can be left out. The tag is a bit of a misnomer, it doesn't mean "all platforms" but rather "all other platforms" or "another unspecified platform". If something actually applies to all platforms, then you would simply not wrap that content in any of the platform tags. I don't see how `{% all %}` is useful to be honest and would even go as far as to remove this tag/block entirely.

tests/browser/browser.js
Simran-B: Upon review, I don't think that there is a way to test the user agent detection / overriding code with Cheerio. The only test that can be ported IMO is `should have a platform switcher`, which is purely rendering-based.

heiskr (approved):
🚢

tests/browser/browser.js
Simran-B: Actually, that test also relies on JavaScript execution in the third expect(). The class `selected` for the active tab isn't rendered but added at runtime: https://github.com/github/docs/blob/0c8f74fa4307be81de0d1d1e4ecf88148c3e3553/javascripts/display-platform-specific-content.js#L35

Simran-B: I also can't seem to get jest to execute my rendering test. Does it require the test server to run at port 4001 or something? I keep getting promise resolve timeouts but don't do anything different to existing tests.

Simran-B:
There are a few more things I want to do tomorrow:
- Add a few more tests
- Remove support for `{% all %}`
- Use optional chaining / nullish coalescing operator?

There's also the open question whether the user OS should be considered at all for selecting a tab on load (when no defaultPlatform is specified).

Simran-B:
`const selectedSwitch = await page.$$('a.platform-switcher.selected')` worked so far, but the `selected` class is set by client-side JavaScript, so this might be sensitive to timing. Not sure how to wait properly though.

Simran-B:
- Added two more tests
- Removed support for {% all %}, I hope I didn't miss anything
- ~Use optional chaining / nullish coalescing operator?~ (not used in code base so far)
- Made the only other Actions runner page with a platform switcher default to Linux. Not sure if the frontmatter should be set for other pages as well even if they don't have tabs yet.

Simran-B:
Did something about the frontmatter handling change in the meanwhile? I don't know where this error comes from all of the sudden:

```
 FAIL  tests/unit/page.js (7.321 s)
  ● Page class › platform specific content › page.defaultPlatform frontmatter

    [
      {
        "reason": "invalid frontmatter entry",
        "message": "YML parsing error!"
      }
    ]

      59 | 
      60 |     if (this.frontmatterErrors.length) {
    > 61 |       throw new Error(JSON.stringify(this.frontmatterErrors, null, 2))
         |             ^
      62 |     }
      63 | 
      64 |     // preserve the frontmatter-free markdown content,

      at new Page (lib/page.js:61:13)
      at Object.<anonymous> (tests/unit/page.js:372:20)
```

Jova-art:
Enable

Jova-art:
Enable

Simran-B:
Looks like `new Page()` needs to be `await Page.init()` now. That fixes the unit test.

Added a little bit of documentation for the `defaultPlatform` frontmatter.

The PR should be good to go now.

contributing/content-markup-reference.md
```diff
@@ -93,12 +95,15 @@ These instructions are pertinent to Windows users.
 
 {% endlinux %}
 ```
-```
-{% all %}
 
-Though rare, these instructions are pertinent to any other operating system.
+The platform-specific content matching the reader's operating system shows by default. You can override this behavior for individual Markdown files with the `defaultPlatform` frontmatter (such as for pages about GitHub Actions runners, which mostly use Linux):
```
emilyistoofunky, line 99:
```suggestion
The platform-specific content matching the reader's operating system shows by default. You can override this behavior for individual Markdown files with the `defaultPlatform` frontmatter (such as for pages about GitHub Actions runners, which mostly use Linux).
```

contributing/content-markup-reference.md
```diff
@@ -93,12 +95,15 @@ These instructions are pertinent to Windows users.
...
 
-{% endall %}
+```
+---
+title: …
+versions: …
+defaultPlatform: linux
+---
```
emilyistoofunky, line 106: @Simran-B Let's document this new frontmatter value in our [docs on frontmatter](https://github.com/github/docs/tree/main/content#frontmatter) and then link to it in like 99 rather than sharing the example here. That way, all frontmatter values stay together in one place. ⚡ 

```suggestion
```

emilyistoofunky (commented):
Thanks for opening this @Simran-B! I left a couple of small comments as a code owner from @github/product-docs-content-strategy. Let me know if you have any questions or when you're ready for a final review of the markup reference guide.

Simran-B:
@emilyistoofunky Good call, I moved and extended the frontmatter description. Please take another look.

heiskr (approved):
Looks good to me for @github/docs-engineering . Very thorough.

chiedo:
@emilyistoofunky please let us know when we have the green light to merge!

emilyistoofunky (approved):
@Simran-B Looks great! :shipit:

PR #1982, opened 2020-12-10
The autoupdater does not support crons

chiedo:
Remove autoupdater as it does not support crons.


chiedo:
Yeah... haha...

morty331:
Lookgood

PR #2088, opened 2020-12-14
Add Amazon ECS / Azure App Service deployment guides

swinton:
### Why:

Building on #2085, adds deployment guides for Amazon ECS / Azure App Service as part of a group of select partner deployment guides. 

/cc @runleonarun, @imjohnbo, @jpadams 👋 

### What's being changed:

* Adds a new "deployment guide" for Amazon ECS and Azure App Service based on the existing starter workflows:
    * https://github.com/actions/starter-workflows/blob/master/ci/aws.yml.
    * https://github.com/actions/starter-workflows/blob/master/ci/azure.yml.

### Check off the following:
- [ ] All of the tests are passing.
- [ ] I have reviewed my changes in staging. (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**)
- [ ] For content changes, I have reviewed the [localization checklist](https://github.com/github/docs/blob/main/contributing/localization-checklist.md)
- [ ] For content changes, I have reviewed the [Content style guide for GitHub Docs](https://github.com/github/docs/blob/main/contributing/content-style-guide.md).

PR #2062, opened 2020-12-14
Add default branch note to workflow_run event

lucascosti:
### Why:

Fixes: https://github.com/github/docs/issues/799

### What's being changed:

This adds the default branch note to the `workflow_run` event description. `workflow_run` triggered workflows only occur if the workflow file is on the default branch.

Preview: https://docs-2062--lucascostiworkflow.herokuapp.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#workflow_run

### Check off the following:
- [x] All of the tests are passing. The failing check is a buggy one, and has been reported to engineering.
- [x] I have reviewed my changes in staging. (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**)
- [x] For content changes, I have reviewed the [localization checklist](https://github.com/github/docs/blob/main/contributing/localization-checklist.md)
- [x] For content changes, I have reviewed the [Content style guide for GitHub Docs](https://github.com/github/docs/blob/main/contributing/content-style-guide.md).

PR #2094, opened 2020-12-14
fix some bugs in new workflow

rachmari:
### Why:

It looks like there are two bugs in the new `.github/workflows/triage-unallowed-contributions.yml` workflow

1. When the check to dismiss a review runs but there are no reviews by `github-actions[bot]` the check fails in the `fromJSON` conditional check: https://github.com/github/docs/pull/2062/checks?check_run_id=1548110821
   - To fix this, we need to check that a review result actually exists before getting the properties of the object.
2. When a commit is pushed up and CI runs before the pull request is created, the check to get reviews for the pull request fails:  https://github.com/github/docs/pull/2088/checks?check_run_id=1553035730
    - To fix this, I've added an exit code when the pull request number is undefined and a comment that indicates why the test failed and how to re-run it. Open to other suggestions. There isn't a more direct way to get a pull request from a push event that I'm aware of. 

I added a bit more logging for the future just so we have some basic info.

### Check off the following:

### DIFF
--- a/.github/workflows/triage-unallowed-contributions.yml
+++ b/.github/workflows/triage-unallowed-contributions.yml
@@ -48,7 +48,7 @@ jobs:
               .sort((a, b) => new Date(b.submitted_at) - new Date(a.submitted_at))
               .shift()
 
-            console.log(`Pull request reviews authored by the `github-action` bot: ${botReviews})
+            console.log(`Pull request reviews authored by the github-action bot: ${botReviews}`)
             return botReviews
 
       - name: Get files changed
--- a/.github/workflows/triage-unallowed-contributions.yml
+++ b/.github/workflows/triage-unallowed-contributions.yml
@@ -22,7 +22,7 @@ jobs:
             }).data
               .map(pull => pull.number)
               .shift()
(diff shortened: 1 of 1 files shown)


heiskr (approved):
Makes sense to me

.github/workflows/triage-unallowed-contributions.yml
```diff
@@ -20,8 +22,17 @@ jobs:
...
+              .map(pull => pull.number)
+              .shift()
 
-            return pulls.data.map(pull => pull.number).shift()
+            if (pullNumber) {
+              console.log(`Pull request number: ${pullNumber}`)
+              return pullNumber
+            } else {
+              console.log(`When this workflow ran, the associated pull request was not yet created. Pushing a new commit after you've created a pull request will automatically re-run this workflow, or you can manually re-run the workflow.`)
+              process.exit(1)
```
lucascosti, line 34: @rachmar Will this result in all squashes/merges to `main` resulting in a failed check?

Perhaps the `push` event for this workflow should be excluded from running when on `main`?

Issue #2177, opened 2020-12-16
Instructions for installing the github-pages gem are confusing


monfresh:
### What article on docs.github.com is affected?

[Creating a GitHub Pages site with Jekyll](https://docs.github.com/en/free-pro-team@latest/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll#creating-your-site)

### What part(s) of the article would you like to see updated?

Steps 8 - 10 are confusing. If you search for `could not find gem github-pages` in your favorite search engine, you'll find that the instructions caused people to do one or more of the following things:

- they didn't uncomment the `# gem "github-pages"` line
- they tried to run `bundle update github-pages`
- they replaced `VERSION` with the version of `jekyll` listed on the [Dependency versions](https://pages.github.com/versions/) page instead of the `github-pages` version.

Step 7 is also both confusing and incorrect. Running `jekyll 3.9.0 new .` does not work. The `jekyll new` command does not accept a version number, at least not like it is documented in GitHub docs. Also, the Bundler section is more confusing than helpful because for `bundle exec` to work, a Gemfile must exist in the directory, but the instructions don’t mention that. Instead of adding more steps, I think it's best to simplify step 7 to just `jekyll new .`, which will create the Jekyll site using whatever version of Jekyll is currently installed (most likely 4.2.0). 

In step 8, they are told to comment out the jekyll gem, so it doesn't matter if they used a more recent version of jekyll to create the site initially. What matters is that after they save the Gemfile, they need to run `bundle update` because github-pages only supports jekyll 3.9.0 currently. Running `bundle install` will result in dependency issues.

See the linked PR below that I submitted to fix this.

### Additional information

Examples of people who got stuck because of the documentation: 

- https://talk.jekyllrb.com/t/why-cant-i-install-the-github-pages-plug-in-when-i-set-up-jekyll/3743
- https://talk.jekyllrb.com/t/how-to-install-github-pages-with-jekyll/3510

BenWhetton:
I also encountered this issue just last week. It would have been nice if #2178 had been merged already! It's likely costing users a lot of time to have these kinds of errors in such a key guide.
I have a sligtly different solution to propose than #2178 which uses the github-pages gem to ensure the correct version of jekyll is used to set up the jekyll scaffolding. This works even without a system-wide jekyll installation and will ensure that there are not compatibility issues if jekyll 4+ changes the behavior of `jekyll new` in some way

### The problem as I would describe it:
[...]

PR #2178, opened 2020-12-16
Update creating-a-github-pages-site-with-jekyll.md

monfresh:
### Why:

The previous instructions for installing the `github-pages` gem via the Gemfile were confusing. If you search for `could not find gem github-pages` in your favorite search engine, you'll find that the instructions caused people to do one or more of the following things:

- they didn't uncomment the `# gem "github-pages"` line
- they tried to run `bundle update github-pages
- they replaced `VERSION` with the version of `jekyll` listed on the [Dependency versions](https://pages.github.com/versions/) page instead of the `github-pages` version.

Examples of people who got stuck:
- https://talk.jekyllrb.com/t/how-to-install-github-pages-with-jekyll/3510
- https://talk.jekyllrb.com/t/why-cant-i-install-the-github-pages-plug-in-when-i-set-up-jekyll/3743

This is my attempt to make things clearer by being more specific step by step. By pointing to the Rubygems page for `github-pages`, it removes any ambiguity that the previous dependency page introduced since it lists a bunch of different gems, and it's not absolutely clear which version you're supposed to look for, as evidenced by the person who pinned the `github-pages` gem to a version number for the `jekyll` gem.

Here's the issue I created for this PR: https://github.com/github/docs/issues/2177

### What's being changed:

Instead of an image containing the confusing suggestion to run `bundle update github-pages`, I rephrased the instructions in the image so that they can be followed step by step, and added a step to run `bundle update`, since that is what is needed to install the `github-pages` gem, and the previous instructions did not mention that.

I also updated Step 7 because it was both confusing and incorrect. Running `jekyll 3.9.0 new .` does not work. `3.9.0` is being interpreted as a subcommand, and it is not a valid subcommand. Also, the Bundler section is more confusing than helpful because for `bundle exec` to work, a Gemfile must exist in the directory, but the instructions don’t mention that. Furthermore, if we did create a Gemfile by running `bundle init` for example, it would cause an error when running `bundle exec jekyll new .` because Jekyll expects the directory to be empty. Jekyll creates its own Gemfile so it doesn't make sense to start a new Jekyll site with Bundler.

To avoid all of this confusion, I simplified step 7 to just `jekyll new .`, which will create the Jekyll site using whatever version of Jekyll is currently installed (most likely 4.2.0). 

In step 8, they are told to comment out the jekyll gem, so it doesn't matter if they used a more recent version of jekyll to create the site initially. What matters is that after they save the Gemfile, they need to run `bundle update` because github-pages only supports jekyll 3.9.0 currently. Running `bundle install` results in dependency issues.

### Check off the following:


janiceilene:
@monfresh Thanks so much for opening a PR! I'll get this triaged for review ⚡

monfresh:
Commenting to prevent this PR from being closed on Monday.

content/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll.md
```diff
@@ -63,24 +63,30 @@ Before you can use Jekyll to create a {% data variables.product.prodname_pages %
...
+   ```
+8. Open the file called `Gemfile` that was created, and find the line that looks like this:
+   ```shell
+   # gem "github-pages", group: :jekyll_plugins
+   ```
+   and update it so that it looks like this:
+   ```shell
+   gem "github-pages", "~> 209", group: :jekyll_plugins
+   ```
+   If there is a newer version of the `github-pages` gem, replace `209` above with the latest version that is listed in [Rubygems](https://rubygems.org/gems/github-pages).
```
BenWhetton, line 79: I don't think it is good practice to include explicit version numbers which likely will not match what the user sees.

Instructing the user to `find` a specific line is a bit verbose without adding much information.

I think the target audience for this guide should already know that "the file called Gemfile" is typically just known as a "Gemfile". However, the target audience is not expected to know much about jekyll so I think it may be useful to specify that jekyll creates the Gemfile.

Retaining the previous wording but specifying that the version number should be the version of the github-pages gem and where that information can be found is a better solution IMO. To disambiguate with the previously used `VERSION` identifier and help avoid the confusion identified in #2177, we can use the clearer if more verbose `GITHUB-PAGES-VERSION`.
~~~suggestion
8. Open the Gemfile that was created by jekyll.
9. Add the github-pages gem: update the line starting with `# gem "github-pages"`so that it looks like this, replacing _GITHUB-PAGES-VERSION_ with the latest supported version of the `github-pages` gem. This can be found on the "[Dependency versions](https://pages.github.com/versions/)" page of the {% data variables.product.prodname_pages %} site:
   ```shell
   gem "github-pages", "~> GITHUB-PAGES-VERSION", group: :jekyll_plugins
   ```
~~~

content/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll.md
```diff
@@ -63,24 +63,30 @@ Before you can use Jekyll to create a {% data variables.product.prodname_pages %
...
+   Then find the line that looks like this:
+   ```shell
+   gem "jekyll", "~> 4.2.0"
+   ```
+   and disable it by adding a `#` to it:
+   ```shell
+   # gem "jekyll", "~> 4.2.0"
+   ```
```
BenWhetton, line 87: I don't think it is good practice to include explicit version numbers which likely will not match what the user sees.
~~~suggestion
   Comment out the Gemfile line which starts with:
  ```shell
   gem "jekyll"
   ```
  The correct version Jekyll will be installed as a dependency of the `github-pages` gem.
~~~
I think instructing users on how to comment out lines in a Gemfile is out of scope for this guide so this should be sufficient.

BenWhetton (changes requested):
It is important to target guides at a specific audience to avoid bloating the guide with too many low level details which should already be known by the users. I believe there are a couple of instances of this in this pull request (see inline comments)

This pull request introduces some fixed version numbers which will quickly become out-of-date, no longer matching what the users sees. I understand why this has been proposed but I don't think it is a good solution. It will likely cause different kinds of confusion.
I've proposed some changes inline which I think are clearer than the original text without using explicit versions.

monfresh:
Thanks for the suggestions, @BenWhetton. I've accepted them.

content/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll.md
```diff
@@ -63,24 +63,23 @@ Before you can use Jekyll to create a {% data variables.product.prodname_pages %
...
-9. Update the `gem "github-pages"` line so that the line looks like this, replacing _VERSION_ with the current dependency version for `github-pages`. For more information, see "[Dependency versions](https://pages.github.com/versions/)" on the {% data variables.product.prodname_pages %} site.
-```shell
-gem "github-pages", "~> <em>VERSION</em>", group: :jekyll_plugins
-```
+7. To create a new Jekyll site, use the `jekyll new` command:
+   ```shell
+   $ jekyll new .
+   # Creates a Jekyll site in the current directory
+   ```
+8. Open the Gemfile that was created by jekyll.
```
hubwriter, line 71:
```suggestion
8. Open the Gemfile that Jekyll created.
```

content/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll.md
```diff
@@ -63,24 +63,23 @@ Before you can use Jekyll to create a {% data variables.product.prodname_pages %
...
+9. Add the github-pages gem: update the line starting with `# gem "github-pages"`so that it looks like this, replacing _GITHUB-PAGES-VERSION_ with the latest supported version of the `github-pages` gem. This can be found on the "[Dependency versions](https://pages.github.com/versions/)" page of the {% data variables.product.prodname_pages %} site:
+   ```shell
+   gem "github-pages", "~> GITHUB-PAGES-VERSION", group: :jekyll_plugins
+   ```
+   Comment out the Gemfile line which starts with:
+  ```shell
+   gem "jekyll"
+   ```
```
hubwriter, line 79: As the `gem jekyll ...` line comes first in the file, I think it would be good to swap round the 2 edits you have to do in this file:

```suggestion
1. Add "#" to the beginning of the line that starts with `gem "jekyll"` to comment out this line.
1. Add the `github-pages` gem by editing the line starting with `# gem "github-pages"`. Change this line to:

   ```shell
   gem "github-pages", "~> GITHUB-PAGES-VERSION", group: :jekyll_plugins```

Replace _GITHUB-PAGES-VERSION_ with the latest supported version of the `github-pages` gem. You can find this version here: "[Dependency versions](https://pages.github.com/versions/)."

```

hubwriter (commented):
@monfresh - This is a nice improvement to that page. Thanks for taking the time to do this.

I've made a couple of little suggestions for changes. The main one just changes the order of the 2 edits to the Gemfile.

If you agree with these changes, please just apply them and we can get this merged and live on the site. 

Thanks again. 👍

monfresh:
@hubwriter Thanks for the suggestions. I accepted them.

content/github/working-with-github-pages/creating-a-github-pages-site-with-jekyll.md
```diff
@@ -65,24 +65,23 @@ Before you can use Jekyll to create a {% data variables.product.prodname_pages %
...
+   ```shell
+   $ jekyll new .
+   # Creates a Jekyll site in the current directory
+   ```
+8. Open the Gemfile that Jekyll created.
+1. Add "#" to the beginning of the line that starts with `gem "jekyll"` to comment out this line.
+1. Add the `github-pages` gem by editing the line starting with `# gem "github-pages"`. Change this line to:
+
+   ```shell
+   gem "github-pages", "~> GITHUB-PAGES-VERSION", group: :jekyll_plugins```
```
hubwriter, line 78:
```suggestion
   gem "github-pages", "~> GITHUB-PAGES-VERSION", group: :jekyll_plugins
   ```
```

hubwriter (approved):
Great. This is looking good - let's get it merged. 👍

hubwriter:
@monfresh & @BenWhetton - many thanks for contributing to the GitHub documentation. Much appreciated. 🙇‍♂️

PR #2485, opened 2021-01-01
Added if conditional environment variable example

### LINKED ISSUE #2484: Add if conditional environment variable example to “About environment variables” section

https://github.com/github/docs/blob/main/content/actions/reference/environment-variables.md

https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#about-environment-variables

In the current document it seems that $VAR_NAME can be used to refer to environment variable in any place, including in if conditional

I tried to use $ sign 
if: $DEPLOY_PROD == 'true'  
and got
Unexpected symbol: '$DEPLOY_PROD'. Located at position 1 within expression: $DEPLOY_PROD == 'true'
 It will be good to add 

In a step's if conditional you can refer to environment variable using [`env.<env name> `string](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#env-context) e.g.

```yaml
steps:
  - run: echo "Somebody has a case of the mondays!"
    if: env.DAY_OF_WEEK == 'Mon'
```

MNF:
In the current document it seems that $VAR_NAME can be used in if conditional to refer to environment variable. 
I tried to use $ sign 
if: $DEPLOY_PROD == 'true'  
and got
Unexpected symbol: '$DEPLOY_PROD'. Located at position 1 within expression: $DEPLOY_PROD == 'true'

### Why:
fix https://github.com/github/docs/issues/2484

### What's being changed:

Added if conditional environment variable example

### Check off the following:
- [ x] I have reviewed my changes in staging. (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**)
- [ x] For content changes, I have reviewed the [localization checklist](https://github.com/github/docs/blob/main/contributing/localization-checklist.md)
- [ x] For content changes, I have reviewed the [Content style guide for GitHub Docs](https://github.com/github/docs/blob/main/contributing/content-style-guide.md).


muratkalenderoglu (approved):

muratkalenderoglu (approved):

janiceilene:
Thanks so much for opening a PR! I'll get this triaged for review ✨

janiceilene:
Thanks for your patience! Our small team is working our way through reviewing all of the amazing contributions ✨

lucascosti:
👋 Hi @MNF, thanks for this PR!

This issue is due to how and when environment variables are substituted. Using the `$VAR_NAME` notation means that the environment variable will be substituted on the runner by the runner operating system.

The $VAR_NAME notation can be used in `run` steps  because the variable will be substituted in the runner OS, but in most other places in the workflow file, you should use the `env.VAR_NAME` context. 

Things like the `if` key in a workflow file require the variable to be substituted not in the runner OS, but in the workflow processor. There are other keys in addition to `if:` where that's the context is the only way to refer to a variable (e.g. `runs-on` comes to mind when using a build matrix). 

I agree this should be made clearer in the Actions environment variable docs. I'll try to make a change to your PR that hopefully clarifies this 🙂

content/actions/reference/environment-variables.md
```diff
@@ -32,6 +32,14 @@ steps:
...
 You can also use the {% if currentVersion == "free-pro-team@latest" or currentVersion ver_gt "enterprise-server@2.22" %}`GITHUB_ENV` environment file{% else %} `set-env` workflow command{% endif %} to set an environment variable that the following steps in a workflow can use. The {% if currentVersion == "free-pro-team@latest" or currentVersion ver_gt "enterprise-server@2.22" %}environment file{% else %} `set-env` command{% endif %} can be used directly by an action or as a shell command in a workflow file using the `run` keyword. For more information, see "[Workflow commands for {% data variables.product.prodname_actions %}](/actions/reference/workflow-commands-for-github-actions/#setting-an-environment-variable)."
 
+In a step's if conditional you can refer to environment variable using [`env.<env name> `string](/actions/reference/context-and-expression-syntax-for-github-actions#env-context) e.g.
+
+```yaml
+steps:
+  - run: echo "Somebody has a case of the mondays!"
+    if: env.DAY_OF_WEEK == 'Mon'
+```
+
```
lucascosti, line 42:
```suggestion
To use the value of an environment variable in a workflow file, you should use the [`env` context](/actions/reference/context-and-expression-syntax-for-github-actions#env-context). If you want to use the value of an environment variable inside a runner, you can use the runner operating system's normal method for reading environment variables. Although you can use the runner operating system's syntax in a workflow file's `run` key with a shell command (as shown in the example above), for other places in a workflow file you should use the `env` context.

You can also use the {% if currentVersion == "free-pro-team@latest" or currentVersion ver_gt "enterprise-server@2.22" %}`GITHUB_ENV` environment file{% else %} `set-env` workflow command{% endif %} to set an environment variable that the following steps in a workflow can use. The {% if currentVersion == "free-pro-team@latest" or currentVersion ver_gt "enterprise-server@2.22" %}environment file{% else %} `set-env` command{% endif %} can be used directly by an action or as a shell command in a workflow file using the `run` keyword. For more information, see "[Workflow commands for {% data variables.product.prodname_actions %}](/actions/reference/workflow-commands-for-github-actions/#setting-an-environment-variable)."

```

lucascosti:
👋 @MNF, you can see the result of my change here: https://docs-2485--patch-3.herokuapp.com/en/actions/reference/environment-variables#about-environment-variables

Because I've made some changes myself, I'll get another writer on the team to approve the change 🙂

rnerone35 (approved):
``

rnerone35:
God damn dude you just write that?

content/actions/reference/environment-variables.md
MNF: @lucascosti , your explanation about different ways to refer in a workflow and inside the runner is good( it make it more clear for myself), but examples are more prominent in the document. Without example of env.VAR_NAME people will look try to follow only $VAR_NAME format before reading and understanding the explanation that there are 2 different places where environment variables are used.

I also think that the text in your comment https://github.com/github/docs/pull/2485#issuecomment-762649837 will make explanation more clear and deserve to be in the document:

> There are two ways how and when environment variables are substituted.
> 
> The $VAR_NAME notation can be used in run steps because the variable will be substituted in the runner OS, but in most other places in the workflow file, you should use the env.VAR_NAME context.
> 
> Things like the if and runs-on  keys in a workflow file require the variable to be substituted not in the runner OS, but in the workflow processor.

MNF: May be include both notations in the same example?

```
steps:
  - name: Hello world
    if: env.SHOW_MESSAGE==‘true’
    run: echo Hello world $FIRST_NAME $middle_name $Last_Name!
    env:
      FIRST_NAME: Mona
      middle_name: The
      Last_Name: Octocat
```

lucascosti: Thanks for the feedback and the suggestions, @mnf!

> I also think that the text in your comment #2485 (comment) will make explanation more clear ( it made the behavior more clear for myself) and deserve to be in the document:

It's a bit hard to explain because `$VAR_NAME` depends on the runner OS, and will only work on Unix-based runners; e.g. it won't work on Windows runners with the default shell. I've edited it with a bit more explanation whilst keeping it runner-generic 🙂 

> , your explanation about different ways to refer in a workflow and inside the runner is good, but examples are more prominent in the document.

Sure thing! I've modified the existing example with your 'case of the Mondays' example (🏏 🖨️   😁), which also shows a job-level `env` in addition to the current step stuff. 

Both changes are in https://github.com/github/docs/pull/2485/commits/9dd27799324ae334414af280f1c3712169a1c509

martin389 (commented):
Added a small suggestion

content/actions/reference/environment-variables.md
```diff
@@ -21,15 +21,25 @@ versions:
...
+    runs-on: ubuntu-latest
     env:
-      FIRST_NAME: Mona
-      middle_name: The
-      Last_Name: Octocat
+      DAY_OF_WEEK: Mon
+    steps:
+      - name: "Hello world when it's Monday"
+        if: env.DAY_OF_WEEK == 'Mon'
+        run: echo "Hello world $FIRST_NAME $middle_name $Last_Name, sounds like you have a case of the Mondays!"
```
martin389, line 32: I'd suggest replacing this colloquialism, to help with translation/accessibility:
```suggestion
        run: echo "Hello world $FIRST_NAME $middle_name $Last_Name, today is Monday!"
```

lucascosti: For posterity; @martin389 and I had a robust discussion about the _Office Space_ reference in this example. But, eventually we concluded that especially for people are reading this example not in English (or if English isn’t their first language), this might be confusing if they have never seen the movie. 🙁

The docs must come first, even if this was an awesome reference. 😢

content/actions/reference/environment-variables.md
```diff
@@ -21,15 +21,25 @@ versions:
...
+    runs-on: ubuntu-latest
     env:
-      FIRST_NAME: Mona
-      middle_name: The
-      Last_Name: Octocat
+      DAY_OF_WEEK: Mon
+    steps:
+      - name: "Hello world when it's Monday"
+        if: env.DAY_OF_WEEK == 'Mon'
+        run: echo "Hello world $FIRST_NAME $middle_name $Last_Name, today is Monday!"
```
martin389, line 32:
```suggestion
        run: echo "Hello $FIRST_NAME $middle_name $Last_Name, today is Monday!"
```

martin389 (approved):
Looks good to me, thanks for this great PR!
Added a small non-blocking suggestion.

PR #2866, opened 2021-01-13
Fix navigation under Windows (#2364)

Simran-B:
### Why:

Another version is added to the path if it uses backslashes as separators, leading to href not found warnings, see #2364

### What's being changed:

Use POSIX variant of path.join() which always uses forward slashes instead of the operating system specific separator.

### Check off the following:


chiedo:
Thanks @Simran-B! I've added Windows Tests to this PR so we see the failing tests. A member of our team will chime back in as soon as we get the chance!

heiskr (approved):
Awesome, thanks :)

heiskr:
Hi @Simran-B I think there's a few others, I'll take care of this :)

PR #2899, opened 2021-01-14
Update caching suggestions for Gradle Builds

JLLeitschuh:
Generate Gradle cache keys based upon the version of Gradle and remove lock files before caching

### Why:

Docs recommend caching Gradle files that can cause issues if not removed from the cache.
Also, the cache will be invalidated too frequently.

### What's being changed:

The caching recommendations around Gradle builds.

### Check off the following:


content/actions/guides/building-and-testing-java-with-gradle.md
```diff
@@ -106,10 +106,16 @@ steps:
     uses: actions/cache@v2
     with:
       path: ~/.gradle/caches
-      key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+      # Use the Gradle version in the 'gradle-wrapper.properties' as a cache key. Invalidate caches when Gradle version changes
```
JLLeitschuh, line 109:
```suggestion
      # Use the Gradle version in the 'gradle-wrapper.properties' as a cache key. Invalidate caches when Gradle version changes.
```

lucascosti:
👋 @JLLeitschuh, thanks a lot for this PR, but it looks like I just merged similar changes by @odin-delrio in https://github.com/github/docs/pull/2900.

The main difference in your changes that I can see is deleting some cache files before the end of the job. I'm not a Gradle expert by any means, but for other build/package systems, we generally don't delete things like *-`lock` files before a cache is created.

In the OP, you mention that caching these files might  `cause issues if not removed from the cache.` Could you elaborate on what issues this might be?

JLLeitschuh:
If you have the lock files around, it can confuse the next build that comes along.

This is because it leaves behind information like ports & pids that were running on that last system. When Gradle launches again, it tries to use this information that was left behind, but new/non-gradle services may be running on those ports/pids now.

On a normal system shutdown, those processes, when dying would clean those files up, but if you cache them, they haven't had the opportunity to do that yet.

lucascosti:
Thanks a lot for the explanation, @JLLeitschuh 🙇‍♂️ 

I've resolved the conflicts with `main`, and added a few comments to explain the step for removing those files.

Can you have a 👀 and let me know if you are ok for us to merge? 🙂

lucascosti (approved):

JLLeitschuh (commented):
LGTM! Thanks!

PR #2951, opened 2021-01-17
Fix little punctuation error by adding two periods after two words.

JoseDeFreitas:
### Why:

I consider that after those two words there should be a period in order to make it a little easier to read. When I was reading it I found it weird not having periods right after those words, as the words are highlighted and not a part of the following sentences. Also, the first letter of the following words is capitalized, which makes me think that indeed those periods should be there.

I haven't opened an issue because this change is so small that I don't think it's useful to do it. However, I'll open an issue if needed.

### What's being changed:

This change is as small as you see: I've added two periods just after the two "**Required**" words.

### Check off the following:
- [X] I have reviewed my changes in staging. (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**)


janiceilene:
@JoseDeFreitas Thanks so much for opening a PR! I'll get this triaged for review ✨

skedwards88 (approved):
🚀 Thanks for this addition! This matches the style in the REST API docs as well, where we frequently preface parameters with "Required." 

I'll get this merged down for you.

PR #3010, opened 2021-01-19
Revert exit added to step

rachmari:
This reverts a change added to exit the step when a review was added. This appears to be causing a bug, so we'll revert it for now. See original commit: https://github.com/github/docs/commit/871d886c16c6f50ef8223e7c767dd680f23047b2#diff-8e321f21c41c218596cff700227db7676c7d4f295d63d12e811f141567e92271


janiceilene (approved):
LGTM 🚀

PR #3748, opened 2021-02-17
repo sync

Octomerger:
This is an automated pull request to sync changes between the public and private repos.

:robot: This pull request should be merged (not squashed) to preserve continuity across repos, so please let a bot do the merging!


data/reusables/advanced-security/enable-disable-committer-info.md
```diff
@@ -0,0 +1 @@
+When you enable {% data variables.product.prodname_GH_advanced_security %}, {% data variables.product.prodname_dotcom %} shows how many extra committers this will add and prompts for confirmation. If you disable access to {% data variables.product.prodname_GH_advanced_security %}, you can see how many seats this will free before you confirm the change. This makes it easy to see the impact of your changes on the use of your license.
```
arthurnn, line 1: @felicitymay we have noticed this piece in the docs, that is not quite accurate with what we have done, yet

> you can see how many seats this will free before you confirm the change.

We have plans to do that, but we won't in the next weeks.
The way for the user to know the number of seats disabling the repo will free, is to look at the [Unique Committers count](https://docs.github.com/en/github/setting-up-and-managing-your-enterprise/managing-use-of-advanced-security-for-organizations-in-your-enterprise-account#about-committer-numbers-for-enterprise-accounts). However, note, that is only accessible by the admin.

PR #3921, opened 2021-02-23
Fixed `[...] repository uses GitHub Free or GitHub Free, the [...]`

ErnestoBorio:
Added `for organizations` to fix the text rendered in https://docs.github.com/en/github/working-with-github-pages/creating-a-github-pages-site, which reads:
`If the account that owns the repository uses GitHub Free or GitHub Free, the repository must be public.`
In `div#article-contents`, third child, `<p>`

https://github.com/github/docs/blob/main/content/github/working-with-github-pages/creating-a-github-pages-site.md

I guess `variables.product.prodname_free_user` and `variables.product.prodname_free_team` should have different values, but I'm not sure the impact of editing these would have on the whole docs.

Got the example from `div.product-callout.border-purple.bg-purple-light` from the same page, that adds `for organizations`.


janiceilene (approved):
Thanks so much for opening a PR to make this fix @ErnestoBorio! I'll update the branch and get this merged as soon as tests are passing 🍏 

> I guess variables.product.prodname_free_user and variables.product.prodname_free_team should have different values, but I'm not sure the impact of editing these would have on the whole docs.

You're totally right that changing these variables would impact the overall docs, and in this case, that isn't what we'd want. Your fix is totally the right answer!

PR #3945, opened 2021-02-23
Fixed typo in GraphQL API docs (fixes #3943).

### LINKED ISSUE #3943: Typo in Discussions GraphQL API docs

[Using the GraphQL API for Discussions](https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions#repositorydiscussion), added in 87699fd119211756b05dabfee535830af92d1082.

Right now it says:

> Get a discussion. Returns null if discussion with the specified ID **exists**.

While I believe it should say:

> Get a discussion. Returns null if discussion with the specified ID **does not exist**.

radekmie:
A fix for the typo reported in https://github.com/github/docs/issues/3943.


shubhisinha2010 (approved):
Great work and observation!

janiceilene (approved):
Thanks so much for following up with a PR for this fix @radekmie! We're frozen for the weekend, but I'll get this merged in soon 💖

PR #4156, opened 2021-03-02
"Issued at" time should allow for clock drift

### LINKED ISSUE #4155: Apps authentication example sets `iat` to `now`

content/developers/apps/authenticating-with-github-apps.md uses an example the sets `iat` to `now`.  This is fine if the client and server clocks are both synchronized to UTC perfectly.  In the somewhat common case that the client system is out of sync with UTC and is ahead of UTC, setting `iat` to `now` will result in an authentication error.  Users not aware of this potential issue will have difficulty tracking down the error. 

The example should set `iat` to some arbitrary time slightly in the past, 60 seconds ago for instance.   This will allow for some harmless clock drift from UTC without causing hard to diagnose errors.

bitwiseman:
If `iat` is in the future from the server clock perspective GitHub will report an error. 
This can happen if the clock on the client system is significant out of sync with UTC and/or server time.

Adjusting `iat` to an arbitrary 60 seconds in the past means the error will not occur in cases where the client system clock is only off by a few seconds, which is not uncommon.

### Why:

Closes #4155. 

### What's being changed:

The example is being modified slightly.

### Check off the following:


janiceilene:
@bitwiseman Thanks so much for opening a PR! I'll get this triaged for review ⚡

skedwards88 (approved):
Thank you for improving this example! I will get these changes merged down for you.

PR #4498, opened 2021-03-13
Fix incorrect operator example

### LINKED ISSUE #4488: Example of an "if" expression is not correct

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

The example of how to use an expression with an `if:` conditional appears to be inconsistent.
The "Contexts and expression syntax for GitHub Actions" article begins with the following:

> When you use expressions in an `if` conditional, you may omit the expression syntax (`${{ }}`) because GitHub automatically evaluates the if conditional as an expression, unless the expression contains any operators. If the expression contains any operators, the expression must be contained within `${{ }}` to explicitly mark it for evaluation. For more information about if conditionals, see "Workflow syntax for GitHub Actions."

The article then provides the following example usage:

```yaml
name: CI
on: push
jobs:
  prod-check:
    if: github.ref == 'refs/heads/main'
    runs-on: ubuntu-latest
    steps:
      - run: echo "Deploying to production server on branch $GITHUB_REF"
```

The `==` symbol is an operator, and therefore requires the expression syntax.
Using this example without the expression syntax will result in the GitHub Action to fail.

### Additional information

See https://github.com/shipwright-io/build/pull/661

lucalves:
### Why:

I included a syntactic expression in the example of the article [Context and expression syntax for GitHub Actions](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions), because the expression in the example has `==` as an operator, and according to the documentation it needs a `$ {{ }}` to explicitly mark it for evaluation.

**Closes #4488**.

### Check off the following:


janiceilene:
Thanks for opening a PR @lucalves and thanks to @adambkaplan for opening an issue! I'll get this triaged for the team to review ⚡

skedwards88 (approved):
🎉 Thank you for fixing this! I'll get it merged down for you.

jaqx0r:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif also contains this error btw.

cmwilson21:
@jaqx0r Thanks for catching that and reporting it. Would you mind opening a new issue with this information? Then we can triage it and get it up for review ⚡

PR #4549, opened 2021-03-15
docs(actions): make helpful note about env variable visually apparent

roadlittledawn:
### Why:
Scanning this page took us about 30min to find an important tidbit of info re: `GITHUB_TOKEN` being the only available secret available for a runner on a forked repo.

### What's being changed:
Changed a reusable snippet that contained an important tidbit about the GITHUB_TOKEN env var into a note callout. Wasn't sure if it should always be a called out, but we would certainly find that helpful. Otherwise, maybe the snippet should just be wrapped in a note?

### Check off the following:


janiceilene:
@roadlittledawn Thanks so much for opening a PR! I'll get this triaged for review ⚡

skedwards88:
@roadlittledawn Thanks for your contribution! This reusable is used in a few places (3 times in [Events that trigger workflows](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#webhook-events) and once in [Encrypted secrets](https://docs-4549--action-github-token.herokuapp.com/en/actions/reference/encrypted-secrets)). Can you share which section felt like it needed a callout? I ask because in the "Events that trigger workflows" article, this will now result in 3 notes close together, so we may want to consider a different solution, like just wrapping the reusable in a note in the "Encrypted secrets" article if that was where you felt needed the additional highlight:

image

roadlittledawn:
@skedwards88 yeah i kind of wondered if the reusable content snippet should only be given the callout treatment in context of the article and not everywhere. the section we were reading that prompted to file PR was this one:

https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

skedwards88:
Thanks for adding this context! I think adding the `note` markup just around this instance of the resuable would make more sense. Feel free to make this change or to add the `help wanted` label to this PR. Thanks again for your contribution!

hubwriter:
I've applied the changed agreed by @roadlittledawn above - i.e. put the text in a note box just in this article.

hubwriter (approved):

roadlittledawn:
thanks much!

PR #4863, opened 2021-03-27
Fix incorrect operator from the environment variables article

lucalves:
### Why:

I included a syntactic expression in the example of the article [Environment variables](https://docs.github.com/en/actions/reference/environment-variables), in the section [About environment variables](https://docs.github.com/en/actions/reference/environment-variables#about-environment-variables) because the expression in the example has `==` as an operator, and according to the documentation, it needs a `$ {{ }}` to explicitly mark it for evaluation.

There is no open Issue related to this.

### Check off the following:


janiceilene:
@lucalves Thanks so much for opening a PR! I'll get this triaged for review ⚡

shati-patel (approved):
Thank you for the contribution, @lucalves! I'll get this merged shortly 😃 

The existing example seems to work, but you're absolutely right that the [documentation](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions) says 
> If the expression contains any operators, the expression must be contained within ${{ }} to explicitly mark it for evaluation.

Let's update this example to match!

shati-patel (changes requested):
It turns out I was too hasty in approving this 😅 

Because of the `{{` symbols in the example, we need  `raw` tags around the example. Like this: 

````

{% raw %}
```yaml
jobs:
  weekday_job:
    runs-on: ubuntu-latest
    env:
      DAY_OF_WEEK: Mon
    steps:
      - name: "Hello world when it's Monday"
        if: ${{ env.DAY_OF_WEEK == 'Mon' }}
        run: echo "Hello $FIRST_NAME $middle_name $Last_Name, today is Monday!"
        env:
          FIRST_NAME: Mona
          middle_name: The
          Last_Name: Octocat
```
{% endraw %}

````

Could you update this? That should fix the failing PR check!

shati-patel (approved):
Looks great, and the PR checks are now passing! 🎉 Thanks again, @lucalves 😄

PR #4957, opened 2021-03-31
Update troubleshooting-custom-domains-and-github-pages.md

JOJ0:
### Why:

The current docs were misleading to me and I had to ask github support why it doesn't work as expected. Support agent Tammy Metz was very helpful and clarified that actually only the special subdomain www is redirected to the apex domain.

### What's being changed:

Clarify that using both apex _and_ custom subdomain is not generally unsupported and redirects between `www` and apex domain are automatically set up by github.

I wrote "If configured correctly" - there should be a link a place where this is explained. Running out of time this moment but could add this later if you point me in the right direction. Thanks in advance!


tmetz (dismissed):
Thanks for taking the time to submit these PRs, @JOJ0!  I definitely think this clarifies exactly what the confusion was.  Looks good to me, so now we'll see if the Docs team has any suggestions!

janiceilene:
@JOJ0 Thank you so much for opening this PRs 😍 I've triaged this for the team to take a look ✨

lucascosti (approved):
I made a small change, and this is ready to go!

Thanks so much for all these contributions, @JOJ0! 🙇‍♂️

JOJ0:
thumbs up and many thanks! :-)

PR #6664, opened 2021-05-22
git init -b main on a mac

eliork:
When I'm trying to type 

``` 
git init -b main
 ```
on my terminal on macOS I'm getting an error:
```
error: unknown switch `b'
usage: git init [-q | --quiet] [--bare] [--template=<template-directory>] [--shared[=<permissions>]] [<directory>]
```

I think this should be updated.

### Why:

Closes [issue link]

### Check off the following:

- [ ] I have reviewed my changes in staging (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**).

### DIFF
--- a/content/github/importing-your-projects-to-github/importing-source-code-to-github/adding-an-existing-project-to-github-using-the-command-line.md
+++ b/content/github/importing-your-projects-to-github/importing-source-code-to-github/adding-an-existing-project-to-github-using-the-command-line.md
@@ -28,7 +28,7 @@ versions:
 3. Change the current working directory to your local project.
 4. Initialize the local directory as a Git repository.
   ```shell
-  $ git init -b main
+  $ git init main
   ```
 5. Add the files in your new local repository. This stages them for the first commit.
   ```shell

### CODE REVIEW

**Comment by felicitymay** [comment]
Hi - it sounds as if you're using a version of Git that doesn't include this command. It was introduced in [Git 2.28.0](https://git-scm.com/docs/git-init/2.28.0) in July 2020. Can you check which version of Git you're using?

```
git --version
```
It sounds as if possibly this article should mention that you'll need Git 2.28.0 or more recent for this command. For reference, this stack overflow article also mentions a workaround for older versions of Git: https://stackoverflow.com/questions/42871542/how-can-i-create-a-git-repository-with-the-default-branch-name-other-than-maste.

felicitymay:
Since this uses functionality that was introduced over 10 months ago, I'm going to close this issue. 

If you feel strongly that we should update the docs to mention the need for Git 2.28.0+, feel free to open an issue and make this suggestion. Then we can discuss which articles to update with this information.

PR #6615, opened 2021-05-20
Tweak information regarding the GitHub GPG key

jernej-9:
### Why:

Following the discussion #6444 I had with @felicitymay, this pull request includes changes to information regarding GPG verification that have been discussed in that thread.

### What's being changed:

The paragraph with information on GitHub's key that is used to sign web commits has been slightly rearranged to improve readability, and the full fingerprint of the key has been added, which enables everybody to directly verify that the key they downloaded is indeed correct.

### Check off the following:

- [X] I have reviewed my changes in staging (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**).
- [X] For content changes, I have completed the [self-review checklist](https://github.com/github/docs/blob/main/CONTRIBUTING.md#self-review).

---

I think that for security reasons, before merging in this pull request, it is necessary that at least one, but preferably more GitHub employees check the key fingerprint and explicitly write an approval in this pull request stating that the fingerprint is indeed correct.

PR #4945, opened 2021-03-30
Actions: Add deprecationMessage metadata syntax field

Simran-B:
### Why:

Document this useful feature for everyone. Discovered on [github.community](https://github.community/t/feature-request-mark-inputs-as-deprecated/171417)

### What's being changed:

- Add description for `deprecationMessage`
- Mark two variables as inline code

### Check off the following:
- [x] I have reviewed my changes in staging. (look for the **deploy-to-heroku** link in your pull request, then click **View deployment**)
- [x] For content changes, I have reviewed the [localization checklist](https://github.com/github/docs/blob/main/contributing/localization-checklist.md)
- [x] For content changes, I have reviewed the [Content style guide for GitHub Docs](https://github.com/github/docs/blob/main/contributing/content-style-guide.md).

PR #6740, opened 2021-05-24
Describe `$RUNNER_OS`/`$RUNNER_TEMP`/`$RUNNER_TOOL_CACHE` environment variables

### LINKED ISSUE #6127: `$RUNNER_OS`/etc environment variables not documented (GitHub Actions)

https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

This [StackOverflow answer](https://stackoverflow.com/a/57948488) claims that GitHub Actions workflow steps are provided with an environment variable `$RUNNER_OS` that contains the value of `${{ runner.os }}`.

This is confirmed by a simple workflow that just prints out `env | sort`:

```
…
RUNNER_OS=Linux
RUNNER_PERFLOG=/home/runner/perflog
RUNNER_TEMP=/home/runner/work/_temp
RUNNER_TOOL_CACHE=/opt/hostedtoolcache
RUNNER_TRACKING_ID=github_069548eb-c4f3-4542-baac-43ed0d140801
RUNNER_USER=runner
RUNNER_WORKSPACE=/home/runner/work/pysam
…
```

It also appears to be confirmed by the runner source code; see https://github.com/actions/runner/blob/419ed24c1ee8c70eeb096570980e700e833f81f6/src/Runner.Worker/RunnerContext.cs#L13

The affected web page lists environment variables available to workflow steps, including `$CI` and many variables of the form `$GITHUB_xyz`. If (some of) these `$RUNNER_xyz` environment variables are also supported for use by workflow steps, it would be useful if they were added to this listing.

### Additional information

This is somewhat related to the previously-reported #2031, but that issue was interpreted as being about the context expression `${{ runner.os }}` rather than the environment variable `$RUNNER_OS`.

jmarshall:
### Why:

Closes #6127.

### What's being changed:

As noted in #6127, several `$RUNNER_xyz` environment variables are set up by runners corresponding to `${{ runner.xyz }}` context properties. A total of **seven** `RUNNER_*` environment variables are currently set in job steps, and it is unclear which are intended to be used and are supported for use by workflow steps. There are **three** corresponding `runner` context properties that are documented in “[Context and expression syntax for GitHub Actions](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context)”, so this PR adds the environment variables corresponding to these three to the “[Default environment variables](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)” list:

| Environment variable | Description
|---|---
| `RUNNER_OS` | The operating system of the runner executing the job. Possible values are `Linux`, `Windows`, or `macOS`.
| `RUNNER_TEMP` | The path of the temporary directory for the runner. This directory is guaranteed to be empty at the start of each job, even on self-hosted runners.
| `RUNNER_TOOL_CACHE` | The path of the directory containing some of the preinstalled tools for [GitHub]-hosted runners. For more information, see "[Specifications for [GitHub]-hosted runners](/actions/reference/specifications-for-github-hosted-runners/#supported-software)".

(The `tool_cache` property value is not documented as being used for GitHub AE, so this PR just leaves `$RUNNER_TOOL_CACHE` unlisted and undocumented in this case.)

In considering merging this, it would be prudent for GitHub staff to confirm exactly which of these environment variables are intended to be supported for use in workflow steps; in particular, to confirm whether exactly these three are supported as environment variables, similarly to their context property equivalents.

(On #6127 it was suggested that I might attempt to confirm which are supported by contacting GitHub Support. However there is little point in a non-GitHub-employee such as myself doing that, as whoever took responsibility for merging this would need to confirm directly with other GitHub employees (in engineering or support) that the information so obtained was accurate.)

### Check off the following:


Narin506 (dismissed):

janiceilene:
👋 @jmarshall We cannot accept this PR until you've verified with support which environment variables should be documented. After you get that information from the [support team](https://support.github.com/contact) please feel free to open a new issue with that information.

Thanks for your interest in improving the GitHub docs!

jmarshall:
This repository's [CONTRIBUTING.md](https://github.com/github/docs/blob/main/CONTRIBUTING.md#reviewing) says

> We (usually the docs team, but sometimes GitHub product managers, engineers, or supportocats too!) review every single PR.

This implies that sometimes the GitHub Docs team brings PRs to the attention of their GitHub colleagues in the relevant part of Engineering who will review them directly. That would be a much more efficient way to verify which of these environment variables are supported than for me to compose the question a third time in such detail that Support sees its importance, routes the question to the right people, and replies in a reasonable timeframe. But, as noted above, even if that happened it would be pointless because you would have no reason to believe what I, a non-GitHub-employee, reported — so the Docs team would have to verify it directly with Engineering anyway.

I notice that in e.g. https://github.com/github/docs/pull/3060#issuecomment-780266122, https://github.com/github/docs/pull/3788#pullrequestreview-603501261, https://github.com/github/docs/pull/4945#issuecomment-814611367, https://github.com/github/docs/pull/5359#issuecomment-834852572, and https://github.com/github/docs/pull/6615#issuecomment-847767421 the Docs team have themselves consulted their colleagues elsewhere within GitHub to directly verify aspects of these other outsiders' PRs.

Is there a reason why this PR is being treated differently? Thanks.

janderssonse:
Following. The reason given for closing this does not make sense at all IMHO. It basically says that  PR's for GH Docs should go to https://support.github.com/contact first?

janiceilene:
:wave: @jmarshall I’ve spoken with the team and we've decided to reopen this PR. I apologize for closing this PR prematurely. Thank you for taking the time to voice your concerns with our process so that we can continue to improve.

Please note that we’ll still need to validate with the engineering team and don’t know when they’ll have the bandwidth to prioritize. We appreciate your patience.

content/actions/reference/environment-variables.md
```diff
@@ -72,6 +72,9 @@ We strongly recommend that actions use environment variables to access the files
 | `GITHUB_SERVER_URL`| Returns the URL of the {% data variables.product.product_name %} server. For example: `https://{% data variables.product.product_url %}`.
 | `GITHUB_API_URL` | Returns the API URL. For example: `{% data variables.product.api_url_code %}`.
 | `GITHUB_GRAPHQL_URL` | Returns the GraphQL API URL. For example: `{% data variables.product.graphql_url_code %}`.
+| `RUNNER_OS` | The operating system of the runner executing the job. Possible values are `Linux`, `Windows`, or `macOS`.
+| `RUNNER_TEMP` | The path to a temporary directory on the runner. This directory is emptied after each job completes.
```
martin389, line 76: Is this behaviour guaranteed, or might there still be files remaining if a self-hosted runner has previously terminated unexpectedly?

martin389:
(Closing and reopening this PR for CI purposes)

martin389:
Checking questions with support.

content/actions/reference/environment-variables.md
martin389: And just to confirm: is the directory cleared before the job starts, or after the job ends (whether success or failure)?

jmarshall: It seemed obvious so I wasn't explicit about this in the commit message, but this proposed text was copied from the corresponding entries in _context-and-expression-syntax-for-github-actions.md_. So you might apply any wording improvements there too.

jmarshall: The proposed rewording changes a definitive statement (“guaranteed to be empty at [job start]”) to a vague one (“emptied after each job completes” — What exactly does ”complete” mean, in the context of failures? Could anything else be putting files there in between jobs?). In fact for users of GH Actions it doesn't matter exactly when the directory is cleared, because the difference is not observable. The only thing that is observable is whether the the directory is empty at the start of each job, so IMHO the existing text is the most useful for readers. (I suppose some readers may also care that temporary files are not left lying around afterwards either, but sure-to-be-empty-on-start is the critical part for most.)

Looking at [_JobRunner.cs_](https://github.com/actions/runner/blob/main/src/Runner.Worker/JobRunner.cs) (search for `_tempDirectoryManager`), you can see that the current implementation clears the directory both before **and** after each job.

martin389:
Sorry about the delay here, following up 👍

content/actions/reference/environment-variables.md
TingluoHuang: `RUNNER_TEMP` will be cleaned up by the runner in the best effort way. 
the runner will try to clean it up before the job and also after the job, so even you unplug the power from the runner to cause the post-job clean not to run, it will still get cleaned before the next job starts running.
However, the runner can't delete files if the file is created by a different user and the runner user account don't have permission to delete the file.

martin389: Thanks @TingluoHuang - I've updated the description to mention this possibility: https://github.com/github/docs/pull/6740/commits/031ae8476c5aca6d57c6463b75c9eeb6173084a1

martin389:
Received confirmation that these environment variables are supported: RUNNER_OS, RUNNER_TEMP, and RUNNER_TOOL_CACHE 👍

martin389 (dismissed):

martin389 (approved):

martin389:
Changes are now live:

| Published article  | Changes |
| --- | --- | 
“[Default environment variables](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables)” |  Adds entries to table for `RUNNER_OS`, `RUNNER_TEMP`, and `RUNNER_TOOL_CACHE `
"[Context and expression syntax for GitHub Actions](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#runner-context)"| Moved descriptions into reusables for consistency.

Issue #6642, opened 2021-05-21
SSH (backticks and absent command)


fluesvamp:
### What article on docs.github.com is affected?

[Generating a new SSH key and adding it to the ssh-agent
](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)

generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md

### What part(s) of the article would you like to see updated?

_Reduce the amount of provided OS specific information in the file._

Mac and Linux
```
$ eval "$(ssh-agent -s)"
> Agent pid 59566
```
Windows (uses backticks)  

_The command could be the same as on the above OSes._
```
# start the ssh-agent in the background
$ eval `ssh-agent -s`
> Agent pid 59566
```
[Adding your SSH key to the ssh-agent](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent#adding-your-ssh-key-to-the-ssh-agent)

_For Linux and Windows another line could be added accessing a new kill_ssh_agent.md file._

`4. {% data reusables.command_line.kill_ssh_agent %}`

kill_ssh_agent.md (proposed)

``````
Kill the current agent (given by the SSH_AGENT_PID environment variable).
```shell
$ eval "$(ssh-agent -k)"
> Agent pid 59566 killed
```
``````

Here's the sole reference to killing the agent PID.

```
$ grep -nr ssh-agent content | grep kill
content/github/authenticating-to-github/connecting-to-github-with-ssh/working-with-ssh-key-passphrases.md:84:The `ssh-agent` process will continue to run until you log out, shut down your computer, or kill the process.
$ 
```

_Perhaps the OS specific content regarding keychain could be hosted by the vendor and linked to._

<hr>

["Working with SSH key passphrases"](https://docs.github.com/en/articles/working-with-ssh-key-passphrases)

working-with-ssh-key-passphrases.md

[Mac version of the page]
Saving your passphrase in the keychain

_This content is perhaps dated._

[Windows version of the page]
Auto-launching `ssh-agent` on Git for Windows

_Perhaps this content could be shifted to a Wiki page._

### Additional information

If GitHub Inc. wants to hire me to review the entire repository send me an e-mail.

janiceilene:
[...]

PR #6782, opened 2021-05-26
Added update to "Adding your SSH key to the ssh-agent"

RDxR10:
An alternative to establishing a connection to the ssh-auth agent.

### Why:

Added an alternate command here(this PR would solve part of this issue) [https://github.com/github/docs/issues/6642]

### What's being changed:

```
exec ssh-agent bash
```

### Check off the following:

### DIFF
--- a/content/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
+++ b/content/github/authenticating-to-github/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
@@ -170,6 +170,12 @@ Before adding a new SSH key to the ssh-agent to manage your keys, you should hav
 
 1. {% data reusables.command_line.start_ssh_agent %}
 
+  as an alternative, this command can be used as well on the current working directory:
+
+  ```
+  $ exec ssh-agent bash
+  ```
+
   In some Linux environments, you need root access to run the command:
   
   ```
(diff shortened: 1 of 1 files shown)


fluesvamp:
The ssh-agent man page adequately addresses ways to get an agent set up.

I am opposed to the merging of this pull request.

admin-githup (commented):
Goog

janiceilene:
@RDxR10 Thanks so much for opening a PR! I'll get this triaged for review ✨

RDxR10:
Let me know if it needs any further clarification.

lecoursen:
Thanks for taking the time, @RDxR10, but our docs are purposefully opinionated. We document one way to accomplish a task and only mention alternatives if there's a compelling reason to do so. I'm going to close this PR, but let me know if there's further explanation for why this alternative is important to document.

RDxR10:
@lecoursen the reason is that depending on the os or Linux environment, the command may vary.
More detailed info here: https://stackoverflow.com/questions/17846529/could-not-open-a-connection-to-your-authentication-agent.
Variant: https://stackoverflow.com/a/20403535/456814

lecoursen:
> the reason is that depending on the os or Linux environment, the command may vary.

Thanks for clarifying! From those links, it seems that there are more than two possibilities. We can't list them all, so we should instead clarify in `{% data reusables.command_line.start_ssh_agent %}` that the command we're listing is only an example and that you may need a different command depending your environment. Feel free to reopen this PR if you want to update it accordingly!

PR #6941, opened 2021-05-31
Remove 'public' clarification for repos eligible for community files

deniseyu:
### Why:

Closes https://github.com/github/communities/issues/429

### What's being changed:

It looks like there was some historical ambiguity about whether private repos should inherit the `org/.github` repo's community health files (Code of Conduct, Issue templates, etc.) Thanks to @smashwilson's investigation, we now know that the actual behavior is that private repos **do** inherit. This PR updates the docs to reflect that.

We explored making the feature consistent with the docs because it was generating confusion, but we don't have a strong product reason at the moment for enforcing that the community files are only inherited by public repositories. So we're conforming the docs to the behavior!

### Check off the following:


deniseyu:
This change will need to be propagated to the Enterprise docs as well (https://docs.github.com/en/enterprise-server@3.1/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file) - I'm not 100% sure about the process for that, but happy to send another round of PRs!

mcsats (approved):

hendis85 (approved):

hubwriter (approved):
@deniseyu - Change looks good, thanks.
Your changes will appear in both the dotcom and enterprise versions of this page.

PR #7188, opened 2021-06-08
Enable jsx-a11y linter and linting

khiga8:
This PR introduces the accessibility plugin for `eslint` which will enable accessibility-related code linting in React code ([eslint-plugin-jsx-a11y](https://www.npmjs.com/package/eslint-plugin-jsx-a11y)) . Currently, we rely solely on pa11y for catching accessibility violations in high-traffic docs pages. We should make use of this eslint plugin to help detect accessibility issues earlier in the development process. 

However, there were a few steps I had to take to take before I could enable this plugin. `eslint` is currently disabled for `.tsx` files in order to avoid conflict with `prettier`. Therefore, I took the following steps:

1. Allow `eslint` to run on tsx files
2. Introduce `eslint-config-prettier-plugin` to handle conflict between `prettier` and `eslint`
3. Introduce `@typescript-eslint/parser` and `@typescript-eslint` to handle typescript code.
4. Resolve conflicts that arise from enabling these linters. I addressed the unused/duplicate import violations but don't have capacity to address a few rules like camelcase so they are disabled for now.
5. Introduce a11y plugin
6. Inline disable the 14 or so violations so it is known and can be addressed.

**Note**
- Automation can only catch a small percentage of accessibility issues and is by no means a solution to accessibility. It's still valuable to leverage these tools to help catch issues earlier and save time in manual audits.


khiga8:
[Slack convo](https://github.slack.com/archives/C0128FB8J3H/p1623173287451300)

.eslintrc.js
```diff
@@ -24,6 +25,24 @@ module.exports = {
...
+      ],
+      extends: ['plugin:jsx-a11y/recommended'],
+      parser: '@typescript-eslint/parser',
+      rules: {
+        'camelcase': 'off',
+        'no-unused-vars': 'off',
+        'no-undef': 'off',
+        'no-use-before-define': 'off',
+        '@typescript-eslint/no-unused-vars': ['error'],
+      }
```
khiga8, line 45: I've disabled the `camelcase`, `no-undef` rule in typescript files until there is capacity for them to be addressed.

`no-unused-vars` on typescript files will throw false positives so we will disable that, and instead use `'@typescript-eslint/no-unused-vars'`

mikesurowiec (dismissed):
Looks great, thank you for doing this!

mikesurowiec (approved):

PR #7713, opened 2021-06-26
Added build step to prepare script

### LINKED ISSUE #222: first local test run requires `npm run build`
A fresh clone of this repo requires `npm run build` to be run before all the tests will pass. This is an easy thing to forget because you only have to run it once and then `npm test` continues to work without a fresh build.

One way we can work around this is by adding a `pretest` script to package.json with a value of `npm run build`. That way the build will always run first when anyone runs `npm test`.

That would solve the immediate problem for first-time contributors, but it would also incur a performance penalty every time you run the tests locally, as the build currently takes ~4 seconds to run on my superfast computer. Maybe 4s doesn't matter much in the grand scheme of things... but it would be nice to not incur that cost.

🤔 What are some ways we can solve this?

cc @github/docs-engineering @jeffmcaffer

FarazzShaikh:
Added build step to prepare script and Removed redundant calls of build script

### Why:

Closes #222 

### Check off the following:


FarazzShaikh:
I have added `npm run build` to the `prepare` script in `package.json`. As a result, I have attempted to remove all redundant calls of `npm run build`.

However, there are some calls that I am not sure how to handle. Here are they:

- https://github.com/FarazzShaikh/docs/blob/b08a65a90373d6cd92604b7ca63d6fd78b5a1265/.github/workflows/sync-single-english-algolia-index.yml#L51
- https://github.com/FarazzShaikh/docs/blob/b08a65a90373d6cd92604b7ca63d6fd78b5a1265/.github/workflows/test-windows.yml#L64
- https://github.com/FarazzShaikh/docs/blob/b08a65a90373d6cd92604b7ca63d6fd78b5a1265/.github/workflows/test.yml#L69
- https://github.com/FarazzShaikh/docs/blob/b08a65a90373d6cd92604b7ca63d6fd78b5a1265/script/enterprise-server-deprecations/archive-version.js#L110
- https://github.com/FarazzShaikh/docs/blob/b08a65a90373d6cd92604b7ca63d6fd78b5a1265/Dockerfile#L58

And maybe other that I might have missed. These calls perform some operations after `npm ci` and before `npm run build`.

Chelle76 (approved):

janiceilene:
@FarazzShaikh Thanks so much for opening a PR! I'll get this triaged for review ⚡

janiceilene:
Thanks for opening this PR @FarazzShaikh! Sometimes we need to run packages without making a build, having to build to install would be a deal breaker for some processes, so we can't accept this PR. 

Thanks so much for your interest and passion in improving the GitHub docs 💛

FarazzShaikh:
@janiceilene thank you for your review. You should close #222 as well if this cannot be implemented.

JamesMGreene:
@FarazzShaikh We really appreciate the PR you offered up. 🙇🏻‍♂️ 

Unfortunately, when the larger team reviewed your PR, one of the valid issues mentioned was that this change (adding a `prepare` script to run the `npm run build` script) would've forced us to _always_ run the build, even when we don't want or need to, and thus slowing down certain processes (e.g. in many of our Actions workflows). That was a perspective that I hadn't considered when I encouraged you to create this PR. 😓 

Since npm does not currently provide us a mechanism to opt-out of that behavior (other than `--ignore-scripts`, which also ignores all of the _dependencies'_ lifecycle scripts, too 😬 🙅🏻‍♂️), we decided it was better for us to stick with the opt-in behavior that we currently have. While we understand that does put a slight burden on contributors to execute that one extra step, we believe it is something that is easily enough learned and worked around. 🤷🏻‍♂️ 

Thank you especially for your thoroughness in taking the extra step of updating the contributor documentation, the Codespaces configuration, and the Actions workflows as well. ✨ 

We are thankful for your involvement and effort. We would love to see additional contributions from you in the future. ❤️

FarazzShaikh:
Understood, Thanks for letting me know!  😄

PR #7730, opened 2021-06-26
[WARNING] I am not supposed to do this. I think you have a security issue.

julianxhokaxhiu:
This PR is a simple test about creating a PR inside a repository I don't own permissions to operate, although I can create PRs into.

Apparently you can do this in any repo and I've reported this already to Support but they didn't found the source of the issue. I think I found it. Just go on any repository you want and you don't own, click on `Pull Requests` and then `New Pull Request`, pick two branches you don't own and create a PR, just like this one.

I think that opening PRs inside a repository by any other user should be allowed only from their personal forks into the main forked repository. If that's not the case, I lost the point where Github started to allow this.

Would you be able to consider ironing this part and fixing it instead please?

Thank you in advance,
Julian


Nguyenvanduong98ht:
https://github.com/github/docs/commit/854238357105a6da2bdc1f32e14a343ff0352a98

Nguyenvanduong98ht:
> [8542383](https://github.com/github/docs/commit/854238357105a6da2bdc1f32e14a343ff0352a98)

PR #7824, opened 2021-06-30
Update 'Using a specific shell'

### LINKED ISSUE #6554: Document that the GitHub Actions shell examples assume what you set in run are saved to a temporary file

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell

The table of shells gives examples of how they are used, but they are all from the perspective of passing in a file. I think clarifying that the text in the `run` key is saved to a temp file and the example command is used on that temp file would help make the examples make more sense.

### Additional information

For instance, the Python example is:
```YAML
steps:
  - name: Display the path
    run: |
      import os
      print(os.environ['PATH'])
    shell: python
```

And the shell docs say that Python uses `python {0}`. That would only work if what is specified in `run` was saved to a temp file.

room-27:
I attempted to clarify that a specified shell will execute commands in each 'run' keyword from the context of a temporary file where they are saved. My (first) edit probably needs refining from suggestions, but I hope it only clears confusion for now, not causes any.

### Why:

Closes https://github.com/github/docs/issues/6554

### What's being changed:

I added the following statement (italics) under [Using a specific shell](https://github.com/room-27/docs/blob/3e182e98e47e2ac5fba4db699a43f9dbcc37a165/content/actions/reference/workflow-syntax-for-github-actions.md#using-a-specific-shell):
> You can override the default shell settings in the runner's operating system using the shell keyword. You can use built-in shell keywords, or you can define a custom set of shell options. _The shell command run internally will execute a temporary file, containing the commands you specify in each run keyword._

### Check off the following:


janiceilene:
@room-27 Thanks so much for opening a PR! I'll get this triaged for review ✨

content/actions/reference/workflow-syntax-for-github-actions.md
```diff
@@ -799,7 +799,7 @@ Using the `working-directory` keyword, you can specify the working directory of
 
 ### Using a specific shell
 
-You can override the default shell settings in the runner's operating system using the `shell` keyword. You can use built-in `shell` keywords, or you can define a custom set of shell options.
+You can override the default shell settings in the runner's operating system using the `shell` keyword. You can use built-in `shell` keywords, or you can define a custom set of shell options. The shell command run internally will execute a temporary file, containing the commands you specify in each `run` keyword.
```
lucascosti, line 802:
```suggestion
You can override the default shell settings in the runner's operating system using the `shell` keyword. You can use built-in `shell` keywords, or you can define a custom set of shell options. The shell command that is run internally executes a temporary file that contains the commands specifed in the `run` keyword.
```

lucascosti (approved):
Thanks for this PR, @room-27! I've made a minor change to the wording, and I'll merge this in 👍

PR #8411, opened 2021-07-21
Clarify that `always()` has a side effect

CyberShadow:
One straight-forward interpretation of the previous wording was that the function simply always returns `true`, with no other side effects. A reasonable conclusion would then be that, in an `if` expression, `&& always()` is redundant, and can be omitted. However, this in incorrect, as the function does have the side effect of forcing the step to be executed in addition to returning `true`.

### Why:

(This is a relatively minor change, does it need an issue?)

### What's being changed:

Just the description of the `always()` function.

### Check off the following:


ramyaparimi:
@CyberShadow Thanks so much for opening a PR! I'll get this triaged for review :zap:

Adriamo132 (approved):

ivan0117 (approved):
p

ivan0117 (changes requested):
p

ivan0117 (commented):
p

DoctorSky1967:
Help reset pls.

skedwards88 (approved):
🚀 Thanks for this addition! I'll get this merged down for you.

PR #8627, opened 2021-07-28
Set up npm and run npm ci to run husky to get githooks for LFS

heiskr:
Set up npm and run npm ci to run husky to get githooks for LFS


gracepark (approved):
👍

PR #8720, opened 2021-08-02
Remove the wrong note about boolean literals

### LINKED ISSUE #8719: Wrong document for literals of expression in GitHub Action workflow

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#literals

> Boolean literals are not case sensitive, so you can use true or True.

### Additional information

The description is not true. When I tried `True` in an expression as follows,

```yaml
tmp:
  runs-on: ubuntu-latest
  steps:
    - run: echo '${{ True }}'
    - run: echo '${{ False }}'
```

https://github.com/rhysd/actionlint/actions/runs/1089744414

it reported a syntax error:

```
The workflow is not valid. .github/workflows/ci.yaml (Line: 99, Col: 14): Unrecognized named-value: 'True'. Located at position 1 within expression: True .github/workflows/ci.yaml (Line: 100, Col: 14): Unrecognized named-value: 'False'. Located at position 1 within expression: False
```

it means that only lower case of `true` and `false` are correct values for boolean literals. I also tried other cases like `TRUE` or `FALSE` but they also reproduced the error.

rhysd:
### Why:

Closes #8719

### What's being changed:

This PR removes the wrong note about boolean literals.

### Check off the following:


rhysd:
I confirmed the deployment: https://docs-8720--issue-8719.herokuapp.com/en/actions/reference/context-and-expression-syntax-for-github-actions#literals

ramyaparimi:
@rhysd Thanks so much for opening a PR! I'll get this triaged for review :zap: Thanks so much for adding the deployment link 💖

lucascosti (approved):
Thanks a lot for these fixes, @rhysd! 🚀

PR #8786, opened 2021-08-04
Remove advice about Actions if conditional operators

### LINKED ISSUE #8700: Incorrect documentation regarding omitting the expression syntax for if conditionals

* https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions
* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif
* https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif

All the sections linked above contain the following snippet:

> When you use expressions in an `if` conditional, you may omit the expression syntax (`${{ }}`) because GitHub automatically evaluates the `if` conditional as an expression, unless the expression contains any operators. If the expression contains any operators, the expression must be contained within `${{ }}` to explicitly mark it for evaluation.

However, according to my own experience (and a lot of examples I find on the web), the exception (*“unless the expression contains any operators”*) does not seem to be true. For example, the expression of the following `if` is evaluated.

    jobs:
      my-job:
        runs-on: ubuntu-latest
        if: github.event_name == 'push' && github.ref == 'refs/heads/master'

Note that `==` and `&&` are both operators. I think it is pretty confusing that a lot of examples on the web omit `${{ }}` while the documentation claims that it would not work.

lucascosti:
### Why:

Closes https://github.com/github/docs/issues/8700

### What's being changed:

Remove advice about Actions `if` conditional operators. `if` conditions don't need to be enclosed in `${{ }}`. Our examples still show `${{ }}` because it helps avoid yaml escaping issues.

Stage preview: https://docs-8786--lucascosti-patch-1.herokuapp.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

### Check off the following:


martin389 (approved):
Looks good to me 👍

PR #8817, opened 2021-08-05
Use COPY instead of ADD

mostafahussein:
### What's being changed:

Small change in Dockerfile syntax, will make it use `COPY` instead of using `ADD` as long as we are dealing with local files.


ramyaparimi:
@mostafahussein 

91CH (approved):

Payaspx (approved):

gracepark:
Thank you so much for the suggestion - this file is currently used internally by a separate team, so we do not currently update this file. I'll speak internally with the team. Thank you again!

PR #9098, opened 2021-08-17
Clarify that Dependabot does not support Maven as token-only

brrygrdn:
### Why:

A user received a configuration error when attempting to create a Dependabot configuration with a Maven Repository that used only a token for authentication as suggested by the current documentation.

### What's being changed:

On further investigation, we do not actually support Tokens for Maven Repositories at this time as token-only authentication is specific to GitHub Package Registry and unsupported by common Maven Repository specifications.

Since adding support for token-only access to GPR is a feature request we don't plan to solve immediately, let's update the documentation to reflect the current implementation.

### Check off the following:


androideveloper (dismissed):
Looks good! Thanks for the clarification!

ramyaparimi:
@brrygrdn Thanks so much for opening a PR! I'll get this triaged for review :zap:

skedwards88 (commented):
Thank you for this correction! Does the example under [this section](https://docs-9098--brrygrdnupdat-depen.herokuapp.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#registries) (line 498-501) need to change as well?

brrygrdn:
Thanks for spotting that @skedwards88! I've made an update to fix that example as well.

skedwards88 (approved):
LGTM! I'll get this merged down.

L2021 (approved):
It's good

PR #9311, opened 2021-08-25
Update "Restrictions for accessing a cache" - make clear that cache entries are shared between workflows

### LINKED ISSUE #9310: Not clear/correct under "Restrictions for accessing a cache" - that cache entries shared between workflows
### Code of Conduct

https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

Tested this tonight:

- With multiple workflows within a repository
- If a cache entry is created from one workflow under a branch (lets say `main`)....
- From _another_ workflow in the same repository executed against the same branch (e.g. `main`) - the existing cache entry can be fetched successfully.

The current documentation gives the impression that isolation is not only provided between branches, but _workflows_ - which to my surprise, isn't the case.

> Access restrictions provide cache isolation and security by creating a logical boundary between different workflows and branches.

### Additional information

_No response_

magnetikonline:
### Why:

Closes #9310 

Here are the two example workflows used (in the same repository of course) to prove this:

```yaml
name: One

on:
  push:

jobs:
  one:
    runs-on: ubuntu-latest
    steps:
      - run: |
          echo "workflow-one-item" >cache.me
      - uses: actions/cache@v2
        with:
          path: |
            cache.me
          key: cache-key-01
```

```yaml
name: Two

on:
  push:

jobs:
  main:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/cache@v2
        with:
          path: |
            cache.me
          key: cache-key-01
      - run: |
          cat cache.me
```

So basically, triggering workflow `One` will populate a new cache entry of `cache-key-01`. Executing `Two` afterwards will happily pull the `cache-key-01` with the contents `workflow-one-item` right away. There is no isolation or namespacing between the workflows.

I think this is good to know and can be a powerful item for the toolkit - but also could be causing troubles for some if taking the current documentation as absolute.

### What's being changed:

Clearing up the documentation around [Restrictions for accessing a cache](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) to make it clear that sharing of cache entries between workflows within the same repository for the same branch is very much possible.

This is the key line I've added: 

> Multiple workflows within a repository share cache entries. A cache created for a branch within a workflow can be accessed and restored from another workflow for the same repository and branch.

### Check off the following:


ramyaparimi:
@magnetikonline 

ghost (changes requested):
شكرا لكم 
اصلح كل شىء

ethanpalm (approved):
Thanks for creating this PR, @magnetikonline! Your addition looks great. We'll get it merged as soon as possible.

Thanks for taking the time to contribute to the docs!

magnetikonline:
Not a problem @ethanpalm - happy to help. 👍 

Was also good to clarify that this is expected behaviour and not an unintended side-effect!

PR #9426, opened 2021-08-30
Update workflow-commands-for-github-actions.md

### LINKED ISSUE #9919: Actions: Improve "stop-commands" example code
### Code of Conduct

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions

The section [Stopping and starting workflow commands](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#stopping-and-starting-workflow-commands) has an example of using `stop-commands`:

```Shell
echo "::stop-commands::pause-logging"
[...]
echo "::pause-logging::"
```

This example uses a hardcoded resume token, which is not best practice. Also, it's confusing that the command to resume processing is `::pause-logging::`

There's a much better example in this blog post: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

```Shell
# disable command workflow processing
echo "::stop-commands::`echo -n ${{ github.token }} | sha256sum | head -c 64`"

# log untrusted output
echo "untrusted output"

# enable workflow command processing
echo "::`echo -n ${{ github.token }} | sha256sum | head -c 64`::"
```

This creates a resume token based on the GITHUB_TOKEN, so it's different on every workflow run. Doing this makes it almost impossible for the untrusted output to resume command processing.

UPDATE:
An example based on this code would be ideal: https://github.com/actions/runner/issues/807#issuecomment-916236648

It adds a mask for the resume token, so it can never appear in the workflow run logs.

fhammerl:
### Why:

This PR emphasises the importance of keeping your `::stop-commands::` tokens secure.

### Check off the following:


ramyaparimi:
@fhammerl 

OutLAw-Dev (approved):

OutLAw-Dev (dismissed):

fhammerl:
Check out the changes here: https://docs-9426--fhammerl-patch-1.herokuapp.com/en/actions/reference/workflow-commands-for-github-actions#stopping-and-starting-workflow-commands

ramyaparimi:
@fhammerl I have triaged the PR for a writer's review ✨ We are currently in a repo freeze, but we will get this merged as soon as we can 🐬 Thanks a bunch for your patience 💖

content/actions/reference/workflow-commands-for-github-actions.md
```diff
@@ -210,22 +210,35 @@ echo "::add-mask::$MY_NAME"
 
 Stops processing any workflow commands. This special command allows you to log anything without accidentally running a workflow command. For example, you could stop logging to output an entire script that has comments.
 
-### Example stopping workflow commands
+{% warning %}
 
-``` bash
-echo "::stop-commands::pause-logging"
-```
+**Warning:** Make sure the token you're using is randomly generated and different between runs. A **hash** of your `github.token` is a good choice for this.
```
martin389, line 215: Would this suggestion still be accurate?
```suggestion
**Warning:** Make sure the token you're using is randomly generated and unique for each run. A **hash** of your `github.token` is a good choice for this.
```

martin389 (commented):
👋 Thanks for this PR @fhammerl! I've added a small suggestion

content/actions/reference/workflow-commands-for-github-actions.md
```diff
@@ -210,22 +210,35 @@ echo "::add-mask::$MY_NAME"
...
 
-### Example starting workflow commands
+### Example stopping and starting workflow commands
 
-``` bash
-echo "::pause-logging::"
+{% raw %}
+
+```yaml
+workflow-command-job:
```
anywherepilot, line 228: Technically fine but looks a bit off to use four spaces indent for this level and two for all others.

This workflow is valid, but I don't know about the workflow commands so can't sign off on that 😬 

Maybe emphasize that this is a job only and should be inserted into (and further indented to fit into) a workflow file `jobs` section.

martin389: 👋 Thanks @anywherepilot - I've updated the YAML in this commit: https://github.com/github/docs/pull/9426/commits/6479a7ffa6f3819c65dffd2bb4fcb48fdcbd3e26

SirJosh1987 (dismissed):

content/actions/reference/workflow-commands-for-github-actions.md
```diff
@@ -210,22 +210,36 @@ echo "::add-mask::$MY_NAME"
 
 Stops processing any workflow commands. This special command allows you to log anything without accidentally running a workflow command. For example, you could stop logging to output an entire script that has comments.
 
-### Example stopping workflow commands
+To stop the processing of workflow commands, pass a unique token to `stop-commands`. To resume processing workflow commands, pass the same token that you used to stop workflow commands.
```
martin389, line 213: 👋 Hi @fhammerl -- I've added this sentence to further explain how the value is used. What do you think of this?

content/actions/reference/workflow-commands-for-github-actions.md
```diff
@@ -210,22 +210,36 @@ echo "::add-mask::$MY_NAME"
...
+
+```yaml
+jobs:
+  workflow-command-job:
+    runs-on: ubuntu-latest
+    steps:
+      - name: disable workflow commands
+        run: |
+          echo '::warning:: this is a warning'
+          echo "::stop-commands::`echo -n ${{ github.token }} | sha256sum | head -c 64`"
```
rcowsill, line 235: As mentioned in actions/runner/issues/807, the runner currently echoes the whole `stop-commands` line to the log. Until this changes it may be advisable to use `add-mask` to avoid leaking the resume token, eg:
```suggestion
          echo "::add-mask::`echo -n ${{ github.token }} | sha256sum | head -c 64`"
          echo "::stop-commands::`echo -n ${{ github.token }} | sha256sum | head -c 64`"
```

ylemkimon: I don't think using the GitHub token for purposes other than authentication is a good idea. Also, it makes the `stop-commands` token deterministic.

rcowsill: @ylemkimon The idea to hash the github token for use as the resume token comes from [this GitHub security blog post](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/#patching-your-actions-and-workflows)

The documentation is a bit unclear on this, but `github.token` changes every run. From https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret:

> Before each job begins, GitHub fetches an installation access token for the job. The token expires when the job is finished.

rcowsill:
This will fix issue #9919 if it gets merged.

content/actions/reference/authentication-in-a-workflow.md
```diff
@@ -19,7 +19,7 @@ shortTitle: Authentication in a workflow
 
 ## About the `GITHUB_TOKEN` secret
 
-{% data variables.product.prodname_dotcom %} automatically creates a `GITHUB_TOKEN` secret to use in your workflow. You can use the `GITHUB_TOKEN` to authenticate in a workflow run.
+At the start of each workflow run, {% data variables.product.prodname_dotcom %} automatically creates a unique `GITHUB_TOKEN` secret to use in your workflow. You can use the `GITHUB_TOKEN` to authenticate in a workflow run.
```
martin389, line 22: @fhammerl - I've updated this line to explain this feature a bit more clearly. Let me know what you think 👍

content/actions/reference/workflow-commands-for-github-actions.md
martin389: I've added an update to this PR to make this clearer 👍

ylemkimon: Still, the GitHub token is used everywhere, `@actions/checkout` saves it to  git config by default (`persist-credentials: true`), it may be passed to actions. I think a randomly generated string would be better.

ylemkimon: But adversaries getting aceess to the GitHub token would be more serious issue, so I guess it wouldn't be a much problem.

content/actions/reference/authentication-in-a-workflow.md
fhammerl: Agreed, sounds good 👍

content/actions/reference/workflow-commands-for-github-actions.md
rcowsill: > I've added an update to this PR to make this clearer +1
> https://github.com/github/docs/pull/9426/files#diff-dedecb8a8261bc05cad61be2f94b7531543481eab957be90a173c77f09e2df32R22

That change makes the documentation very clear about the fact that the github token is unique per run :+1:

Unfortunately the example code does still leak the resume token. The fact that it's unique per run makes this difficult to exploit, but not impossible.

I'd still like to have the suggested `add-mask` line added to the example.

martin389 (approved):

PR #9916, opened 2021-09-09
Bump update in dependabot and fix bugs 

djibal:
Merge  feature requests to update in and fix bug


djibal (commented):
Review  important elements of these codes in

PR #10039, opened 2021-09-14
Copy GPG key by clip command

ghost:
This will make it easier to copy GPG key instead of manually selecting and copying it.


rouisibra:
Oops sorry this works only on windows

PR #10213, opened 2021-09-21
Remove incorrect warning block

JaredEgolf:
Actions can be configured on GHES 3.1+ on Azure Blob storage as well giving three viable go-forward paths.

### Why:

Because the warning is factually incorrect.  An alternative would be to update the warning text as appropriate.
Similar changes are required on the [enabling-github-actions-with-amazon-s3-storage](https://docs.github.com/en/enterprise-server@3.1/admin/github-actions/enabling-github-actions-for-github-enterprise-server/enabling-github-actions-with-amazon-s3-storage) page for 3.1 and 3.2.  (I have not verified the same on GHES 3.0.)

### Check off the following:


ramyaparimi:
@JaredEgolf 

Marksmith1987 (approved):

lucascosti:
👋 @JaredEgolf, thanks for opening this PR!

> Actions can be configured on GHES 3.1+ on Azure Blob storage as well giving three viable go-forward paths.

🤔 I'm a little confused about why this warning needs removal. It states:

> The only supported **S3 storage providers** are Amazon S3 and MinIO Gateway for NAS.

Azure Blob Storage [has its own article here](https://docs.github.com/en/enterprise-server@3.1/admin/github-actions/enabling-github-actions-for-github-enterprise-server/enabling-github-actions-with-azure-blob-storage). The warning is meant to convey that we don't support using any other S3 'compatible' storage providers other than the ones listed.

Please let me know if I've misunderstood 🙂

JaredEgolf:
@lucascosti I completely see your point!  
When I look at it from the perspective of "requirements" and only see S3 storage providers and no mention of the Azure alternative it makes me feel mislead.  On its face its entirely factual and correct.  I think perhaps it's just my preference that the warning block be updated to explain ALL options that can be "slotted" into the requirement but I agree with you that I was incorrect and the way that it reads now is still factually correct.  🙇🏼

lucascosti:
thanks, @JaredEgolf. I think the source of the issue is because you're looking at an S3-specific article, and all the prereqs are specific to the provider the article is written for. In that warning we link to the [Getting started with GitHub Actions for GitHub Enterprise Server](https://docs.github.com/en/enterprise-server@3.1/admin/github-actions/enabling-github-actions-for-github-enterprise-server/getting-started-with-github-actions-for-github-enterprise-server#external-storage-requirements) article, which is the broader article that explains common prereqs across all configs and lists all the supported storage providers.

I'd hesitate to make the S3-specific warning longer to list all other supported storage providers, especially as the list may grow in the future.

PR #10230, opened 2021-09-22
Create runner-name-description.md

fhammerl:
### Is this issue for a product update or release that is not yet available to customers?

No? Then please continue. Thank you!

### What article(s) is affected?

`runner` context

https://docs-10230--add-runner-name-de.herokuapp.com/en/actions/learn-github-actions/contexts#runner-context

### Describe the desired docs changes and the problem they solve for customers.

I’ve added a new context variable (`runner.name`) to the runner. I’d like to update the documentation to tell users about this new variable.

I’d like to add a row above `runner.os`, called `runner.name`. 
Description: The name of the specific runner executing the run.

### Who does this affect?

People writing workflows or GitHub Actions

### What is the impact to users?

Users writing workflows or actions can now use a new context variable called `runner.name`.


fhammerl:
@martin389 thanks for the tips on how to add this change.

The variable `runner.name` is currently NOT included in any of the examples. Let me know if this is something I should do as part of this PR.

ramyaparimi:
@fhammerl 

content/actions/learn-github-actions/contexts.md
```diff
@@ -130,6 +130,7 @@ The `runner` context contains information about the runner that is executing the
 
 | Property name | Type | Description |
 |---------------|------|-------------|
+| `runner.name` | `string` | {% data reusables.actions.runner-name-description %} |
```
lucascosti, line 133: Hey @fhammerl, thanks heaps for this PR! 🙇‍♂️ 

Normally, we don't create reusables for text that is only used in one place. The reusables for the other contexts in this table are there because those contexts are the same as similarly named environment variables that we document here: https://docs-10230--add-runner-name-de.herokuapp.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

I did a quick test, and it looks like the `runner.name` is available as the `$RUNNER_NAME` default environment variable.

Could you please add the environment variable to that article and use the same reusable for the description?

lucascosti (changes requested):

BehzadLotfiNoshahr:
Good

content/actions/learn-github-actions/contexts.md
fhammerl: Thanks, @lucascosti!

Of course, added the env var.

As a sidenote, should we mention that in the case of Hosted Runners, `runner.name` resolves to 'Hosted Runner' (or 'Hosted Runner {runnerIndex}' in case you have more than one in your pool)? 

Looking at the docs like they are, my first question would be 'So what's `runner.name` in case of Hosted Runners?'.

lucascosti: Thanks, @fhammerl! Yep, I agree; feel free to add that to the description 👍

lucascosti: Hmm, I went to add this change, but it looks like my tests the hosted runner name is "GitHub Actions <number>". I think we'll leave this for now, and follow-up later.

lucascosti (approved):
Looks good!

PR #10997, opened 2021-10-09
Removed outdated refrence to git shell

placoderm:
According to this:
https://github.com/desktop/desktop/issues/340
https://github.com/desktop/desktop/issues/340#issuecomment-361716966

git shell is no longer included in the Desktop version.

It appears that the text in the item following the one I edited is auto generated. Now it says:
> If you are using Git Bash, turn on ssh-agent:

The first part, "If you are using Git Bash" is now superfluous since the alternative of using the desktop app is no longer there.

This is my very first ever pull request! Please be patient with me.

### Why:
Outdated/incorrect information


placoderm:
> 
> 
> https://github.com/github/docs/blob/main/actions/reference/workflow-syntax-for-github-actions#permissions

I'm getting a 404 for that page.

felicitymay (approved):
@placoderm - many apologies for the delay in getting to your PR. It looks as if we may have had a triage automation hiccup 😞 

Your change looks like a great call. Thanks for raising a PR to update the docs, so that they don't confuse users with outdated information. 💖

PR #11288, opened 2021-10-18
Require pull request before merging section (first time contributor)

joshbruce:
The UI doesn't say "Require pull request review before merging" it says "Require pull request before merging," which seems to be leading to some confusion on the feature.

I'm trying to protect `main` from being pushed to directly by anyone, including myself (admin). For my projects it's mainly just me and my public stuff isn't really popular. I sometimes find myself working in `main` unexpectedly (after deleting a branch and pulling in the update). Creating a branch rule and enabling require PR seems to be the suggested or desired behavior to achieving this restrict push:

- https://stackoverflow.com/questions/46146491/prevent-pushing-to-master-on-github
- https://github.community/t/how-to-disable-require-approvals-for-a-protected-branch-with-require-a-pull-request-before-merging/205555

As of this writing, the combination of "Require pull request before merging" and "Include administrators" can cause me to be unable to merge anything I submit; because I can't approve my own PRs (good policy - just not conducive to what I'm hoping to accomplish - no push to main).

Another option seems to be modifying Git directly, which seems like a workaround given the "Require pull request before merging" feature (as named) seems purpose built to handle this:

- https://hiltonmeyer.com/articles/protect-git-branch-and-prevent-master-push.html

Being able to uncheck the disabled checkbox for "Required approving reviews" would be helpful; or, adding a 0 option from the drop down, which would make the feature able to do what its name implies (a PR is required before merge - no push allowed). Or, updating the UI label to be "Require pull request *review* before merging" - which wouldn't solve the desire to have a way to require a pull request.

Apologies if this feels like a feature request inside of a documentation recommendation. The purpose of the submission is to ensure the documentation text matches the current UI text (or maybe get the UI updated to match the documentation) - and to help lead users to a solution to the problem of wanting to require pull requests on certain branches; with or without a review and approval.

### Why:

Documentation doesn't match UI and doesn't lead to resolution of a use case. (see above)

### What's being changed:

Screen Shot 2021-10-18 at 6 46 07 PM

### Check off the following:


lecoursen (commented):
Thanks @joshbruce, it looks like the UI text changed since we last updated this article. I'll get this merged to fix that up after applying my suggestions. I suggest sending your product feedback via https://support.github.com! 😊

lecoursen (approved):

joshbruce (commented):
Seems fair.

PR #11329, opened 2021-10-20
CodeQL Python: Add note about `pip install -e .` with pip 21.3

RasmusWL:
### Why:

We encountered a problem with a manual setup of python dependencies for CodeQL ([reference](https://github.com/Cog-Creators/Red-DiscordBot/pull/5412)) and I wanted to get the solution into the official docs.

### What's being changed:

- Add note about using `pip install -e .` instead of `pip install .`

### Check off the following:


lee-dohm:
👋 @RasmusWL ... it looks like @mchammer01 has an open question here. Is this something you'll be able to continue working on?

RasmusWL:
Sorry about the very late reply to this. We've now added a fix to the CodeQL Python extractor (v 5.19), so once that is released (within the new few weeks), it should not be a problem any more.

I do not think we need a temporary addition to the docs, so I'm going to close this PR.

PR #11841, opened 2021-11-10
distinguish API before and after GHES 3.2

pberthonneau:
### Why:

I made 2 changes in the sentence for ghes:  
   
   To monitor results from secret scanning across your **private** repositories **or your organization**, you can use the secret scanning API. For more information about API endpoints, see "Secret scanning."  
  
1. **or your organization** is only valid from 3.2 because before there is not secret scanning API at org level.

2. I have also removed the word **private** because for ghes with advanced security, secret scanning work the same way on public, internal and private repos. I assume this is the same for ghec.  
  
  Note that a similar mention of **private** in the [API](https://docs.github.com/en/enterprise-server@3.2/rest/reference/secret-scanning) looks wrong to me. I'm using it with public repositories.

### Check off the following:

- [x] I have been able to review my changes in staging (look for "Automatically generated comment" and click **Modified** to view your latest changes).


ramyaparimi:
@pberthonneau 

content/code-security/secret-scanning/about-secret-scanning.md
```diff
@@ -72,7 +72,7 @@ For more information about viewing and resolving {% data variables.product.prodn
 Repository administrators and organization owners can grant users and teams access to {% data variables.product.prodname_secret_scanning %} alerts. For more information, see "[Managing security and analysis settings for your repository](/github/administering-a-repository/managing-security-and-analysis-settings-for-your-repository#granting-access-to-security-alerts)."
 
 {% ifversion fpt or ghes > 3.0 or ghec %}
-To monitor results from {% data variables.product.prodname_secret_scanning %} across your private repositories or your organization, you can use the {% data variables.product.prodname_secret_scanning %} API. For more information about API endpoints, see "[{% data variables.product.prodname_secret_scanning_caps %}](/rest/reference/secret-scanning)."{% endif %}
+To monitor results from {% data variables.product.prodname_secret_scanning %} across your repositories{% ifversion ghes > 3.1 %} or your organization{% endif %}, you can use the {% data variables.product.prodname_secret_scanning %} API. For more information about API endpoints, see "[{% data variables.product.prodname_secret_scanning_caps %}](/rest/reference/secret-scanning)."{% endif %}
```
felicitymay, line 75: Here, I think we need to version "private" for Free, pro, team and for GHEC instead of removing it. I'll confirm with the secret scanning team.

```suggestion
To monitor results from {% data variables.product.prodname_secret_scanning %} across your {% ifversion fpt or ghec %}private {% endif %}repositories{% ifversion ghes > 3.1 %} or your organization{% endif %}, you can use the {% data variables.product.prodname_secret_scanning %} API. For more information about API endpoints, see "[{% data variables.product.prodname_secret_scanning_caps %}](/rest/reference/secret-scanning)."{% endif %}
```

felicitymay (commented):
@pberthonneau - thanks for opening this PR ✨ 

> or your organization is only valid from 3.2 because before there is not secret scanning API at org level.

You're right. This is a bug in the current docs and your fix for this looks great 👍🏻 

> I have also removed the word private because for ghes with advanced security, secret scanning work the same way on public, internal and private repos. I assume this is the same for ghec.

I think you're also right that the word "private" here is wrong for GHES, but I think that it is correct for GHEC where public repositories are treated differently. I'll check with the secret scanning team. If I'm right, then I'll apply my suggested change to your PR and look at whether we can also get the API reference docs updated for GHES.

felicitymay (approved):
I've checked the preview: [About secret scanning](https://docs-11841--patch-2.herokuapp.com/en/enterprise-server@3.1/code-security/secret-scanning/about-secret-scanning) and this now accurately:

- Mentions organizations only for GHES 3.2+ and GitHub.com
- Mentions private repositories only for GitHub.com versions

I've opened an internal issue for the other bug you spotted. The API reference docs for GHES should not mention "private repositories" 🙇🏻‍♀️

felicitymay:
The failing test is a known issue and unrelated to these changes, so this PR can be merged if all other tests pass.

PR #12095, opened 2021-11-17
update name of secret

### LINKED ISSUE #12094: The documentation suggests an invalid value for the name of a secret
### Code of Conduct

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions

The article contains the following example configuration:

```
steps:
  - name: My first action
    env:
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      FIRST_NAME: Mona
      LAST_NAME: Octocat
```

However, when attempting to follow this example, I was presented with an error in the UI as follows:

"Failed to add secret. Secret names must not start with GITHUB_."

If this is not a valid naming convention for secrets, then it shouldn't be used in the documentation.

### Additional information

error when attempting to create a secret

lpmi-13:
When attempting to follow the guide and set a value for a secret called `GITHUB_SECRET`, the UI presented the following error:

"Failed to add secret. Secret names must not start with GITHUB_."

If this is not a valid naming convention for secrets, then we shouldn't use it in the documentation.

### Why:

Closes https://github.com/github/docs/issues/12094

### What's being changed:

The name of the secret in the example.

### Check off the following:


ramyaparimi:
@lpmi-13 

content/actions/learn-github-actions/workflow-syntax-for-github-actions.md
```diff
@@ -1051,7 +1051,8 @@ Public actions may specify expected environment variables in the README file. If
 steps:
   - name: My first action
     env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # note: github secrets may not begin with `GITHUB_`
```
felicitymay, line 1054: I can see why you added the note, but I think you probably wouldn't have thought to prefix the token name with `GITHUB_` if we hadn't chosen a poor example here. I think that now you've fixed the example, this note is unnecesary.

```suggestion
```

felicitymay (commented):
Hi @lpmi-13 👋🏻 

Thank you so much for this update to the docs. You're right - this is a bad example. I'm not sure how it slipped through the review process 🙈 

I'm going to make one change, but otherwise this looks good to merge 👍🏻

felicitymay (approved):

felicitymay:
@lpmi-13 - I did a quick check of the rest of the docs and found some other places where we used the same example. 🤔 

I'm checking internally that your change is correct before merging the change - just in case I've missed something important. It's possible that there are other places we should make the same fix.

felicitymay:
@lpmi-13 - I'm glad that I checked with the Actions docs team 😌 

It turns out that this example uses the special `GITHUB_TOKEN` that is created automatically for each workflow run. This is explained here: [Automatic token authentication](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret).

I'm glad that you raised this PR to highlight the confusion here 🙇🏻‍♀️ 
I'm going to report this to the team who maintain this area of the product and suggest that provide a tailored error message if someone tries to create a secret called `GITHUB_TOKEN`.

lpmi-13:
A different error message would be very helpful. Thanks for the quick response on this 👍

PR #12378, opened 2021-11-26
Add documentation for conditional composite steps (`runs.steps[*].if`)

fhammerl:
### Why:

People creating composite actions could not use conditional steps before. Changing the docs would make this change more explicit.

### What's being changed:

By merging https://github.com/actions/runner/pull/1438, we started supporting conditional composite steps.

Added a new paragraph: https://docs-12378--composite-action-s.herokuapp.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsif

### Check off the following:


ramyaparimi:
@fhammerl 

content/actions/creating-actions/metadata-syntax-for-github-actions.md
```diff
@@ -259,6 +259,36 @@ For more information, see "[`github context`](/actions/reference/context-and-exp
 **Required** The shell where you want to run the command. You can use any of the shells listed [here](/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell). Required if `run` is set.
 {% endif %}
 
+#### `runs.steps[*].if`
+
+**Optional** You can use the `if` conditional to prevent a step from running unless a condition is met. You can use any supported context and expression to create a conditional.
```
thboop, line 264: `You can use any supported context and expression to create a conditional.` This is easy for me to understand because i'm really close to the code, but for users who aren't that familiar with actions they may not know that we specifically do not allow you to access the `secrets` in composite actions, would it be good to call our or mention that here again just for clarity?

content/actions/creating-actions/metadata-syntax-for-github-actions.md
```diff
@@ -259,6 +259,36 @@ For more information, see "[`github context`](/actions/reference/context-and-exp
...
+
+{% data reusables.github-actions.expression-syntax-if %} For more information, see "[Expressions](/actions/learn-github-actions/expressions)."
+
+**Example: Using contexts**
+
+ This step only runs when the event type is a `pull_request` and the event action is `unassigned`.
+
+ ```yaml
+steps:
+  - name: My first step
```
thboop, line 274: nit: consider removing the `-name` tag, it doesn't do a whole lot in composite actions.

content/actions/creating-actions/metadata-syntax-for-github-actions.md
```diff
@@ -259,6 +259,36 @@ For more information, see "[`github context`](/actions/reference/context-and-exp
...
+
+**Example: Using status check functions**
+
+The `my backup step` only runs when the previous step of a composite action fails. For more information, see "[Expressions](/actions/learn-github-actions/expressions#job-status-check-functions)."
+
+```yaml
+steps:
+  - name: My first step
+    uses: octo-org/action-name@main
+  - name: My backup step
```
thboop, line 287: Might be cool to show the drawn out example of `github. action_status` and say this is functionally equivalent to `if: ${{ github.action_status == 'failure' }}

content/actions/creating-actions/metadata-syntax-for-github-actions.md
```diff
@@ -259,6 +259,36 @@ For more information, see "[`github context`](/actions/reference/context-and-exp
...
+ ```yaml
+steps:
+  - name: My first step
+    if: {% raw %}${{ github.event_name == 'pull_request' && github.event.action == 'unassigned' }}{% endraw %}
+    run: echo This event is a pull request that had an assignee removed.
+```
+
+**Example: Using status check functions**
+
+The `my backup step` only runs when the previous step of a composite action fails. For more information, see "[Expressions](/actions/learn-github-actions/expressions#job-status-check-functions)."
```
thboop, line 281: Might be worth adding a footnote or something to this [page](https://docs.github.com/en/actions/learn-github-actions/expressions#job-status-check-functions) to indicate that in composite actions these refer to the success/failure of the composite action not the job

fhammerl: I agree. Should we also distinguish between  composite `step-if` (depends on composite action's success)  from `pre-if` and `post-if` (depends on job's success)?

What do you think about this:

Rename `Job Status Check Expressions` to `Status Check Expressions`, explain that these expressions can be deconstructed using `job.status` in a `job.step` and `github.action_status` in a `composite.step`, respectively, and show some examples like you mentioned: `if: ${{ github.action_status == 'failure' }}`

content/actions/creating-actions/metadata-syntax-for-github-actions.md
fhammerl: Do we say somewhere else that `secrets.*` is not supported in composite actions? I feel the [main](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) composite page would be a better place to make that clear.

content/actions/creating-actions/metadata-syntax-for-github-actions.md
thboop: Sounds good to me

content/actions/creating-actions/metadata-syntax-for-github-actions.md
thboop: Maybe here? https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

Doesn't have to be done in this pr

skedwards88: > 

It would be super helpful to document what context are not supported in [the contexts article](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability) (and then to link to it here and from the [creating a composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) article).

Since you are a Hubber, you can [open an issue in docs-content](https://github.com/github/docs-content/issues/new/choose) and the docs team can work with you to document this info. 

(cc @martin389 for the Actions docs improvement project)

content/actions/creating-actions/metadata-syntax-for-github-actions.md
skedwards88: If a user is using a third party action, will it be obvious to them whether the action is a composite action?

Or does this apply to [all small-a actions](https://docs.github.com/en/actions/creating-actions/about-custom-actions#types-of-actions) (not just composite actions)?

skedwards88 (commented):
💖 Thanks for opening this PR! I left some preliminary comments before realizing that more changes are probably incoming. Feel free to ping me for another review once it is ready.

content/actions/creating-actions/metadata-syntax-for-github-actions.md
fhammerl: @skedwards88 This only applies to composite actions, other actions can't have conditional steps.

The idea is that maintainers of the composite action should only have to consider the action itself being successful, not the job executing it.

For example, let's say a job is failing, but its last step calls a composite action with `if:always()`. We want this composite action to start out with a state of `success` and track changes to its state independent of the job.

fhammerl:
Deconstruction explained: https://docs-12378--composite-action-s.herokuapp.com/en/actions/learn-github-actions/expressions#evaluate-status-explicitly

Step-if: https://docs-12378--composite-action-s.herokuapp.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsif

Pre-if and Post-if: https://docs-12378--composite-action-s.herokuapp.com/en/actions/creating-actions/metadata-syntax-for-github-actions#pre

@thboop Let me know what you think

content/actions/learn-github-actions/expressions.md
```diff
@@ -317,6 +317,28 @@ steps:
...
+
+Instead of using one of the methods above, you can evaluate the status of the job or composite action that is executing the step directly:
+
+#### Example for workflow step
+
+```yaml
+steps:
+  ...
+  - name: The job has failed
+    if: {% raw %}${{ job.status == 'failure' }}{% endraw %}
```
thboop, line 330: maybe add a yaml comment that says `# this is the same as failure()

thboop (dismissed):
LGTM

skedwards88 (approved):
Thanks for these updates! I'll get this merged down for you.

PR #12454, opened 2021-11-29
Making it easier to understand dependency review for public versus private repos

NickLiffen:
### Why:

Making it easier to understand where Dependency Review is available. This has caused some confusion. Hopefully, the proposed changes make it easier to understand but I welcome feedback 👍 

### Check off the following:


ramyaparimi:
@NickLiffen 

NickLiffen:
Thanks, @ramyaparimi 💯 I also welcome someone better at grammar than me to propose changes 

For example, another way of maybe phrasing this is:

```
Dependency review is available for all public repositories, as well as private repositories owned by organizations where ...
```

I very much welcome feedback on the proposed changes 👍

data/reusables/gated-features/dependency-review.md
```diff
@@ -1,3 +1,3 @@
-{% ifversion fpt or ghec %}Dependency review is available for all public repositories and for private repositories owned by organizations where {% data variables.product.prodname_GH_advanced_security %} is enabled.{% endif %}
+{% ifversion fpt or ghec %}Dependency review is available for all public repositories. As well as private repositories owned by organizations where {% data variables.product.prodname_GH_advanced_security %} is enabled.{% endif %}
```
mchammer01, line 1: I think that the initial text is clear but I'm open to updates if this caused confusion with our customers. What about this?
(Will wait to hear from @jhutchings1 before applying this suggestion)
```suggestion
{% ifversion fpt or ghec %}Dependency review is available for all public repositories, as well as well as private repositories owned by organizations where {% data variables.product.prodname_GH_advanced_security %} is enabled.{% endif %}
```

mchammer01 (commented):
@NickLiffen - thanks for suggesting an improvement. 
I've added a suggestion, let me know what you think 😃

data/reusables/gated-features/dependency-review.md
NickLiffen: yeah it caused some confusion, to be fair I even read it wrong 😮‍💨 I think the purpose is making it a little more apparent. I read this as you needed a GHAS licence to run this on both public repos and private repos. I think it makes sense to decouple it, like you have ❤️

jhutchings1: I like @mchammer01 's suggestion above. The content is accurate in regards to feature availability, but if we can make it more clear, we should.

NickLiffen: thanks, @jhutchings1 👍 @mchammer01 are we good to commit your suggestion, please? I think this pleases everyone ❤️ thank you all.

mchammer01:
@NickLiffen - thank you for highlighting that this sentence was confusing. I hadn't realized until you mentioned it.
Will apply the suggestion and merge this PR.
Thanks again 💖

mchammer01 (approved):
LGTM ✨

PR #12468, opened 2021-11-30
Clarify that webhook deliveries receive all commits in a push

danielcompton:
### Why:

Closes [issue link] (N/A)

### What's being changed:

I was reading the docs for https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push and found it a bit tricky to understand this line:

> An array of commit objects describing the pushed commits. (Pushed commits are all commits that are included in the compare between the before commit and the after commit.) The array includes a maximum of 20 commits. If necessary, you can use the Commits API to fetch additional commits. This limit is applied to timeline events only and isn't applied to webhook deliveries.

I've made an update which I _think_ is correct, but please let me know if this is not the correct interpretation. I'm assuming that Timeline events refers to [Issue Timeline events](https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types)?

### Check off the following:


ramyaparimi:
@danielcompton 

vgrl:
Hi :wave: @danielcompton, thanks for raising a PR to address the issue, and for your patience while we reviewed it.

I agree the language used when describing the `commits` key is confusing.

To answer your question in the OP:

> I've made an update which I _think_ is correct, but please let me know if this is not the correct interpretation. I'm assuming that Timeline events refers to [Issue Timeline events](https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types)?

Yes, that's partially correct. Timeline events are [Issue events](https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types) and also any activity that can be retrieved using the [Events API](https://docs.github.com/en/rest/reference/activity#events). In this case, there is only one timeline event associated with a [`push` event object](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#push), and that is the [`PushEvent` event type](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pushevent).

We've tried to explain the difference for the specific case above, that is between the `payload` property of event types, and webhook payloads for `push` events, in the below paragraph from [Webhook payload object common properties](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties), but I can appreciate this wording may be just as confusing:

> The unique properties for a webhook event are the same properties you'll find in the `payload` property when using the Events API. One exception is the `push` event. The unique properties of the `push` event webhook payload and the `payload` property in the Events API differ. The webhook payload contains more detailed information.

So the behaviour we're attempting to describe and clarify here is that if more than 20 commits were made in a single push, then not all commits would be visible in the `push` event `payload` property that the Events API returns, but all commits would be visible in the `push` event webhook payload.

What I suggest we do to made this distinction clearer, is to remove entirely the mention of the number of commits limit in the Webhooks events and payloads article. We can however leave the mention in the [`PushEvent` - Events API](https://docs.github.com/en/developers/webhooks-and-events/events/github-event-types#pushevent) article.

content/developers/webhooks-and-events/webhooks/webhook-events-and-payloads.md
```diff
@@ -1041,7 +1041,7 @@ Key | Type | Description
...
-`commits`|`array` | An array of commit objects describing the pushed commits. (Pushed commits are all commits that are included in the `compare` between the `before` commit and the `after` commit.) The array includes a maximum of 20 commits. If necessary, you can use the [Commits API](/rest/reference/repos#commits) to fetch additional commits. This limit is applied to timeline events only and isn't applied to webhook deliveries.
+`commits`|`array` | An array of commit objects describing the pushed commits. (Pushed commits are all commits that are included in the `compare` between the `before` commit and the `after` commit.) For timeline events only, the array includes a maximum of 20 commits. If necessary, you can use the [Commits API](/rest/reference/repos#commits) to fetch additional commits. Webhook deliveries will receive an array containing all of the pushed commits.
```
vgrl, line 1044: As I've [suggested](https://github.com/github/docs/pull/12468#issuecomment-1012681029), let's remove entirely the commit limit language from the description, in this article.
```suggestion
`commits`|`array` | An array of commit objects describing the pushed commits. (Pushed commits are all commits that are included in the `compare` between the `before` commit and the `after` commit.)
```

vgrl (changes requested):
This change should hopefully remove some of the confusion you've highlighted. This PR is good to merge once you've committed my suggestion.

vgrl (approved):
🚀 ✨

PR #12688, opened 2021-12-06
Updated concurrency example to cancel any in-progress job or run

jpbnetley:
changed `group: ${{ github.head_ref }}`  to ${{ github.head}}`

### Why:

I copied the example mentioned where it states `group: ${{ github.head_ref }}` but have found that didn't work.
When I changed it to `group: ${{ github.head}}` the work flow worked as expected.
IE: cancelled the older job and starting the new one.

The error message I got before the change was: `The workflow is not valid. .github/workflows/main.yml (Line: 12, Col: 10): Unexpected value ''`

Closes [issue link]

### What's being changed:
Under the section `concurrency example to cancel any in-progress job or run`
changed `group: ${{ github.head_ref }}`  to `${{ github.head}}`

### Check off the following:


ramyaparimi:
@jpbnetley 

ethomson:
Thanks for the PR @jpbnetley!  In your case, I suspect that you weren't running the job against a pull-request trigger.  `head_ref` is only set for PR runs.

I think that generalizing this - as you've done - is the right move, now this will work with PR or regular builds and so may be a better example.  I might suggest that we add a bit more explanation about what `${{ github.ref }}` is and how it limits concurrency to the single branch.  Would you be willing to add a little more explanation there?

ethomson (approved):
Approved with optional feedback.

jpbnetley:
> Thanks for the PR @jpbnetley! In your case, I suspect that you weren't running the job against a pull-request trigger. `head_ref` is only set for PR runs.
> 
> I think that generalizing this - as you've done - is the right move, now this will work with PR or regular builds and so may be a better example. I might suggest that we add a bit more explanation about what `${{ github.ref }}` is and how it limits concurrency to the single branch. Would you be willing to add a little more explanation there?

Hi @ethomson, yes you are correct.
I had the trigger set to the branch trigger. IE
```
push:
    branches: [master]
```

Thanks for the detailed explanation.
After you pointed out what the difference between is betweeb `${{ github.ref }}` and `${{ github.head_ref }}`, I found the declarations of the Environment variables [here](https://docs.github.com/en/actions/learn-github-actions/environment-variables) on the GitHub help page.

> `GITHUB_REF`: The branch or tag ref that triggered the workflow. For example, refs/heads/feature-branch-1. If neither a branch or tag is available for the event type, the variable will not exist.

> `GITHUB_HEAD_REF`: Only set for pull request events. The name of the head branch.

So I am thinking if I update the description [here](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-using-concurrency-to-cancel-any-in-progress-job-or-run) under the section `Example: Using concurrency to cancel any in-progress job or run`

I could maybe give an example for `on pull request`, and link to the documentation explaining what the environment variables mean  [here](https://docs.github.com/en/actions/learn-github-actions/environment-variables)

And then the second example would be `on branch trigger`, with the example as I mentioned in this pull request.

Do you recommend I make the above mentioned updates to this pull request, or is a separate pull request preferred?

skedwards88 (approved):
Thanks for your thoughtful replies on this! Generalizing this example will be quite helpful to users. Our repos are frozen until early January, but we will get this merged for you once they are unfrozen.

> Do you recommend I make the above mentioned updates to this pull request, or is a separate pull request preferred?

I would recommend making the additional updates in a separate PR. That will enable this PR to merge in case you don't have time to make the updates until a later date.

PR #12722, opened 2021-12-07
Document how to use secrets with `if:` conditionals in GitHub Actions workflows

### LINKED ISSUE #6861: Document how to skip a step if a secret isn't present

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions
https://docs.github.com/en/actions/reference/encrypted-secrets

* Something in the `if` set should mention that `secrets.` isn't available
* Something in the context set should mention that `secrets.` isn't available in `if`
* Something in the three should explain how to [work around this](https://github.com/actions/runner/issues/953#issuecomment-821537265)

### Additional information

It is common for projects to include a secret for deploying things and an action that uses that secret.

When one forks that repository, the secret won't be present and the workflow will break.

This experience is incredibly painful.

With just a bit of extra code in a workflow, the ❌ 's go away. But, w/o documentation on how to do it, 99% of workflow authors won't.

Here's an example which fixed one (of a number of failing workflows):
https://github.com/jsoref/argo-cd/commit/ae52e407e6b36d2ee65a20abe30d8065aa3d73cf

```yaml
      - name: deploy
        if: ${{ github.event_name == 'push' && env.HAVE_PERSONAL_TOKEN == 'true' }}
        uses: peaceiris/actions-gh-pages@v2.5.0
        env:
          HAVE_PERSONAL_TOKEN: ${{ secrets.PERSONAL_TOKEN != '' }}
          PERSONAL_TOKEN: ${{ secrets.PERSONAL_TOKEN }}
```

br3ndonland:
### Why:

Closes github/docs#6861

Users have requested documentation on how to use secrets with `if:` conditionals, especially how to skip a GitHub Actions workflow step if a secret isn't present.

This PR is based on the helpful guidance provided in https://github.com/github/docs/issues/6861#issuecomment-870757186.

Related to this, as far as I know, the docs don't mention the default return value for an expression referencing an unset secret. Based on trial and error, the return value of `${{ secrets.SECRET_IS_NOT_SET }}` appears to be an empty string. I've been digging around [actions/runner](https://github.com/actions/runner) to try to find where this default value is set, but I'm not sure. Maybe in [`SecretUtility.cs`](https://github.com/actions/runner/blob/HEAD/src/Sdk/Common/Common/Utility/SecretUtility.cs)?

### What's being changed:

- Add a complete workflow example to [`jobs.<job_id>.steps[*].if`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepsif), demonstrating how to skip a step if a secret is not present
- Add an explanation to "[Using encrypted secrets in a workflow](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow)"
- Cross-reference the two pages

### Check off the following:


ramyaparimi:
@br3ndonland 

danielray1 (commented):
I approve these changes on which I do believe it's rite an will work as it should

danielray1 (commented):
I do think this is the best way to handle it

Dorota1661:
❤️

sob., 11 gru 2021, 00:16 użytkownik danielray1 ***@***.***>
napisał:

Dorota1661:
Dziękuję

Awan27091987:
All delete device secre all

Dorota1661:
1772

Dorota1661:
12772

skedwards88:
@br3ndonland Thank you for this addition! I think this info will be super helpful to users. I don't have permissions to edit your PR to resolve the merge conflict. The file has been renamed to `content/actions/using-workflows/workflow-syntax-for-github-actions.md`, so if you move your addition to that file, I think that should resolve the conflict. Once you resolve the conflict, we'll get this reviewed and merged for you.

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -342,6 +342,34 @@ steps:
...
+name: Run a step if a secret has been set
+on: push
+jobs:
+  my-jobname:
+    runs-on: ubuntu-latest
+    env:
+      SECRET_IS_SET: ${{ secrets.SECRET_IS_SET }}
+      SECRET_IS_NOT_SET: ${{ secrets.SECRET_IS_NOT_SET }}
+    steps:
+      - if: ${{ env.SECRET_IS_SET }}
```
ethomson, line 362: This will be truthy, but I wonder if an explicit comparison to an empty string would really drive the point home?

br3ndonland: Good suggestion @ethomson, thank you. -> aa53817

Rather than referencing two secrets (`${{ secrets.SECRET_IS_SET }}` and `${{ secrets.SECRET_IS_NOT_SET }}`), I rewrote this section to reference a single secret (`${{ secrets.SECRET_IS_SET }}`), and updated the `if:` conditionals to compare with empty strings as suggested. Hopefully this is a little clearer.

br3ndonland: Updated again to use same environment variable and secret names as the cross-referenced "Encrypted secrets" page (e0b2cfb).

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -342,6 +342,31 @@ steps:
     uses: actions/heroku@1.0.0
 ```
 
+### Example: Using secrets
```
hubwriter, line 345: To match level of previous subheadings:

```suggestion
#### Example: Using secrets
```

hubwriter (approved):
@br3ndonland - This looks great. Thanks for making this change. 

The example is really good. I'm sure this will help a lot of people. 👍 

I'll get this merged/published.

m-kuhn:
I think this would deserve a hint on the implication that the secret is leaked to every step within the job, regardless if it's required for a certain step or not (hence violating the principle of least privilege).

PR #13418, opened 2021-12-27
building-and-testing-ruby.md - remove wildcard, version, clarify x.0

dentarg:
### Why:

The wildcard doesn't work in setup-ruby: https://github.com/ruby/setup-ruby/issues/251

### What's being changed:

Docs on how to build and test with Ruby on GitHub Actions

### Check off the following:

Credits to @MSP-Greg for these changes!

### DIFF
--- a/content/actions/automating-builds-and-tests/building-and-testing-ruby.md
+++ b/content/actions/automating-builds-and-tests/building-and-testing-ruby.md
@@ -65,7 +65,7 @@ jobs:
 
 ## Specifying the Ruby version
 
-The easiest way to specify a Ruby version is by using the `ruby/setup-ruby` action provided by the Ruby organization on GitHub. The action adds any supported Ruby version to `PATH` for each job run in a workflow. For more information see, the [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby).
+The easiest way to specify a Ruby version is by using the `ruby/setup-ruby` action provided by the Ruby organization on GitHub. The action adds any supported Ruby version to `PATH` for each job run in a workflow. For more information and available Ruby versions, see [`ruby/setup-ruby`](https://github.com/ruby/setup-ruby).
 
 Using Ruby's `ruby/setup-ruby` action is the recommended way of using Ruby with GitHub Actions because it ensures consistent behavior across different runners and different versions of Ruby.
 
@@ -87,13 +87,13 @@ Alternatively, you can check a `.ruby-version` file  into the root of your repos
 
 ## Testing with multiple versions of Ruby
 
-You can add a matrix strategy to run your workflow with more than one version of Ruby. For example, you can test your code against the latest patch releases of versions 2.7, 2.6, and 2.5. The 'x' is a wildcard character that matches the latest patch release available for a version.
+You can add a matrix strategy to run your workflow with more than one version of Ruby. For example, you can test your code against the latest patch releases of versions 3.0, 2.7, and 2.6.  YAML parsers will truncate trailing zeros from numeric values, and often convert floats to integers.  So, an unquoted 3.0 is converted to an integer, so it will match any minor version, as in 3.0.3, 3.1.0, etc.  Likewise, 2.10 needs to be quoted, otherwise it will be interpreted as 2.1.
 
 {% raw %}
 ```yaml
 strategy:
   matrix:
-    ruby-version: [2.7.x, 2.6.x, 2.5.x]
+    ruby-version: ['3.0', 2.7, 2.6]
 ```
 {% endraw %}
 
@@ -119,7 +119,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby-version: [2.7.x, 2.6.x, 2.5.x]
+        ruby-version: ['3.0', 2.7, 2.6]
 
     steps:
       - uses: actions/checkout@v2
@@ -225,7 +225,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu, macos]
-        ruby: [2.5, 2.6, 2.7, head, debug, jruby, jruby-head, truffleruby, truffleruby-head]
+        ruby: [2.6, 2.7, '3.0', head, debug, jruby, jruby-head, truffleruby, truffleruby-head]
     continue-on-error: {% raw %}${{ endsWith(matrix.ruby, 'head') || matrix.ruby == 'debug' }}{% endraw %}
     steps:
       - uses: actions/checkout@v2


content/actions/automating-builds-and-tests/building-and-testing-ruby.md
```diff
@@ -87,13 +87,13 @@ Alternatively, you can check a `.ruby-version` file  into the root of your repos
 
 ## Testing with multiple versions of Ruby
 
-You can add a matrix strategy to run your workflow with more than one version of Ruby. For example, you can test your code against the latest patch releases of versions 2.7, 2.6, and 2.5. The 'x' is a wildcard character that matches the latest patch release available for a version.
+You can add a matrix strategy to run your workflow with more than one version of Ruby. For example, you can test your code against the latest patch releases of versions 3.0, 2.7, and 2.6.  YAML parsers will truncate trailing zeros from numeric values, and often convert floats to integers.  So, an unquoted 3.0 is converted to an integer, so it will match any minor version, as in 3.0.3, 3.1.0, etc.  Likewise, 2.10 needs to be quoted, otherwise it will be interpreted as 2.1.
```
dentarg, line 90:
```suggestion
You can add a matrix strategy to run your workflow with more than one version of Ruby. For example, you can test your code against the latest patch releases of versions 3.0, 2.7, and 2.6.  The YAML parser used by GitHub Actions will truncate trailing zeros from numeric values, and often convert floats to integers.  So, an unquoted 3.0 is converted to an integer, so it will match any minor version, as in 3.0.3, 3.1.0, etc.  Likewise, 2.10 needs to be quoted, otherwise it will be interpreted as 2.1.
```

It is good that we document this gotcha, but it sounds like it is a bug that should be addressed: https://github.com/actions/runner/issues/849#issuecomment-984495423

ramyaparimi:
@dentarg 

content/actions/automating-builds-and-tests/building-and-testing-ruby.md
MSP-Greg: IDK.  The text I wrote was meant to not be specific, as most coders would prefer [3.0, 2.7, 2.6] to ['3.0', '2.7', '2.6'].  Each item is really an array match to a version string (split into an array), which isn't a number (ie, major.minor.patch, or 3.0.1, worse with alphas, betas, rc's, etc).  Hence, I wanted to make readers aware of it without implying a yaml parser is working incorrectly.  Not the best explanation, but...

**off-topic:** For people who don't code Ruby, it's a non-typed language, but 1/3 = 0 and 1/3.0 = 0.333...
Hence, Ruby considers 3 to be an int, and 3.0 to be a float.  So, Ruby's default yaml parser considers `3` to be an int, and `3.0` to be a float.

eregon (commented):
Looks good to me :+1:

Mdbevabepari:
> ### Why:
> The wildcard doesn't work in setup-ruby: [ruby/setup-ruby#251](https://github.com/ruby/setup-ruby/issues/251)
> 
> ### What's being changed:
> Docs on how to build and test with Ruby on GitHub Actions
> 
> ### Check off the following:
> * [ ]  I have reviewed my changes in staging (look for "Automatically generated comment" and click **Modified** to view your latest changes).
> * [ ]  For content changes, I have completed the [self-review checklist](https://github.com/github/docs/blob/main/contributing/self-review.md#self-review).
> 
> ### Writer impact (This section is for GitHub staff members only):
> * [ ]  This pull request impacts the contribution experience
>   
>   * [ ]  I have added the 'writer impact' label
>   * [ ]  I have added a description and/or a video demo of the changes below (e.g. a "before and after video")
> 
> Credits to @MSP-Greg for these changes!

Bangladesh

U8Emerald (commented):
Nice 😎

skedwards88 (changes requested):
Thank you for this improvement! I think it would be best to enclose all of the version numbers in quotes. That will help users who just copy-paste-modify the example without fully understanding the nuance. Then I think the explanation about why can be removed (`YAML parsers will truncate...`).

 (The [linked bug](https://github.com/actions/runner/issues/849) seems to concern treating the float `3.0` as the int `3`, but even if that was fixed, it wouldn't address `2.10` being treated as `2.1`.)

Please let us know if you are willing to make this change!

skedwards88:
I also opened https://github.com/github/docs/issues/14764 to address this more generally

dentarg:
>Please let us know if you are willing to make this change!

While I opened the PR I can't make those changes as the branch is owned by @MSP-Greg , but I'm sure he can make em'

eregon:
Note that 2.10 will never exist as a Ruby version, and it's extremely unlikely there would be any `X.10.Z` version for Ruby.
So Float versions should just work for Ruby, and that's why I decided on this approach early on (it's also what TravisCI supported FWIW).
Unfortunately, the GitHub Actions YAML parser has this known bug https://github.com/actions/runner/issues/849 and there seems to be little effort to fix it.

dentarg:
@MSP-Greg are you able to do the requested change (https://github.com/github/docs/pull/13418#pullrequestreview-866701350)? If not, I can probably open up a new PR from my fork

MSP-Greg:
@dentarg @skedwards88

I wrote this specifically directed towards Ruby users, and Ruby is often people's first coding language (bootcamps, Rails, etc).  Beginners are also unfamiliar with CI systems.  I didn't want to imply anything was wrong, just point out differences.

So, I'm happy with the PR.

dentarg:
@skedwards88 I made the requested changes in https://github.com/github/docs/pull/15240, please review that one instead

content/actions/automating-builds-and-tests/building-and-testing-ruby.md
ethomson: Coders may prefer `[3.0, 2.7, 2.6]` for simplicity, but those aren't numbers, they're strings and they should be quoted.  Otherwise you will run into exactly the problem that you describe where `2.10` needs to be quoted to disambiguate from `2.1`.  I would illustrate the happy path in the doc and then explain _why_ you've quoted _all_ the version numbers instead of trying to tell people when they need to quote.

MSP-Greg: @ethomson

Thanks.  I don't know what's best in this doc, but there are a lot of workflows in the wild that only quote what is needed, ie '3.0'.  So, I wanted some background on why...  Also, see #15240

skedwards88:
Closing in favor of https://github.com/github/docs/pull/15240

content/actions/automating-builds-and-tests/building-and-testing-ruby.md
ethomson: That makes sense @MSP-Greg.  I think that it's safe to say that mine is an unpopular opinion, so I'd like to give some time to give others a chance to provide feedback, but I think that we should encourage users to quote version numbers and ensure that the starter workflows follow our guidance here.

skedwards88: > Coders may prefer [3.0, 2.7, 2.6] for simplicity, but those aren't numbers, they're strings and they should be quoted. 

I agree with this. The semantic version isn't actually a number, it just works in some cases to treat it as such. Although people often leave out the quotes, we try to craft examples in the docs that illustrate best practices and that will work if users copy-paste-edit without being aware of nuances like this.

@dentarg opened a PR that does this: https://github.com/github/docs/pull/15240

PR #13616, opened 2022-01-01
Correct "return" in expressions

### LINKED ISSUE #13615: Expressions: issues with "returns"
### Code of Conduct

 https://docs.github.com/en/actions/learn-github-actions/expressions 

sometime, there are issues with "returns"

In particular, for `format`, "returns" is given before the expression, which is strange to read. In all other examples, the returned value is given after the expression.
Also: 
`contains(github.event.issue.labels.*.name, 'bug')` does not indicate which value is returned. Since `github.event.issue.labels.*.name` is not obviously an array, and it's content is not clear unless you already know what is the value of this property. So this is not really a clear exemple. I believe using an explicit array here would be far better.

### Additional information

_No response_

Arthur-Milchior:
### Why:

Closes #13615

### Check off the following:


ramyaparimi:
@Arthur-Milchior 

skedwards88 (approved):
Thanks for this PR! We'll get this merged down for you.

PR #13617, opened 2022-01-01
Removed $ sign in the commands

earik87:
### Why:

`$` sign in front of the `token` should be removed to avoid confusion. It is common that first-time users will spend time to realize `$` sign is not necessary. (at least for the Windows OS).

### DIFF
--- a/content/rest/guides/getting-started-with-the-rest-api.md
+++ b/content/rest/guides/getting-started-with-the-rest-api.md
@@ -137,10 +137,10 @@ $ curl -i -u <em>your_username</em> {% data variables.product.api_url_pre %}/use
 
 When prompted, you can enter your OAuth token, but we recommend you set up a variable for it:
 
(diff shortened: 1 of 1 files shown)


ramyaparimi:
@earik87 

J-esusleal (commented):
@money

mattpollard (commented):
👋🏻 Thanks for the suggestion, @earik87!

The advice here applies anywhere that you're using the curl command in a Unix-style shell to send API requests to GitHub, including Git Bash in Windows. The paragraph specifies that `$token` is a shell variable in the context of a curl command, so the `$` is necessary. The paragraph also recommends that you do not enter your token in a way that may leave a full token in plain text in your shell history.

If we remove the `$`, the example is no longer accurate—and might also result in the reader using a plaintext token in a shell command, which is a security risk. Because of this, I'm going to close this pull request 💛

Thanks so much for taking the time to contribute to GitHub Docs! 💖

PR #14040, opened 2022-01-13
Use of hyphen in creating task lists 

aviral14112001:
I tried following the instructions line by line provided in the docs file but was unable to create a task list. After that, I tried a bunch of things and came up with the idea that it Is necessary for us to use a hyphen first and then the regular space bar between `[  ]` and hyphen to create a task list.


ramyaparimi (approved):
@aviral14112001 
Thanks so much for opening a PR with this fix! I'll get this merged once tests are passing 🍏

PR #14663, opened 2022-01-25
Prevent settings.xml generation in gradle publishing instructions

### LINKED ISSUE #14662: settings.xml auto-generation can cause issues when publishing via gradle
### Code of Conduct

https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-gradle

The first sample yaml results in the creation of a settings.xml file that can mess with gradle build commands.

Per https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven (and https://github.com/actions/setup-java/issues/510), `setup-java` automatically creates a settings.xml file like so:

``` xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <servers>
    <server>
      <id>github</id>
      <username>${env.GITHUB_ACTOR}</username>
      <password>${env.GITHUB_TOKEN}</password>
    </server>
  </servers>
</settings>
```

If this file is parsed by anything in Gradle (such as a plugin) and the developer doesn't pass along the listed environment variables, this can cause the build to fail. If you plan on doing, say, `./gradlew build` in one step and `./gradlew publish` in another, this can understandably cause issues.

### Additional information

_No response_

MT-Jacobs:
Per https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven (and https://github.com/actions/setup-java/issues/51), `setup-java` automatically creates a settings.xml file like so:

``` xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <servers>
    <server>
      <id>github</id>
      <username>${env.GITHUB_ACTOR}</username>
      <password>${env.GITHUB_TOKEN}</password>
    </server>
  </servers>
</settings>
```

This forces the developer to pass along `GITHUB_ACTOR` and `GITHUB_TOKEN` environment variables every time a gradle command is used or else the xml parsing will fail, claiming that it was `Unable to decrypt local Maven settings credentials`. If you plan on doing, say, `./gradlew build` in one step and `./gradlew publish` in another, this can understandably cause issues.

### Why:

Closes https://github.com/github/docs/issues/14662

### What's being changed:

I added an explicit `overwrite-settings: false` flag to all of the `setup-java` actions.

I also added this doc snippet:

> Note that the `overwrite-settings` option is set to false to prevent the `setup-java` action from automatically generating an unnecessary `~/.m2/settings.xml` file that require specific environment variables.

### Check off the following:

Note - I can't check off the first item because I couldn't find staging. I checked https://github.com/github/docs/actions?query=automatically+generated+content

### DIFF
--- a/content/actions/publishing-packages/publishing-java-packages-with-gradle.md
+++ b/content/actions/publishing-packages/publishing-java-packages-with-gradle.md
@@ -114,7 +114,7 @@ jobs:
 
 ## Publishing packages to {% data variables.product.prodname_registry %}
 
-Each time you create a new release, you can trigger a workflow to publish your package. / in the example below runs when the `release` event triggers with type `created`. The workflow publishes the package to {% data variables.product.prodname_registry %} if CI tests pass. For more information on the `release` event, see "[Events that trigger workflows](/actions/reference/events-that-trigger-workflows#release)."
+Each time you create a new release, you can trigger a workflow to publish your package. The workflow in the example below runs when the `release` event triggers with type `created`. The workflow publishes the package to {% data variables.product.prodname_registry %} if CI tests pass. For more information on the `release` event, see "[Events that trigger workflows](/actions/reference/events-that-trigger-workflows#release)."
 
 You can define a new Maven repository in the publishing block of your _build.gradle_ that points to {% data variables.product.prodname_registry %}.  In that repository configuration, you can also take advantage of environment variables set in your CI workflow run.  You can use the `GITHUB_ACTOR` environment variable as a username, and you can set the `GITHUB_TOKEN` environment variable with your `GITHUB_TOKEN` secret.
 
--- a/content/actions/publishing-packages/publishing-java-packages-with-gradle.md
+++ b/content/actions/publishing-packages/publishing-java-packages-with-gradle.md
@@ -110,7 +110,7 @@ jobs:
 
    For more information about using secrets in your workflow, see "[Creating and using encrypted secrets](/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)."
    
-Note that the `overwrite-settings` option is set to false to prevent the `setup-java` action from automatically generating an unnecessary `~/.m2/settings.xml` file that require specific environment variables.
+Setting `overwrite-settings` to false prevents the `actions/setup-java` action from automatically generating or overwriting an existing `settings.xml` file.
 
 ## Publishing packages to {% data variables.product.prodname_registry %}
 
--- a/content/actions/publishing-packages/publishing-java-packages-with-gradle.md
+++ b/content/actions/publishing-packages/publishing-java-packages-with-gradle.md
@@ -108,9 +108,9 @@ jobs:
 {% data reusables.github-actions.gradle-workflow-steps %}
 1. Runs the [`gradle/gradle-build-action`](https://github.com/gradle/gradle-build-action) action with the `publish` argument to publish to the `OSSRH` Maven repository. The `MAVEN_USERNAME` environment variable will be set with the contents of your `OSSRH_USERNAME` secret, and the `MAVEN_PASSWORD` environment variable will be set with the contents of your `OSSRH_TOKEN` secret.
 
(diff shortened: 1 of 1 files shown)


MT-Jacobs:
Minor fix to staging.

P.S. - one other option would be to somehow enable support for using settings.xml within gradle.

Currently it's not supported though - https://github.com/gradle/gradle/issues/1236 - at minimum the instructions would need to be updated to use a gradle plugin that automates providing the settings.xml information to gradle build scripts. These plugins exist - take https://github.com/mark-vieira/gradle-maven-settings-plugin, for instance, but I didn't find one that was well-maintained.

MT-Jacobs:
One final piece of context:

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-gradle-registry

This doc provides an _entirely different_ strategy for providing the auth user and password:

``` kotlin
plugins {
    id("maven-publish") apply false
}
subprojects {
    apply plugin: "maven-publish"
    publishing {
        repositories {
            maven {
                name = "GitHubPackages"
                url = uri("https://maven.pkg.github.com/OWNER/REPOSITORY")
                credentials {
                    username = project.findProperty("gpr.user") ?: System.getenv("USERNAME")
                    password = project.findProperty("gpr.key") ?: System.getenv("TOKEN")
                }
            }
        }
        publications {
            gpr(MavenPublication) {
                from(components.java)
            }
        }
    }
}
```

`gpr.user` and `gpr.key` are undocumented, from what I can tell. It's possible that I'm missing something there.

ramyaparimi:
@MT-Jacobs 

bwherever:
Hello

content/actions/publishing-packages/publishing-java-packages-with-gradle.md
```diff
@@ -108,6 +109,8 @@ jobs:
 1. Runs the [`gradle/gradle-build-action`](https://github.com/gradle/gradle-build-action) action with the `publish` argument to publish to the `OSSRH` Maven repository. The `MAVEN_USERNAME` environment variable will be set with the contents of your `OSSRH_USERNAME` secret, and the `MAVEN_PASSWORD` environment variable will be set with the contents of your `OSSRH_TOKEN` secret.
 
    For more information about using secrets in your workflow, see "[Creating and using encrypted secrets](/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)."
+   
+Note that the `overwrite-settings` option is set to false to prevent the `setup-java` action from automatically generating an unnecessary `~/.m2/settings.xml` file that require specific environment variables.
```
skedwards88, line 113:
```suggestion
Setting `overwrite-settings` to false prevents the `actions/setup-java` action from automatically generating or overwriting an existing `settings.xml` file.
```

skedwards88 (commented):
Thanks for opening this PR! I am not familiar with Java--is it ok to not have a settings.xml file, or are there cases where you will hit an error without this file?

MT-Jacobs:
> Thanks for opening this PR! I am not familiar with Java--is it ok to not have a settings.xml file, or are there cases where you will hit an error without this file?

Correct, the `settings.xml` file is only needed for Maven as documented [here](https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven).

However, as documented in https://github.com/github/docs/issues/14662#issuecomment-1026214817, I misidentified the root cause of my problem. We _could_ skip generating the `settings.xml` as it's unused, but I no longer think it's worth updating the documentation unless I can find a better way to use Maven repos in Gradle in a GitHub workflow.

Thank y'all very much for taking an initial look at my suggestion.

PR #14761, opened 2022-01-28
unbreak repo-sync by fixing the staging deploy workflow

peterbe:
We need https://github.com/github/docs-internal/pull/24786 into `github/docs` but it won't get there (from `github/docs-internal`) because the next repo-sync (which will have it in the diff) won't actually be using it because the workflow depends on being in `main` before the PR starts.


heiskr (approved):

INzoGravity (commented):
My friend

PR #15613, opened 2022-02-18
Add warnings about env: field

FlashSheridan:
### Why:

Relying on the documentation “workflow-syntax-for-github-actions” for complex actions which use `env:` variables is likely to fail.

### What's being changed:
Add two warnings from painful experience:
In the section on `env:`
-   Note that variables in the map cannot be defined in terms of other variables in the map.  If you need complex manipulation of variables, you may wish to use a non-default shell (see below), but non-default shells do not use the `env` map.
In the section on non-default shells
-   Note that non-default shells do not use the `env` map.


FlashSheridan:
Part of this is wrong; I misremembered Python dict syntax.  I’ll create a new PR with the correct bit.

PR #15622, opened 2022-02-18
Add warning about env vars defined via others

FlashSheridan:
### Why:
Variables in the `env` map cannot be defined in terms of other variables in the map, which may surprise those used to regular shell environment variables.

### What's being changed:
Add note to documentation.


ramyaparimi:
@FlashSheridan 

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -184,7 +184,7 @@ A boolean specifying whether the secret must be supplied.
 
 ## `env`
 
-A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps[*].env`](#jobsjob_idstepsenv).
+A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps[*].env`](#jobsjob_idstepsenv). Note that variables in the `env` map cannot be defined in terms of other variables in the map.  If you need complex manipulation of variables, you may wish to use a non-default shell (see below).
```
ethomson, line 187: 🤔  I don't feel like I understand entirely how changing the shell helps here yet - I don't think that we have documentation below about complex variable interpolation.  I think that the "see below" is about how to configure non-default shells, but it feels like a bit of a cliffhanger to tell people that they could do complex manipulation but not follow up on that below.

```suggestion
A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps[*].env`](#jobsjob_idstepsenv). Note that variables in the `env` map cannot be defined in terms of other variables in the map.
```

ethomson:
This makes sense.  Thanks!

content/actions/using-workflows/workflow-syntax-for-github-actions.md
FlashSheridan: It’s subjective (hence “may wish“), but I found dealing with these variable much easier after switching to Python commands.  (Anyone who is actually confident programming Bash is free to disagree.)

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -184,7 +184,7 @@ A boolean specifying whether the secret must be supplied.
 
 ## `env`
 
-A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps[*].env`](#jobsjob_idstepsenv).
+A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps[*].env`](#jobsjob_idstepsenv). Note that variables in the `env` map cannot be defined in terms of other variables in the map.
```
lecoursen, line 187: Let's move this to its own line, so that the link is the last part of this line. 😊 
```suggestion
A `map` of environment variables that are available to the steps of all jobs in the workflow. You can also set environment variables that are only available to the steps of a single job or to a single step. For more information, see [`jobs.<job_id>.env`](#jobsjob_idenv) and [`jobs.<job_id>.steps[*].env`](#jobsjob_idstepsenv). 

Variables in the `env` map cannot be defined in terms of other variables in the map.
```

lecoursen (dismissed):

lecoursen (approved):

PR #15839, opened 2022-02-25
Github Actions Release Trigger | docs wrongly state that release-Action is triggered on draft

kohlerdominik:
### Why:

My tests lead me to the assumption, that the `released`-action is only triggered when a release is done (either directly or by finishing a draft).

The docs say, `release` is also triggered when a release-draft is published. But IMO this isn't the case. And it wouldn't make sense with an action called `release`, as a draft is NOT a release.

### What's being changed:

The docs state, that a Webhook with the `action=released` is triggered, when a draft is published.

> `released`: a release or draft of a release is published, or a pre-release is changed to a release

But IMO this is neither the case, nor would it make sense regarding the naming. A draft is NOT a release.

### Check off the following:


ramyaparimi:
@kohlerdominik 

Danielferencz21:
Kodex

davidmerwin (commented):
❤️

davidmerwin:
❤️

lucascosti (approved):
Thanks for this fix, @kohlerdominik.

PR #16021, opened 2022-03-03
[actions] remove redundant note from `push` event

kaznovac:
### Why:

Closes n/a

### What's being changed:

there are two notes about combining the `push` event filters;
both the same and they both fit on the same screen (2k, 15");
removed the first one (so both note mentioned filter properties get introduced by doc)

### Check off the following:

### DIFF
--- a/content/actions/using-workflows/events-that-trigger-workflows.md
+++ b/content/actions/using-workflows/events-that-trigger-workflows.md
@@ -948,21 +948,6 @@ on:
       - 'releases/**'
 ```
 
-{% note %}
-
-**Note:** {% data reusables.actions.branch-paths-filter %} For example, the following workflow will only run when a push that includes a change to a JavaScript (`.js`) file is made to a branch whose name starts with `releases/`:
-
-```yaml
-on:
-  push:
-    branches:    
-      - 'releases/**'
-    paths:
-      - '**.js'
-```
-
-{% endnote %}
-
 #### Running your workflow only when a push of specific tags occurs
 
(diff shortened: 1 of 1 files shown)


ramyaparimi:
@kaznovac 

lucascosti:
Thanks for this PR, @kaznovac. Because the note is included in subsections that are specific to different situations (workflows when a push to specific branches occurs, vs specific files), users might only read the section relevant to their situation. We need to have the note in both places, because often people won't read the whole `push` section.

kaznovac:
@lucascosti thanks for the review...
than by your reasoning (at least) one note is missing for the middle section https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

lucascosti:
@kaznovac 🤔 I don't think so, because IIRC, the behavior in the note is specific to only the `branches` and `paths` filters (not `tags`). 🙂

kaznovac:
@lucascosti yes but this restrictions are not specific to the `branches` and `paths` only, those can be combined with `tags`. see here for example https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags

lucascosti:
@kaznovac no, the restriction does not apply to `tags` and `branches`/`paths`. Filters are normally on an `OR` basis, but if you define both `branches` and `paths`, the workflow will only run when both those specific filters are satisfied.

e.g. as per [your link](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches-and-tags), if you define `branches` and `tags` filters, the workflow will run when either filter is satisfied.

kaznovac:
@lucascosti a bit confusing but makes sense (as both tags and branches are got references), thanks for clarification TIL.

PR #16167, opened 2022-03-10
Python versions in github action YAML examples should be type string, not float

### LINKED ISSUE #14764: Quote language versions in YAML examples
### Code of Conduct

YAML examples in https://docs.github.com/actions

YAML treats floats like `2.10` as `2.1`. This can lead to workflows using a different language version than intended. We should update the YAML examples to quote language versions to avoid this. To help users who copy-paste-modify without understanding the nuance, we should quote all language versions, even ones that would not be affected by this parsing.

For example:

```
jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: [3.6, 3.7, 3.8, 3.9]
```

should be changed to 

```
jobs:
  build:

    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: ["3.6", "3.7", "3.8", "3.9"]
```

Then if a user copies the example but wants to include version 3.10, they can easily plug in the new value without knowing they need to add quotes.

### Additional information

_No response_

## Actions articles with examples that need quotes

- [ ] [Building and testing Node.js](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs)
- [ ] [Building and testing Ruby](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-ruby)
- [ ] [Complete example for GitHub Actions](https://docs.github.com/en/actions/migrating-to-github-actions/manually-migrating-to-github-actions/migrating-from-circleci-to-github-actions#complete-example-for-github-actions)
- [ ] [GitHub Actions syntax for a matrix](https://docs.github.com/en/actions/migrating-to-github-actions/manually-migrating-to-github-actions/migrating-from-travis-ci-to-github-actions#github-actions-syntax-for-a-matrix)
- [ ] [Example: Using contexts to create matrices](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-contexts-to-create-matrices)

newmanrs:
### Why:

Github action YAML examples on the [python/js hybrid doc page](https://docs.github.com/actions/automating-builds-and-tests/building-and-testing-nodejs-or-python?langId=py) have versions using floats, not string.  This potentially creates an issue where users editing the examples to python version 3.10 will error after it is parsed as 3.1 (float) and later cast into "3.1" (string).  
Fixes #14764, at least for python examples.

### Check off the following:


ramyaparimi:
@newmanrs 

peterbe (approved):
I agree. 
It actually works with numbers too, surprisingly. 
Example: https://github.com/peterbe/django-fancy-cache/blob/e3cb62ff1474fb2fa839993ec92cc6436839b0c8/.github/workflows/ci.yml#L14
There, `3.10` had to become `"3.10"` so it wasn't YAML parsed as `3.1` but I much prefer to use quotation marks to avoid the risk in the first place. 

👍

PR #16275, opened 2022-03-15
Added explanation about masking outputs

### LINKED ISSUE #16274: Masked outputs are not passed and are considered as secrets
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#masking-a-value-in-log

### Additional information

There were multiple issues in the actions/runner repository ([1061](https://github.com/actions/runner/issues/1061),[1498](https://github.com/actions/runner/issues/1498)) and the community [discussion](https://github.community/t/skip-output-aws-account-id-since-it-may-contain-secret/123791) related to this update.

nikola-jokic:
### Why:

Closes #16274 

### What's being changed:
Proposed an explanation that you should not use add-mask with outputs.

### Check off the following:


ramyaparimi:
@nikola-jokic 

content/actions/using-workflows/workflow-commands-for-github-actions.md
```diff
@@ -190,7 +190,7 @@ echo "::endgroup::"
 ::add-mask::{value}
 ```
 
-Masking a value prevents a string or variable from being printed in the log. Each masked word separated by whitespace is replaced with the `*` character. You can use an environment variable or string for the mask's `value`.
+Masking a value prevents a string or variable from being printed in the log. Each masked word separated by whitespace is replaced with the `*` character. You can use an environment variable or string for the mask's `value`. Masked outputs are considered as secrets, and will be redacted on the runner. See [Workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions).
```
lucascosti, line 193: Thanks for this clarification, @nikola-jokic. I think we can add a bit more explanation as to what being 'considered as secrets' means. Is this rewording accurate?

```suggestion
Masking a value prevents a string or variable from being printed in the log. Each masked word separated by whitespace is replaced with the `*` character. You can use an environment variable or string for the mask's `value`. When you mask a value, it is treated as a secret and will be redacted on the runner. For example, after you mask a value, you won't be able to set that value as an output.
```

nikola-jokic: Hi @lucascosti, you are right. I am committing your changes. Thank you!

lucascosti (approved):
Awesome! Marking this as ready to merge when the current deployment freeze is finished.

Hmnibsheer:
Hot

PR #16350, opened 2022-03-18
chore: replace deprecated String.prototype.substr()

CommanderRoot:
### Why:

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### What's being changed:

On one line substr(1) is being changed to slice(1). Both return the same values.

### Check off the following:


ramyaparimi:
@CommanderRoot 

rsese (approved):
thank you 👍 !

PR #16456, opened 2022-03-22
Fix broken title

nhedger:
### Why:

Fix a broken title.

I suspect a newline between `{% endpowershell %}` and the title is needed, otherwise, the title does not render properly.

Screenshot 2022-03-22 at 13 02 28

### What's being changed:

Adds a newline before the `## Grouping log lines` title.


nhedger:
I've checked the modification on the preview environment and the change restores the title as expected.

ramyaparimi (approved):
@nhedger 
Thanks so much for opening a PR with this fix! I'll get this merged once tests are passing 🍏

PR #16506, opened 2022-03-24
Add missing cli parameter to curl command

gijswijs:
To see the returned HTTP headers of an API request using your app's client ID and secret,
you need to add the `-I` parameter.


ramyaparimi:
@gijswijs The PR has a lot more changes than adding `-I` to the curl command. Please revert back all the other changes and I will triage this for a review 💖 .

gijswijs:
@ramyaparimi I see my Prettier formatter had a field day with that file. I forgot to turn it off.
I've reverted the commit, and added a new one with only that single change.  I can make a different PR if that's easier. Just let me know.

PR #16533, opened 2022-03-25
npm publication

Abhi-lab2:
npm i random-number-generator-masai


Abhi-lab2:
use npm i random-number-generator-masai to use the package

PR #16600, opened 2022-03-28
Actions self-hosted runners - Fix incorrect installation instruction

rbuchberger:
### Why:

The instructions for installing a self-hosted actions runner on linux as another user are incorrect, and if followed fail with an error. This change corrects those instructions.

[svc.sh](https://github.com/actions/runner/blob/408d6c579c36f0eb318acfdafdcbafc872696501/src/Misc/layoutbin/systemd.svc.sh.template#L65) simply takes the [second argument](https://github.com/actions/runner/blob/408d6c579c36f0eb318acfdafdcbafc872696501/src/Misc/layoutbin/systemd.svc.sh.template#L8); there's no `--user` flag. If provided, the script tries to find a user literally named `--user` and fails with "Group not available".

There's no associated issue.

### What's being changed:

Remove `--user` flag from `./svc.sh install <USER>` instructions, which results in desired behavior.

### Check off the following:

### DIFF
--- a/content/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service.md
+++ b/content/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service.md
@@ -78,10 +78,10 @@ You can manage the runner service in the Windows **Services** application, or yo
    ```
 {% endmac %}
 
-The command takes an optional `user` argument to install the service as a different user.
+The command takes an optional second argument to install the service as a different user.
 
 ```shell
-./svc.sh install --user <em>USERNAME</em>
+./svc.sh install <em>USERNAME</em>
 ```
 
 ## Starting the service


ethomson:
Thanks for this! It looks like the `svc.sh` only accepts a user argument when adding a systemd service - it will silently fail on macOS and, worse, there's no `svc.sh` on Windows at all.

I think that your change is :100: but it would be great if you could pull it up into the linux conditional (so, line https://github.com/github/docs/pull/16600/files#diff-9880f13992e379bdd4479c01ab92296f51c3923c521fcab04d56c85f96481adfL67 if I'm looking at the file correctly.)

ramyaparimi:
@rbuchberger I am checking to see if you are still working on this PR 💖 .

rbuchberger:
Hi @ramyaparimi - sorry for the long delay! I was on vacation and only returned today. I'll have a look.

dreamingskey:
What good are words if they dont have definitions as integrity of moral relation the as a group constructed this in the key I was givin and it stops n

Deloww (commented):
> content/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service.md`[content/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service.md](content/actions/hosting-your-own-runners/configuring-the-self-hosted-runner-application-as-a-service.md) `

PR #16715, opened 2022-03-31
docs(graphql): order token scopes by form order

Bamblehorse:
### Why:

The order of the scopes on the [Forming calls with GraphQL page](https://docs.github.com/en/graphql/guides/forming-calls-with-graphql) is not the same as the order of the form to [create a personal access token](https://github.com/settings/tokens/new).

It might also be useful to remove `repo` from the list - as it implies checking all the nested repo boxes.

A second useful clarification here might be documenting the parent scope as well as the child scope - to allow for easy scanning.

## Scope order on [Forming calls with GraphQL page](https://docs.github.com/en/graphql/guides/forming-calls-with-graphql)
image

https://docs.github.com/en/graphql/guides/forming-calls-with-graphql

## Scope order on [create a personal access token page](https://github.com/settings/tokens/new)
github com_settings_tokens_new

### Check off the following:


skedwards88 (approved):
This seems reasonable to me! We'll get this merged down for you.

PR #16717, opened 2022-03-31
Added information on continuing to run workflows

HammyHavoc:
### Why:

I received an email from GitHub warning me that my workflow would be disabled as there hasn't been activity. I clicked through, and it takes me to https://github.com/HammyHavoc/hammyhavoc/actions?query=workflow:%22Latest+blog+post+workflow%22 instead of https://github.com/HammyHavoc/hammyhavoc/actions/workflows/blog-post-workflow.yml?query=workflow%3A%22Latest+blog+post+workflow%22, thus the `Continue running workflow` option doesn't appear. I also clicked `Learn More` in the email, and the documentation didn't explain this.

### What's being changed:

I've added the information to prevent others from wasting 15m+ trying to figure this out. This isn't the first time this has happened to me.

### Check off the following:


ramyaparimi:
@HammyHavoc 

data/reusables/actions/scheduled-workflows-disabled.md
```diff
@@ -1 +1 @@
-To prevent unnecessary workflow runs, scheduled workflows may be disabled automatically. When a public repository is forked, scheduled workflows are disabled by default. In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days.
+To prevent unnecessary workflow runs, scheduled workflows may be disabled automatically. When a public repository is forked, scheduled workflows are disabled by default. In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days. You can continue to run scheduled workflows without repository activity by visiting the workflow's page and clicking `Continue running workflow`.
```
ethomson, line 1: I'm generally +1 on documenting this, it seems useful. I'm not sure that it should be in the reusable, though, it seems like it should only be a part of the disabling-and-enabling-a-workflow page. It would fit in nicely there and breaking it out of the reusable allows us to expand the instructions so that it's a little more verbose and step-by-step than just "[visit] the workflow's page".

I wonder if the docs team has any thoughts here...

skedwards88: I agree with @ethomson . This reusable is also used in the "Usage limits, billing, and administration" article, but the added info is really only relevant to the ["Disabling and enabling a workflow" article](https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow#enabling-a-workflow). Adding this info to the "Disabling and enabling a workflow" article (`content/actions/managing-workflow-runs/disabling-and-enabling-a-workflow.md`) instead makes sense. 

(To help orient you in that file: this reusable is referred to as`{% data reusables.actions.scheduled-workflows-disabled %}`. You can find more info about reusables [here](https://github.com/github/docs/blob/main/data/reusables/README.md).)

I also agree that we could give users a bit more context. Since this is almost identical to enabling a workflow, maybe we can can say something like:
```
To continue to run a scheduled workflow that was disabled due to no repository activity, follow the steps under [Enabling a workflow](#enabling-a-workflow) and click `Continue running workflow`.
```

Alternatively, we could add a new section that is similar to the [`Enabling a workflow` section](https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow#enabling-a-workflow)

janiceilene:
👋 @HammyHavoc I'm going to close this up since there's been no activity for a while, but please feel free to reopen if you'd like to address the review comments in https://github.com/github/docs/pull/16717#discussion_r844247282 💖

PR #16793, opened 2022-04-03
fix: code examples search component edge case

### LINKED ISSUE #16313: Code example search isn't escaping the regex
### Steps to reproduce:

1. Go to <http://localhost:4000/en/actions#code-examples>
2. Type in `*` (one star) and hit Search

### Expected result:

Not an error. 

### Actual result

```
CodeExamples.tsx?27e8:25 Uncaught SyntaxError: Invalid regular expression: /*/: Nothing to repeat
```

---

extra notes ✏️ 

* It's any character into the search that can cause an invalid regular expression. E.g. `(` or `+`.
* Offending code: https://github.com/github/docs/blob/f36f0995fef072cd4085af0c33ced9478cd6371d/components/landing/CodeExamples.tsx#L25

riteshsp2000:
### Why:

Closes #16313 

### What's being changed:
- The `docs/components/landing/CodeExamples.tsx` component search includes a step where regex is created of the term to be searched.
- Earlier, special characters like `*` or `(` weren't escaped which lead to the entire site crashing.
- Now, before the regex creation, all special characters are escaped by replacing the string containing a special character with `/{special_character}`

### Check off the following:


ramyaparimi:
@riteshsp2000 

components/landing/CodeExamples.tsx
```diff
@@ -22,7 +22,8 @@ export const CodeExamples = () => {
   const isSearching = !!search
   let searchResults: typeof productCodeExamples = []
   if (isSearching) {
-    const matchReg = new RegExp(search, 'i')
+    // The following replace method escapes special characters in regular expression creation.
+    const matchReg = new RegExp(search.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&'), 'i')
```
peterbe, line 26: Would it work to just replace anything that isn't `\w`?

heiskr: I'm wondering what the disadvantage of using `includes` instead of `test` below would be, instead of needing to evaluate RegExp. I would also default to an allowlist instead of a denylist.

rsese: Thanks so much for picking this up @riteshsp2000 ⚡ !  fwiw I tested your PR manually and it resolves the issue that where you would get an error when searching for `*` for example -- but I'll defer to peter and kevin on the discussion above.

heiskr (approved):

PR #17089, opened 2022-04-14
Specify return value for undefined properties in workflow contexts

rossjrw:
### Why:

The documentation for workflows doesn't mention what happens when you access a nonexistent property. This is important to know - maybe it throws an error? It returns an empty string, so mentioning it will save someone an experiment to find out.

### What's being changed:

Add a line to https://docs.github.com/en/actions/learn-github-actions/contexts.

Could possibly be more appropriate for https://docs.github.com/en/actions/learn-github-actions/expressions; however, I don't think there's a way to access a property in an expression without going via a context, so this seems fine.

### Check off the following:


ramyaparimi:
@rossjrw 

content/actions/learn-github-actions/contexts.md
```diff
@@ -57,6 +57,8 @@ In order to use property dereference syntax, the property name must:
 - start with `a-Z` or `_`.
 - be followed by `a-Z` `0-9` `-` or `_`.
 
+Attempting to access a non-existent property on an existing context will return an empty string.
```
ethomson, line 60: Minor suggestion to align with the existing language.

```suggestion
Attempting to dereference a non-existent property will evaluate to an empty string.
```

ethomson:
Thanks @rossjrw - I added a minor suggestion but this is a helpful improvement! 🎉

content/actions/learn-github-actions/contexts.md
```diff
@@ -57,6 +57,8 @@ In order to use property dereference syntax, the property name must:
 - start with `a-Z` or `_`.
 - be followed by `a-Z` `0-9` `-` or `_`.
 
+Attempting to dereference a non-existent property will evaluate to an empty string.
```
felicitymay, line 60: To bring this into line with our style guide, I'm going to suggest a rephrasing too.

```suggestion
If you attempt to dereference a non-existent property, it will evaluate to an empty string.
```

felicitymay (commented):
@rossjrw - thanks for this improvement to the page 💖

felicitymay (approved):

PR #17199, opened 2022-04-19
Color change labels

### LINKED ISSUE #15989: Landing pages: Update the code examples area to differentiate between different types of content
**This issue is reserved for MLH fellow program**

The code examples area was originally created for the Actions landing page and most examples there are of the same type (actions). This has been adapted for use on the Code security landing page. One key difference between actions and code security is that we need to display a variety of different content types, for example:

- Great examples of `security.md` files (security policy)
- Examples of workflows including the CodeQL action `codeal-analysis.yml` and maybe third party static analyzers (code scanning)
- Examples of security advisories
- Examples of `dependabot.yml` files (Dependabot version updates)

Some visual identifier for different types of examples, so it's clearer what people will see when they click on an item. It would be good to be able to filter by example type too - similarly to how you can in the All guide section of the Guides sub-landing page.

cmwilson21:
### Why:

Closes [15989](https://github.com/github/docs/issues/15989)

### What's being changed: 
After much digging and trial and error runs, three files have been changed. 
The code-examples.yml file has been updated so that each tag has a name and a color designated to it.
The ProductLandingContext.tsx has been changed to reflect that tags attribute is no longer an array of strings, but an array of objects that are strings.
The CodeExampleCard.tsx has been changed so that the map on line 23 returns a Label that includes the tag.color for the background style, and the tag.name for the actual name of the Label.

Screen Shot 2022-04-19 at 11 26 04 AM

Screen Shot 2022-04-19 at 11 26 18 AM

### Check off the following:


ramyaparimi:
@cmwilson21 

rsese:
Thanks so much @cmwilson21 🙇 !   Curious if you have any thoughts on this one @joeoak?  I ask because we somewhat recently removed colors from the release notes pages (e.g. https://docs.github.com/en/enterprise-server@3.4/admin/release-notes -- the headings like `Security fixes` and `Bug fixes` used to be color coded but we were advised to remove the colors by the design folks).

Also going back to the issue I see the suggestion mentions:

> Some visual identifier for different types of examples, so it's clearer what people will see when they click on an item.

I guess a question is does the different colored labels make it clearer what you're getting when you click on an item?

joeoak:
Thanks for the @mention @rsese.

* Let me track down information on the recommendation to remove color coding from headers on the release notes page.
  * I suspect these are different issues and that applying colors to label components is ok (e.g. https://primer.style/react/Label).
* I think tags could be helpful indicators - but perhaps more in categorization/showing which examples are related.

rsese:
> * I think these labels are helpful indicators - but perhaps more in categorization/showing which examples are related?

hmm yeah it does seem like if showing related examples was the intent, then the colors could be useful?  But since the issue asked if there was a way to make it clear what kind of example people will get before clicking on it, I wonder if there's some other approach we could take 🤔 

> * Looks like the recommendation was to avoid applying color to text rather than discouraging the use of color coding overall.

ahh right I vaguely remember some of the discussion and I think we also talked about how the color choices can be kinda arbitrary and aren't necessarily aligned with something (e.g. why color X for this category?) and wouldn't be consistent across the docs/product?

gracepark:
Hm, so I've been working with @cmwilson21 on this, and I have couple things (all non blocking this PR):

1) I think the different color labels are nice, and I think it still somewhat aligns with the rest of .com. As seen here we seem to be using colors to differentiate between languages:
Screen Shot 2022-04-29 at 2 08 40 PM

In terms of what the colors represents, and why Go is Blue, I don't know the reason. 

2) We should also do the same to the Search label as @joeoak pointed out in a followup! I completely forgot about those. 

3) It might be a good idea to revisit the colors in a follow up as well - for example, `Security advisory` makes sense that it is red , but should be bundle the colors? Maybe everything in the category of security would be red, like `CodeQL` and `Security Policy`? Just a suggestion since there are quite a few colors, and maybe bundling could give the colors a bit more meaning beyond just distinction. 

4) In terms of showing the user what _kind_ of examples these are, I agree with @rsese that we might want to rethink this section and possibly come up with a different approach that's more obvious for users.

rsese (commented):
Thanks for the 👓 Grace!  I think those sound like great followups 👍!  It's also possible we might update the landing page design in the future as well but we'll see! 

I did notice an issue when playing with the preview app -- looks like in the case where someone doesn't specify a tag color, then the tags aren't shown at all.  E.g. on this page:

https://docs-17199-f4be0f.preview.ghdocs.com/en/codespaces

There are no tags rendering but we can see them on production:

https://docs.github.com/en/codespaces

Looks like the example file for that category wasn't updated with tag colors:

https://github.com/github/docs/blob/adb6476fac4ede4ec8774704b70d7c805f11392c/data/product-examples/codespaces/code-examples.yml

There's probably a couple of other categories that have examples that need the tags updated with colors as well?  I guess the code could also try to handle that case and specify a fallback color maybe but since the type we created says a tag should have a name and a color, maybe it makes more sense to update all the tags?

janbrasna:
@gracepark FYI the language colors come from [@github/linguist](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml)

components/context/ProductLandingContext.tsx
```diff
@@ -24,7 +24,10 @@ export type CodeExample = {
   description: string
   languages: string // single comma separated string
   href: string
-  tags: Array<string>
+  tags: Array<{
+    name: string
+    color: string
+  }>
 }
 export type Product = {
```
xo-r32, line 32: @cmwilson21

components/landing/CodeExampleCard.tsx
```diff
@@ -20,10 +20,10 @@ export const CodeExampleCard = ({ example }: Props) => {
           dangerouslySetInnerHTML={{ __html: example.description }}
         />
         <div className="d-flex flex-wrap">
-          {example.tags.map((tag) => {
+          {example.tags.map((tag, id) => {
             return (
-              <Label key={tag} variant="small" sx={{ bg: 'accent.emphasis', mb: 1, mr: 2 }}>
```
xo-r32, line 25: Gkm

components/context/ProductLandingContext.tsx
xo-r32, line 32: {}1.90

jaystyles:
@dependabot merge

jaystyles:
@dependabot close

SaberParsa (commented):
🤑

components/context/ProductLandingContext.tsx
Mralfa68: > @cmwilson21

heiskr (approved):

Hi2885 (commented):
Export type

janiceilene:
I'm going to close this PR! Priorities shifted and we moved away from this work ❤️

PR #17266, opened 2022-04-21
Add note about how to specify event in github action

xieyuschen:
### Why:

Hi, I'm new to GitHub action and I would like to push the built docker image to the docker hub when I commit. I pasted the demo yml to try it but I was blocked and spent a lot of time solving it, because our doc doesn't note that we need to specify the event that triggers the docker action. 
So I add a note in our docs and hope it could help the other newer.

### What's being changed:
Add a note for the newer to let them know there is an event that can be set to control when the event is triggered.

### Check off the following:

### DIFF
--- a/content/actions/publishing-packages/publishing-docker-images.md
+++ b/content/actions/publishing-packages/publishing-docker-images.md
@@ -104,7 +104,7 @@ jobs:
           labels: {% raw %}${{ steps.meta.outputs.labels }}{% endraw %}
 ```
 
-The above workflow checks out the {% data variables.product.prodname_dotcom %} repository, uses the `login-action` to log in to the registry, and then uses the `build-push-action` action to: build a Docker image based on your repository's `Dockerfile`; push the image to Docker Hub, and apply a tag to the image.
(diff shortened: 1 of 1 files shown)


ethomson:
🤔  We do actually mention it in the [top of the section](https://docs-17266-4ecd2a.preview.ghdocs.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-docker-hub) and link to the "[Events that trigger workflows](https://docs-17266-4ecd2a.preview.ghdocs.com/en/actions/reference/events-that-trigger-workflows#release)" page, which I think makes sense.

I don't think that it hurts to restate that, though. My suggestion would be to bring the trigger to the top of the paragraph instead of the bottom, and not as a "note".  For example:

> The above workflow runs when you create a release in your repository. The workflow checks out the GitHub repository, uses the login-action to log in to the registry, and then uses the build-push-action action to: build a Docker image based on your repository's Dockerfile; push the image to Docker Hub, and apply a tag to the image.

I would also apply this consistently to the three workflows.

PR #17345, opened 2022-04-24
Note major versions are not automatically tagged

### LINKED ISSUE #17071: Clarify That GitHub Actions Don't Automatically Understand Semantic Versioning
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses

> Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.

This section can be read to imply that GitHub Actions understands semantic versioning, but it does not. In order for users to specify a major action version, the action maintainer has to tag it themselves. It would be good to clarify that tagging a major version only works if the maintainer chooses to support it. In my personal opinion, the risks of being broken by a bad release outweigh the benefits of always using the latest version of an action (at a specified major version). I would suggest pinning a specific version and using Dependabot to keep up-to-date instead.

### Additional information

_No response_

Kurt-von-Laven:
The previous wording can be misinterpreted to suggest that GitHub Actions itself automatically tags major versions.

Also, remove assurance that pinning the major version protects your workflow from breaking. This is only true if the action you depends on never introduces any bugs that affect you. In other words, it's false in practice. The new copy seeks to highlight that the three presented options form a continuum from maximum security/stability, minimum convenience to minimum security/stability, maximum convenience.

### Why:

Closes #17071.

### What's being changed:

See [rich diff](https://github.com/github/docs/pull/17345/files?short_path=ef18f68#diff-ef18f687729e1ee0549e9b4324265cf19757637bf6a744d01213c7fcc988399f).

### Check off the following:


Kurt-von-Laven:
In case anyone is following this thread, evidently when Octomerger merges a batch of pull requests at once, GitHub incorrectly states that the author of each PR force-pushed their own branch, which, to be clear, I didn't, and automatically closes the pull request as if it had not been merged (which it has).

ramyaparimi:
Ah!! That explains why I never triaged this since it shows that the author closed the PR. I will let the team know about this. 

In the mean time, could you let me know if this PR has to be re-opened and triaged?

Ah!! So weird, I am not able to re-open the PR, probably because the branch for this PR is deleted. 
I will check with the team on this and get back to you. Thanks so much for your patience 💛 .

Kurt-von-Laven:
Yeah, it's probably worth bringing up simply because it's somewhat confusing for those who aren't familiar with the inner workings of Octomerger. There's no action needed specific to this PR, because it has already been merged by Octomerger. You are correct that PRs cannot be reopened once the branch has been deleted, although even if that weren't true, there would be nothing left on the branch (that isn't already on main) to merge at this point. Thanks for noticing my random comment on a closed PR in any case!

ramyaparimi:
Thanks so much for your valuable insights and contributions to the GitHub docs 💖 .

Kurt-von-Laven:
Hmm... it appears I was mistaken, and this change was never merged to main. This pull request replaced "Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work." with "If the action publishes major version tags, using them allows you to receive critical fixes and security patches while still maintaining compatibility." You can no longer see the original diff here on account of it having been merged, but the [orphaned commit](https://github.com/github/docs/commit/10970867b028ff60e82273797ad8323a7814c99f) remains as evidence of this. The [Octomerger force-push](https://github.com/github/docs/compare/10970867b028ff60e82273797ad8323a7814c99f..df90a49243831b0143064ba56b5af46b3227cd3a) actually includes the opposite change, which explains why my change [never appeared](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses). This does seem like a more serious bug in Octomerger than I initially realized. In summary, Octomerger automatically closed this pull request without merging it, and this action was displayed in the GitHub UI as a force-push by me that I didn't actually do.

rachmari:
@Kurt-von-Laven @ramyaparimi this certainly looks like a bug. I've opened an internal engineering issue to take a look at it. Thank you for reporting this and I apologize for the inconvenience.

PR #17535, opened 2022-05-02
Update managing-a-branch-protection-rule.md

raphaelguye:
Bypassing Pull Requested only work if you have **not** configured required status checks to pass before merging.
It should be added in the documentation.


raphaelguye:
Is there any chance to change this behaviour? Can I create an issue somewhere?

I would like that we can also bypass even if some status checks are configured.

ramyaparimi:
@raphaelguye 

stevecat:
Hi @raphaelguye, thanks for opening this! Can we ask you to first speak to support (either via the [support portal](https://support.github.com/) or [community forum](https://github.community/)) to confirm if this behaviour is intended or needs to be addressed by our engineering team?

PR #17628, opened 2022-05-04
Adding REST API option for Promoting and Demoting

enyil:
### Why:

Site administrators can also be promoted and demoted utilizing the REST API. The fact that is missing in this piece of documentation makes it seem like REST API is not an option. So I'm adding it instructions and references for REST API promote site administrator and demote.

Closes [issue link]

### What's being changed:

I'm adding it instructions and references for REST API promote site administrator and demote.

### Check off the following:


ramyaparimi:
@enyil 

lecoursen (changes requested):
Thanks @enyil, I think it's a great idea to mention the corresponding REST endpoints here!

However, we don't want to duplicate content in this article that already exists in the REST docs. Let's remove the new sections you've added and instead add one line above "Promoting a user from the enterprise settings", that says something like this:

>You can also promote and demote site administrators using the REST API. For more information, see  "[Promote a user to be a site administrator](/rest/enterprise-admin/users#promote-a-user-to-be-a-site-administrator)" and "[Demote a site administrator](rest/enterprise-admin/users#demote-a-site-administrator)."

mattpollard:
Thanks for looking out for our customers, @enyil ❤️ Agreed that this would be helpful to improve! Would the approach @lecoursen suggested solve the problem for those customers? If so, want to go ahead and implement those changes and re-request review? We're also happy to pick this up if you're otherwise occupied ☺️

PR #17681, opened 2022-05-07
Suggest semantic versioned tags only to suitable action publishers

KengoTODA:
Hello, thanks for making the doc open and easy to discuss!
Here I want to discuss the idea to improve the consistency, hoping that it helps to improve users' experience and security.

### Why:

The [security guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) recommends users avoid using tags when using third-party actions.
See "Pin actions to a full length commit SHA" and "Pin actions to a tag only if you trust
the creator" on the page for detail.

But the ["Releasing and maintaining actions" doc](https://docs.github.com/en/actions/creating-actions/releasing-and-maintaining-actions) suggests all action-authors provide semantic versioned tags, that will motivate action users to use tags instead of full length commit SHA.

Between these official docs, we have an inconsistent description at this moment.

### What's being changed:

This proposal changed semantic versioned tags suggested _only for internal-use actions_.

It is OK for internal-use actions, like private actions used for audits because the author is not a "third-party" for users and it should be trusted by them.

### Check off the following:

Thanks for checking this PR! :wave:

### DIFF
--- a/content/actions/creating-actions/releasing-and-maintaining-actions.md
+++ b/content/actions/creating-actions/releasing-and-maintaining-actions.md
@@ -70,7 +70,10 @@ Here is an example process that you can follow to automatically run tests, creat
 
    * When a release is published or edited, your release workflow will automatically take care of compilation and adjusting tags.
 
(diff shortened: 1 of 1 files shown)


janiceilene:
@KengoTODA Thanks so much for opening a pull request! Our docs are prescriptive and this doesn't follow along with our current guidance. I'm going to close this PR 💛 

Thanks for your interest and passion in improving the GitHub docs ✨

PR #17724, opened 2022-05-09
Fix typo in expressions.md

ivanramosnet:
It should be '<=' instead of '<' for Less than or equal

### Why:

It should be '<=' instead of '<' for Less than or equal

### Check off the following:

### DIFF
--- a/translations/es-ES/content/actions/learn-github-actions/expressions.md
+++ b/translations/es-ES/content/actions/learn-github-actions/expressions.md
@@ -84,7 +84,7 @@ env:
 | `.`                       | Desreferencia de propiedad |
 | `!`                       | No                         |
 | `<`                    | Menor que                  |
-| `<`                    | Menor o igual              |
+| `<=`                   | Menor o igual              |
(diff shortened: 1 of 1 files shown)


RazvyBD:
________________________________

ivanramosnet:
I understand that Github does not accept changes to the files in the translations directory, as mentioned in the [documentation](https://github.com/github/docs/blob/main/contributing/types-of-contributions.md#earth_asia-translations), because they work in collaboration with Crowdin. But in this case, it is not a translation error but typography of a comparison operator.

Maybe you could communicate it to the translation team so that they take it into account since I don't see that it can be contributed to the translation in Crowdin.

ramyaparimi:
@ivanramosnet 
Thanks for opening this! Unfortunately, we are not able to accept changes for translated content. Our translation process involves an integration with an external service at crowdin.com, where all translation activity happens. We hope to eventually open up the translation process to the open-source community, but we're not there yet.

See https://github.com/github/docs/blob/main/contributing/types-of-contributions.md#earth_asia-translations.

PR #17776, opened 2022-05-11
Note major versions are not automatically tagged

### LINKED ISSUE #17071: Clarify That GitHub Actions Don't Automatically Understand Semantic Versioning

Kurt-von-Laven:
The previous wording can be misinterpreted to suggest that GitHub Actions itself automatically tags major versions. Also, remove assurance that pinning the major version protects your workflow from breaking. This is only true if the action you depends on never introduces any bugs that affect you. In other words, it's false in practice.

### Why:

Closes #17071.

### What's being changed:

See [rich diff](https://github.com/github/docs/pull/17776/files?short_path=ef18f68#diff-ef18f687729e1ee0549e9b4324265cf19757637bf6a744d01213c7fcc988399f).

### Check off the following:


Pasio47DBG:
7d66b76

Kurt-von-Laven:
@Pasio47DBG, please see [#17345](https://github.com/github/docs/pull/17345#issuecomment-1116495131).

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -413,7 +413,7 @@ Selects an action to run as part of a step in your job. An action is a reusable
 
 We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
```
jsoref, line 414: This isn't really your thing, but if someone is going to review this, maybe they should apply this change as well at the same time:

```suggestion
We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
```

Afaict there's no such thing as a `docker tag number`, just `docker tags`.

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -413,7 +413,7 @@ Selects an action to run as part of a step in your job. An action is a reusable
...
 - Using the commit SHA of a released action version is the safest for stability and security.
-- Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.
+- If the action publishes major version tags, using them allows you to receive critical fixes and security patches while still maintaining compatibility.
```
jsoref, line 416: I'd lean towards:
```suggestion
- If the action publishes major version tags, using them should allow you to receive critical fixes and security patches while still retaining compatibility (within the abilities of the publisher).
```

In that it's almost worth noting that "yes, an action author may try to not break things, but, they're probably human, so they might anyway".

Kurt-von-Laven: I agree with you completely, and my first draft was more explicit about that fact, but I thought a more minimal change may be more likely to be accepted. Happy to incorporate this change or something similar, but let's see whether GitHub finds bandwidth to review this pull request.

content/actions/using-workflows/workflow-syntax-for-github-actions.md
Kurt-von-Laven: I also agree with this; let's see whether the reviewer from GitHub would prefer a separate pull request with this change. Happy to throw an extra commit into this one though.

jaystyles:
@dependabot merge

jaystyles:
@ dependably close

content/actions/using-workflows/workflow-syntax-for-github-actions.md
martin389: 👋 I'll add this suggestion to the PR, and then check with an engineering team to confirm the preferred wording

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -411,9 +411,9 @@ A name for your step to display on {% data variables.product.prodname_dotcom %}.
...
 
-We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
+We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
 - Using the commit SHA of a released action version is the safest for stability and security.
-- Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.
+- If the action publishes major version tags, using them should allow you to receive critical fixes and security patches while still retaining compatibility (within the abilities of the publisher).
```
martin389, line 416: Added some small edits
```suggestion
- If the action publishes major version tags, you should expect to receive critical fixes and security patches while still retaining compatibility. Note that this behavior is at the  publisher's discretion.
```

martin389:
Thanks for this PR! 👍 I've added some small edits to prepare it for review from an engineering team

content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -365,9 +365,9 @@ A name for your step to display on {% data variables.product.prodname_dotcom %}.
...
 
-We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag number. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
+We strongly recommend that you include the version of the action you are using by specifying a Git ref, SHA, or Docker tag. If you don't specify a version, it could break your workflows or cause unexpected behavior when the action owner publishes an update.
 - Using the commit SHA of a released action version is the safest for stability and security.
-- Using the specific major action version allows you to receive critical fixes and security patches while still maintaining compatibility. It also assures that your workflow should still work.
+- If the action publishes major version tags, you should expect to receive critical fixes and security patches while still retaining compatibility. Note that this behavior is at the publisher's discretion.
```
martin389, line 370:
```suggestion
- If the action publishes major version tags, you should expect to receive critical fixes and security patches while still retaining compatibility. Note that this behavior is at the discretion of the action's author.
```

martin389 (approved):

PR #18126, opened 2022-05-23
Add GHES Reserved Name

alejndr0:
### Why:

The github-enterprise org is automatically created on a new GHES instance, so a user can't create an org with that name.
Closes github/ghes#2361

### What's being changed:

Adding a short paragraph about reserved names in GitHub enterprise instances.
### Check off the following:

### DIFF
--- a/content/admin/configuration/configuring-your-enterprise/site-admin-dashboard.md
+++ b/content/admin/configuration/configuring-your-enterprise/site-admin-dashboard.md
@@ -186,6 +186,10 @@ For example, the following words are reserved, among others:
 
 For the full list or reserved words, navigate to "Reserved logins" in the site admin dashboard.
 
+## Reserved names
+
+When installing {% data variables.product.product_location %} a `github-enterprise` organization is created by default which is used by system services.
+
 {% if ghas-committers-calculator %}
 ## {% data variables.product.prodname_advanced_security %} Committers
 


janiceilene:
👋 @alejndr0 Thanks for opening a PR! It looks like this is still in draft mode. Let me know when it's all ready for a writer review and I'll get it triaged ✨

alejndr0:
@janiceilene Ready! 😄

mattpollard (changes requested):
Thanks for the pull request, @alejndr0 🙇🏻 While @manue1 rightly pointed out that `github-enterprise` is not strictly a "reserved login" in https://github.com/github/ghes/issues/2361#issuecomment-1123563437, I'd say that what we've written already in this article is a clear description that effectively covers `github-enterprise` (despite the product's internals).

> Certain words are reserved for internal use in your GitHub Enterprise Server instance, which means that these words cannot be used as usernames.

`github-enterprise` meets that criteria: GitHub has reserved the username for internal use, and so the reader cannot use it as a username. This is clear and user-centric.

If the reader's "moment of need" is that they're trying to create an account called `github-enterprise` and they cannot, then the best possible experience would be to explain why they can't do that within the product itself, when they try to use the username. The difference in internal implementation isn't relevant to the reader.

Including a list of these usernames here in the documentation creates maintenance burden, as engineering will need to remember to update the list (or let Docs Content know that we need to update the list) whenever we add or remove a reserved name. Also, are there other system accounts that fit this criteria? If so, we should list them in this reference material. If not, we should mention that the list is incomplete—and ideally, point someone to a list of the usernames within the product itself (which doesn't currently exist, as we exclude system accounts within the list of reserved logins).

If we don't want to change the product, and instead want to enumerate any or all system logins within our documentation (or decide that to meet another "moment of need", we want to provide a complete list of system accounts that don't appear in the list of reserved logins), then I don't think this information belongs in "[Site admin dashboard](https://docs-18126-d51c91.preview.ghdocs.com/en/enterprise-server@3.5/admin/configuration/configuring-your-enterprise/site-admin-dashboard)", because this aspect of the product doesn't have anything to do with the site admin dashboard (and no feature within the dashboard exposes these names).

My recommendation would be to close this pull request for now, and create a new (internal) issue [in `github/docs-content`](https://github.com/github/docs-content/issues/new?assignees=&labels=Improve+existing+docs&template=improve-existing-docs.yml&title=%5BImprovement%5D%3A+) to request a more general improvement. The improvement to suggest would be for us to determine a better location for all of our documentation about usernames GitHub's enterprise products reserve (and thus cannot be chosen by end users when creating a new account). We could cross-link to and from that documentation in the "[Site admin dashboard](https://docs-18126-d51c91.preview.ghdocs.com/en/enterprise-server@3.5/admin/configuration/configuring-your-enterprise/site-admin-dashboard#reserved-logins)" article, and thus also increase the likelihood that someone will find what they're looking for via search or browsing the site by providing a dedicated article on reserved usernames.

Let me know if you have questions!

PR #18240, opened 2022-05-26
Clarify order precedence

### LINKED ISSUE #17058: CODEOWNERS file syntax documentation does not make it explicitly clear that a pattern can only be used once
[maintainer edit]
This issue is reserved for the MLH Fellowship program.

### Code of Conduct

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

The existing documentation for file syntax does not make it explicitly clear that a pattern can only be specified once, and that if >1 instance of a pattern is present, the last instance will take precedence.

For example the following example would only request review from `@doctocat`, `@octocat` would not be included in the reviewers.

```
apps/ @octocat
apps/ @doctocat
```

I would propose that a line be added to the `CODEOWNERS syntax` section indicating that a repeated pattern is not supported.

### Additional information

Additionally, I believe an issue should be raised internally against the validation of codeowners files - the above example will be marked as valid by the CODEOWNERS validation, and will not provide any indication to the end user that only the last instance of the pattern will be used.

### Content plan
Found [here](https://github.com/github/docs/issues/17058#issuecomment-1099307334)

ffalor:
### Why:

Closes #17058

As pointed out by @lecoursen there is already an example about last matching pattern taking precedence, but in the example, it is referring to `*       @global-owner1 @global-owner2` as the preceding matching pattern.

Instead of referring to `*       @global-owner1 @global-owner2` in my PR I put a full example directly under the comments since some readers may see the large block of comments as `breaks` indicating the previous example is irrelevant.

### Check off the following:

### DIFF
--- a/content/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners.md
+++ b/content/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners.md
@@ -72,9 +72,10 @@ If any line in your CODEOWNERS file contains invalid syntax, the file will not b
 
 # Order is important; the last matching pattern takes the most
 # precedence. When someone opens a pull request that only
-# modifies JS files, only @js-owner and not the global
-# owner(s) will be requested for a review.
-*.js    @js-owner
+# modifies JS files, only @js-owner and not @doctocat 
+# will be requested for a review.
+*.js @doctocat
+*.js @js-owner
 
 # You can also use email addresses if you prefer. They'll be
 # used to look up users just like we do for commit author


janiceilene:
@ffalor Thanks so much for opening a PR! I'll get this triaged for review ⚡

stevecat:
Hey @ffalor! Thanks for opening this ✨ 

I don't think we should include a broken example, even with the explanation in the comment block above, in "Example of a CODEOWNERS file". I fear folks will miss the comment and try and use it.

Could you fix the example and instead add a new paragraph in "CODEOWNERS syntax" that explains this behaviour? I would suggest adding it under the second paragraph.

Let us know if you have any questions!

japui (commented):
Quando alguém abre um pull request que só
# modifica arquivos JS, apenas @js-owner e não o global. Atenção para não 🚱  prejudicar o sistema.

janiceilene:
It doesn't look like any additional commits have been made to update the example described in https://github.com/github/docs/pull/18240#issuecomment-1141878788 💛  I'm going to close this PR, but @ffalor let us know if you want to continue working on this and we can reopen it!

PR #18270, opened 2022-05-26
Update failure with additional conditions

### LINKED ISSUE #18269: Docs on expressions with extra conditions not clear
### Code of Conduct

https://docs.github.com/en/actions/learn-github-actions/expressions#failure

The docs mention how to use `failure()` to catch failing jobs and then goes on to say it's the same as `job.status == 'failure'` or `github.action_status == 'failure'`, however from my testing this isn't true.

Using either of the above fail to run after a failing job, e.g.

```yml
steps:
  - name: Fail demo
     run: exit 1
  - name: The job has failed
    if: ${{ job.status == 'failure' }}

```

The docs suggest `The job has failed` is meant to run here, but from my testing it doesn't. Not sure if this is a bug with github actions or the docs are just wrong or if I'm just missing something.

I think these broken examples should be removed and a practical example showing how to catch a specific failing job rather than all jobs would be useful here.

### Additional information

_No response_

alexbrazier:
### Why:

Closes #18269

### What's being changed:

The docs in the "Evaluate Status Explicitly" section suggest equivalents of `failure()` but neither work. I've updated the docs to remove them and also added a practical example of only running a failure jobs when a **specific** job fails.

#### Example action with previous broken code and updated working code
```
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: fail demo
        id: demo
        run: exit 1

      - name: action_status - does not run
        if: ${{ github.action_status == 'failure' }}
        run: echo ${{ steps.demo.conclusion }}

      - name: status - does not run
        if: ${{ job.status == 'failure' }}
        run: echo "test"

      - name: failure - runs
        if: ${{ failure() }}
        run: echo "test"

      - name: failure with condition - runs
        if: ${{ failure() && steps.demo.conclusion == 'failure' }}
        run: echo "test"

      - name: condition without failure - does not run
        if: ${{ steps.demo.conclusion == 'failure' }}
        run: echo "test"
```

image

From the above testing `failure()` is **always** required if you want the job to run on a failure regardless of other expressions that might match.

### Check off the following:


albertoEnstein:
Onw

albertoEnstein:
Hub

lucascosti:
👋 @alexbrazier Thanks for opening this PR!

`github.action_status` in your example is not used correctly, as that context only applies to workflows that use a composite action (which your example workflow doesn't). So that value wouldn't be populated, and the behavior you're seeing is expected

However, I have reproduced what you're seeing for `job.status` and `steps.demo.conclusion` in a workflow. I have [echoed the values of those contexts in a step ](https://github.com/lucascosti/actions-public-playground/runs/6646864736?check_suite_focus=true#step:3:7), and they do contain the string `failure`, so I'm not sure why the comparison operator is not working.

I'll see if I can get an engineer or product person to have a look at this, as it might be a bug.

lucascosti:
👋 @alexbrazier, I've confirmed that the behavior is because there is an implicit `success()` condition in `if` conditionals that don't include status check function. This is mentioned in the start of this section, and you can see this behavior in [a workflow run with debug logging here](https://github.com/lucascosti/actions-public-playground/runs/6682566910?check_suite_focus=true#step:5:2).

So, your fix of including `failure()` in the condition works, or alternatively including `always()`. I'll do a small edit to your PR and merge it 🚀

lucascosti (approved):
👍

alexbrazier:
Hi @lucascosti, that makes sense - I had assumed there was either a bug somewhere or an issue with the docs. Thanks for looking into it and merging the changes!

PR #18297, opened 2022-05-28
Clarify requirements on property names

gsrohde:
This change clarifies the requirements on property names when using property dereference syntax.

### Why:

The explanation for the syntax of property names in unclear.  It currently reads:

> In order to use property dereference syntax, the property name must:
>
> - start with `a-Z` or `_`.
> - be followed by `a-Z` `0-9` `-` or `_`.

This could easily be read as only specifying requirements on the first and second characters of a property name.  Moreover, taken literally, it rules out the possibility of one-character property names.

The explanation in the section https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id describing the required format of a job id (for example) is clearer— 

> The `<job_id>` must start with a letter or `_` and contain only alphanumeric characters, `-`, or `_`.

—and I have adopted it here (suitably adapted).

### Check off the following:


cmwilson21:
@gsrohde Thanks so much for opening a PR! I'll get this triaged for review ✨

martin389 (approved):
This is much better, thanks!

PR #18389, opened 2022-06-02
Correct mistyped branch name and mistyped full reference.

### LINKED ISSUE #18386: Confusing presentation of filtering events by ref
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/triggering-a-workflow

and

**Showing the category portion of the namespace of a ref only muddies the presentation**

In the articles _Using Workflows/Triggering a workflow_ (https://docs.github.com/en/actions/using-workflows/triggering-a-workflow) and _Using Workflows/Workflow syntax for GitHub Actions_ (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions), it seems to me not at all useful and potentially confusing to mention the full ref names when talking about filtering by branch and tag names.

As an example, take the section labeled _Example: Including branches and tags_ on the _Triggering a workflow_ page (https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#example-including-branches-and-tags).  The exposition mentions that (for example) a workflow given by
```yaml
on:
  push:
    # Sequence of patterns matched against refs/heads
    branches:    
      - main
      - 'mona/octocat'
      - 'releases/**'
    # Sequence of patterns matched against refs/tags
    tags:        
      - v2
      - v1.*
```
will run whenever there is a push event to (for example) a tag whose name starts with `v1.`, like `v1.9.1` (`refs/tags/v1.9.1`).

While the sample workflow suggests that match patterns are implicitly anchored so that in using the pattern `v1.*` we are matching against `v1.9.1` and not against `refs/tags/v1.9.1`, it seems to me that reader could easily be led to believe that the pattern `refs/tags/v1.*` could have been used here.  Conversely, there seems to me no utility in mentioning `refs/tags/v1.9.1` here since it is clear from the key of the filter pattern list—`branches` or `tags`—which sort of reference the filter pattern is being matched against.

---

Note that the actual files containing these full ref names are the reusables `data/reusables/actions/workflows/section-run-on-specific-branches-or-tags` and `data/reusables/actions/workflows/section-triggering-a-workflow-branches.md`.  Both are used in both articles, and together they contain 9 parenthesized references to branch refs (refs of the form `refs/heads/<branch name>`) and 4 parenthesized references to tags refs (refs of the form `refs/tags/<tag name>`.  I would like to see these all removed.

 

### Additional information

_No response_

gsrohde:
### Why:

In the example of using `branches-ignore` to prevent pull-requests to certain targets from triggering a workflow action, the branch name given in the example wouldn't actually be ignored according to the pattern given.

Also, the full ref given as corresponding to the given branch name is incorrect.

### What's being changed:

Part of the text for the example currently reads

> ... For example, the following workflow would run whenever there is a `pull_request` event unless the pull request is targeting:
> - ...
> - A branch whose name matches `releases/**-alpha`, like `beta/3-alpha` (`refs/releases/beta/3-alpha`)

But a branch name of `beta/3-alpha` _doesn't_ match the pattern `releases/**-alpha` because the branch name doesn't start with `releases/`.  I have changed the branch name used in the example to one that actually _does_ match—namely, `releases/beta/3-alpha`.

Also, the string given as the corresponding _full ref_—`refs/releases/beta/3-alpha`—omits the `heads` component.  If it is to correspond to the given branch name (after my update to it), it should read `refs/heads/releases/beta/3-alpha`.  But as I point out in issue #18386, I see no benefit to mentioning the full ref in these examples since they aren't what the patterns are matched against and are likely to only confuse the reader.  If, per my suggestion, these full refs are deleted from the examples, this portion of my proposed update becomes moot.

### Check off the following:


cmwilson21:
@gsrohde Thanks so much for opening a PR! I'll get this triaged for review :zap:

skedwards88 (approved):
Thanks for this fix! We'll get this merged down for you.

PR #18513, opened 2022-06-09
Update "can not"

13prashant:
Both cannot and can not are acceptable spellings, but the first is much more usual.
We would use can not when the "not" forms part of another construction such as "not only." 
For example: These green industries can not only create more jobs, but also promote sustainable development of the land.

### Why:

Closes [issue link]

### Check off the following:


janiceilene (approved):
Thanks so much for taking the time to open a PR with this fix @13prashant 💖 I'm going to update the branch and get this merged once tests are passing 🍏

PR #18524, opened 2022-06-10
Fixed method for checking cache-hit

schneider10:
When checking cache-hit, this action will return an empty string on no hit, not 'false'. Therefore the correct method is to now check for !=true (see: https://github.com/actions/cache#Skipping-steps-based-on-cache-hit)

### Why:

This doc is out of date or sharing incorrect info for the latest cache action.

### What's being changed:
The documented method to check for a cache hit.

### Check off the following:


japui:
Método corrigido para verificar o acerto de cache #

lucascosti:
> When checking cache-hit, this action will return an empty string on no hit, not 'false'. 

👋 @N-Usha @bishal-pdMSFT Could you have a look at this and see if it might be a bug in the `cache` action? I've [replicated the behavior here (with debug logging on)](https://github.com/lucascosti/actions-public-playground/runs/6930351491?check_suite_focus=true#step:5:46).

In the action's code, [it does look like a `false` value should be set for the output for a cache miss](https://github.com/actions/cache/blob/136d96b4aee02b1f0de3ba493b1d47135042d9c0/src/utils/actionUtils.ts#L26-L28), and there does seem [to be a test for this](https://github.com/actions/cache/blob/136d96b4aee02b1f0de3ba493b1d47135042d9c0/__tests__/actionUtils.test.ts#L91) in the repo too.

N-Usha:
Yes @lucascosti , the doc may need an update.. Including @aparna-ravindra who's verifying this from the code.

aparna-ravindra:
> In the action's code, [it does look like a false value should be set for the output for a cache miss](https://github.com/actions/cache/blob/136d96b4aee02b1f0de3ba493b1d47135042d9c0/src/utils/actionUtils.ts#L26-L28), and there does seem [to be a test for this](https://github.com/actions/cache/blob/136d96b4aee02b1f0de3ba493b1d47135042d9c0/__tests__/actionUtils.test.ts#L91) in the repo too.

Unfortunately, the method in the test case you referred is not used in the action. The value for `cache-hit` is set to false in three cases:
1. Cache feature is not available ([here](https://github.com/actions/cache/blob/d55d005ab0cbd9310b447311bcd1661be90843f5/src/restore.ts#L10)).
2. If there was a cache restored - but on one of the [restore-keys](https://github.com/actions/cache/tree/main#inputs) and not the primary key ([here](https://github.com/actions/cache/blob/d55d005ab0cbd9310b447311bcd1661be90843f5/src/restore.ts#L52))
3. Any unhandled exception ([here](https://github.com/actions/cache/blob/d55d005ab0cbd9310b447311bcd1661be90843f5/src/restore.ts#L61)).

However, if no cache was restored (either on the primary key or the restore-keys), then the `cache-hit` output value is not set. 
Therefore, the right way to test is to check for `cache-hit != true` as we see in the examples [here](https://github.com/actions/cache#Skipping-steps-based-on-cache-hit)

lucascosti:
> @lucascosti, Unfortunately, the method in the test case you referred is not used in the action. The value for `cache-hit` is set to false in three cases:

Thanks a lot for confirming this, @aparna-ravindra! 🙇‍♂️  My next question (possibly more for @N-Usha) is this behavior deliberate? If `cache-hit` is a boolean for determining if a cache hit on `key` happens, shouldn't it return `false` for any situation where there wasn't a cache hit on `key`?

I guess I'm trying to think of a case where returning a `null` value for when no cache was restored would be the desired/expected outcome. 🙃 I would have expected a null value returned for situations 1. (Cache feature is not available) and 3. (Unhandled exception) in the post above.

If it is deliberate, I'll take care of getting the docs updated (e.g. [this section](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-output-of-the-cache-action) will also need updating).

aparna-ravindra:
Given that there are two possible values (null and false) that indicate cache-miss, checking for `cache-hit != true` will address both cases. Hence, I vote for updating the doc and thereby keeping it consistent with the example in the actions/cache public repository as well.

jsoref:
Sounds like this is related to https://github.com/actions/cache/pull/717 / https://github.com/actions/cache/issues/661#issuecomment-1019511446

aparna-ravindra:
@lucascosti Please feel free to ping if you need any more inputs from me.

aparna-ravindra:
@lucascosti ,
we have a couple more requests (like [this](https://github.com/actions/cache/pull/876) one) asking for doc update. How do we go about triaging this PR?

lucascosti:
> @lucascosti , we have a couple more requests (like [this](https://github.com/actions/cache/pull/876) one) asking for doc update. How do we go about triaging this PR?

Sorry, @aparna-ravindra; I've been on some extended leave and this one slipped through the cracks. I'll get this reviewed and merged. 👍

lucascosti (dismissed):
Although being able to match on a cache hit on a `false` result would also be ideal, these changes more accurately describe the action's current behavior. 👍

lucascosti (approved):

bchen1029:
@lucascosti this PR help setting cache hit on `false` when no hit. I think this is ideal solution as you do. Could you help check this PR if you're free? really appreciate!

lucascosti:
Sorry, @bchen1029; the Actions engineering team will have to review those changes; I'm a docs writer and defer to them for any functionality changes.

PR #18603, opened 2022-06-13
fix: fixed Actions large secret instructions and added warning

titanism:
### Why:

Instructions at <https://docs.github.com/en/github-ae@latest/actions/security-guides/encrypted-secrets#limits-for-secrets> are not thorough enough for the average reader or developer to understand.  More context needs to be added, as well as file paths, folders, and introduction (and links) for `.github` folder in order for developers to understand the purpose of it.  Additionally, there is no warning to readers to NOT copy the unencrypted non-gpg file.

### What's being changed:

Here are the changes we made: <https://github.com/github/docs/pull/18603/commits/bb09c1ca835804ca3da08801e6b0b0582677e5ac>.

We've added several things:

1. Added a note to explicitly copy the `gpg` file extension file and not the unencrypted one.
2. Added examples showing how to copy the file to your GitHub project.
3. Added complete file path for `decrypt_secret.sh` (as there was a doc typo in that a later command showed it being in the `.github/scripts` folder and the initial commands did not state this in the examples).

### Check off the following:


janiceilene:
Thanks for opening a PR 👍 Please fill out the [pull request template](https://raw.githubusercontent.com/github/docs/main/.github/PULL_REQUEST_TEMPLATE.md) in the top comment in this pull request, then we'll get this up for review!

japui:
instruções secretas corrigidas e aviso adicionado

janiceilene:
Thanks so much for filling that out @titanism! I've triaged this PR for review 🌟

lucascosti:
👋 @titanism Thanks a lot for pointing out this procedure could use some improvement! You're right in that there were errors in it. 

Your original changes were good, but I took the opportunity to expand on them to make the procedure fit in more with our styles, including giving it its own heading for easier discoverability.

The main thing that I changed from your suggestions was not to suggest using subfolders in `.github/`. These folders generally have special meaning for GitHub features, and although it's not 'wrong' to put other permanent repo files in there, we'd prefer not to show it in the docs where it doesn't relate to a feature. (this ideally shouldn't have been there in the existing procedure, too 😊).

I've opted for storing the files for this procedure in the root of the repo; I think people following along will have the knowledge to replace paths with their own if they decide to nest it in other repository folders.

I've tested the updates from start to finish and confirmed it works. 🎉 

Thanks again for helping improve this! 🙇‍♂️

lucascosti (approved):

PR #18935, opened 2022-07-05
Different defaults with and without protection

### LINKED ISSUE #18936: Explain different default contexts (fx with and without branch protection)
### Code of Conduct

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches

Some note regarding "default" behavior explicitly, would be helpful.

The documentation may be strictly correct, but certain assumed or non-obvious interpretation of "default" is likely confusing (it was for me). I assume "default" means "always without any overriding setting there is one and only one presumed value". To my surprise, "default" is contextual and may mean the exact opposite in different contexts.

The specific example I ran into (though there are numerous examples):

By default, in both git and github, 'git push --force' is allowed, even to master.
However, with branch protection enabled, then by default 'git push --force' is not allowed, by default.
The documentation clearly states that allow_force_pushes = false is default.
But it's only default in that context and not default (opposite default) in the default context (no branch protection).

### Additional information

_No response_

alexgenaud:
Out of the box, (git push --force) is allowed "by default". With branch protection, it's not allowed "by default". These distinctions are not obvious because of two "default" contexts.

I struggle to explain that here and in text, but I think it's important to note, somehow. It certainly confused me.

Perhaps better would be a table of default values.
Columns to include: "default git", "default github", "default branch protection", "default [other contexts]";
and rows including all the different settings such as "enforce_admins", "allows_force_pushes", "allows_deletions", etc.

### Why:

Why? I was confused by what "default" means. We configure our hundreds of repositories via Terraform. It seemed to me there was no point in setting "default" values to their "default" state. However, it was not clear to me that "default" has a different meaning based on context (with or without branch protection).

Closes [18936] "Explain different default contexts (fx with and without branch protection)"
https://github.com/github/docs/issues/18936

I offer an short paragraph, although I expect a side note box; or a table of settings, defaults, overrides; or rewording of extant paragraphs are likely better changes. 

### Check off the following:

### DIFF
--- a/content/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches.md
+++ b/content/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches.md
@@ -29,6 +29,8 @@ topics:
 
 You can enforce certain workflows or requirements before a collaborator can push changes to a branch in your repository, including merging a pull request into the branch, by creating a branch protection rule.
 
+Simply setting branch protection will set certain "default" behavior. For example, normally force pushes are allowed (git push --force). However, by enabling branch protection, then forced pushes are not allowed (allow_force_pushes = false, by default) unless explicitly re-enabled ( = true).
+
 By default, each branch protection rule disables force pushes to the matching branches and prevents the matching branches from being deleted. You can optionally disable these restrictions and enable additional branch protection settings.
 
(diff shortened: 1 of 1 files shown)


cmwilson21:
@alexgenaud Thanks so much for opening a PR! I'll get this triaged for review :zap:

lecoursen:
Thanks @alexgenaud for your interest in GitHub Docs! The sentence just below the one you've added already says the same thing so I'll close this PR.

PR #19676, opened 2022-08-11
Add the copy button to the code blocks

rajbos:
### Why:
Add the copy button to the code blocks

Add the copy button to the code blocks on this page

### Check off the following:


rajbos:
Interesting to see that the changes in `data/reusables/gpg/list-keys-with-note.md` break the list items in the output, while the other reusable file change works as expected.

image

I can update the PR without this file, but rather have it fixed in a better way :-D.

cmwilson21:
@rajbos Thanks for opening a PR and pointing out the broken output. Even though it's still a draft, I'm going to triage it for the team to get eyes on. Maybe there's a solution for the breakage. ⚡

lukins (commented):
Shell copy

cmwilson21:
👋 Hey again @rajbos! I spoke with the team and there may be one too many spaces leading up to the code block in the second reusable. Would you mind removing one of the spaces before the first ` ``` ` ?

rajbos:
Good catch. That space was already in the original file as well, as well as it already is in the other files as well. I did remove it but the redeploy does not display:

`504.0 GatewayTimeout`

Is there a way to force the redeploy so we can validate?

aitemour79:
Bouddi OK tomorrow

cmwilson21:
I'm not sure what was causing the failure in the redeployment. I'm switching it from a draft to an opened pull request to see if that pushes through. I've marked it as "On hold" so it won't be on the review board until you are ready ✅

cmwilson21:
Hmm, that still didn't fix it. It's interesting because it looks like the note section from the previous reusable (list-keys-with-note) is leaking down into the broken one (copy-gpg-id-key). I still wonder if it has to do with spacing in the list-keys-with-note reusable. Maybe if we were to line them all up per the example in one of our contributing [files](https://github.com/github/docs/blob/main/contributing/content-markup-reference.md#code-sample-syntax-highlighting), it will fix the leakage?

rajbos:
Testing that out. I think the are there (and might need to be) because they are used within list-items (either numbered or not)

data/reusables/gpg/list-keys-with-note.md
```diff
@@ -1,6 +1,6 @@
 
 1. Use the `gpg --list-secret-keys --keyid-format=long` command to list the long form of the GPG keys for which you have both a public and private key. A private key is required for signing commits or tags.
-  ```shell
+```shell{:copy}
```
cmwilson21, line 3:
```suggestion
   ```shell{:copy}
```

data/reusables/gpg/copy-gpg-key-id.md
```diff
@@ -1,9 +1,9 @@
 1. From the list of GPG keys, copy the long form of the GPG key ID you'd like to use. In this example, the GPG key ID is `3AA5C34371567BD2`:
-  ```shell
+ ```shell{:copy}
```
cmwilson21, line 2:
```suggestion
    ```shell{:copy}
```

data/reusables/gpg/list-keys-with-note.md
```diff
@@ -1,6 +1,6 @@
...
   $ gpg --list-secret-keys --keyid-format=long
 ```
```
cmwilson21, line 5:
```suggestion
   ```
```

cmwilson21:
👋 @rajbos I think told you opposite of what was needed 🙈 

I ran this by the team and the thought is that the code block start  ``` on line 3 in the list-keys-with-note.md reusable needs to be tabbed in so that it is in line with the start of the list item above it (the word `Use` here). The break may be happening because the MD can’t tell that the codeblocks are part of the list items without that extra indentation.

rajbos:
> ran this by the team and the thought is that the code block start ``` on line 3 in the list-keys-with-note.md reusable needs to be tabbed in so that it is in line with the start of the list item above it (the word `Use` here). The break may be happening because the MD can’t tell that the codeblocks are part of the list items without that extra indentation.

Seems reasonable, let's see if it will deploy this time 🙌 .

rajbos:
Nope, no success here: 
image

Link: https://docs-19676-b4d73e.preview.ghdocs.com/en/enterprise-server@3.5/authentication/managing-commit-signature-verification/generating-a-new-gpg-key

cmwilson21:
I spoke with the team again and made a couple of changes regarding the spacing (again). I think that's got it this time! I appreciate your patience as we've worked through this one. 👍

Screen Shot 2022-08-26 at 10 04 36 AM

I'll get this merged down, soon! Thanks for the addition of the copy button! 💖

cmwilson21 (approved):
LGTM! ✨

cmwilson21:
Thanks for adding the copy buttons! And for working with us as we were figuring out the breakage. 💖

PR #19982, opened 2022-08-19
Documentation on how to use the release trigger is wrong and confusing.

devantler:
### Why:

The documentation does not correctly describe how to make a release that triggers when you publish a release or change the toggle to prerelease and update the release. To do this, you need to use both the `released` and `prereleased` option. This contradicts what was written in the deleted note.

The toggle mentioned above is this one:

Screenshot 2022-08-19 at 11 40 45

- Deleted note that says you should use `published` to trigger when a prerelease or a release is happening. This is wrong. With `published` it only triggers the first time a release is published.
- Added info on how to trigger workflows with release in typical scenarios, e.g., when you want to trigger each time a prerelease or release is published or edited from prerelease to release or vice-versa.

### Check off the following:

### DIFF
--- a/content/actions/using-workflows/events-that-trigger-workflows.md
+++ b/content/actions/using-workflows/events-that-trigger-workflows.md
@@ -1067,16 +1067,17 @@ on:
 
 {% endnote %}
 
-{% note %}
-
-**Note:** The `prereleased` type will not trigger for pre-releases published from draft releases, but the `published` type will trigger. If you want a workflow to run when stable *and* pre-releases publish, subscribe to `published` instead of `released` and `prereleased`.
-
-{% endnote %}
-
 Runs your workflow when release activity in your repository occurs. For information about the release APIs, see "[Release](/graphql/reference/objects#release)" in the GraphQL API documentation or "[Releases](/rest/reference/releases)" in the REST API documentation.
 
-For example, you can run a workflow when a release has been `published`.
+For example, you can run a workflow whenever a release or a prerelease is published, including if you change a prerelease to a release.
 
+```yaml
+on:
+  release:
+    types: [released, prereleased]
+```
+
+If want to only run a workflow the first time a prerelease or a release is published you do as follows:
 ```yaml
 on:
   release:


cmwilson21:
@devantler Thanks so much for opening a PR! I'll get this triaged for review :zap:

content/actions/using-workflows/events-that-trigger-workflows.md
```diff
@@ -1067,16 +1067,17 @@ on:
 
 {% endnote %}
 
-{% note %}
-
-**Note:** The `prereleased` type will not trigger for pre-releases published from draft releases, but the `published` type will trigger. If you want a workflow to run when stable *and* pre-releases publish, subscribe to `published` instead of `released` and `prereleased`.
-
-{% endnote %}
-
```
lucascosti, line 1075: I've done some testing, and it looks like the current advice _is_ correct, but I do understand that it is confusing for your use case.

* ``The `prereleased` type will not trigger for pre-releases published from draft releases`` this is still true.
* ``If you want a workflow to run when stable *and* pre-releases publish, subscribe to `published` instead of `released` and `prereleased``, this is also true.
 
  But, you are correct that `published` does not trigger a workflow run if you edit an existing (and already published) prerelease to change it into a release (or vice-versa). This is because the 'release' was already published, and you are changing an already-published release.

I think the only way to cover all of these situations is to trigger on all three types (i.e. `types: [published, released, prereleased]`), but it does generate duplicate workflow runs for the common use case of publishing a release without having a draft first.

Before we change the docs on this, I will try to get somebody from the releases team to weigh in on what might be the best approach here.

devantler: I just tried the following:

1. Create a draft release with a new tag, a title and a description.
2. Save the draft.
3. Leave the page.
4. Open the draft again.
5. Publish the draft as a prerelease.
6. The action starts and deploys the project.

My action uses the following triggers:

```
  release:
    types: [prereleased, released]
    tags:
      - "v*"
 ```
 
Maybe this works because of the `tags: -"v*"`, but it is very confusing given the docs say this: "The `prereleased` type will not trigger for pre-releases published from draft releases".
 
I would at least suggest that this approach is mentioned, as it works flawlessly to trigger releases when:

1. A prerelease or a release with a tag is published from a draft release
2. A prerelease or a release with a tag is published from a non-draft release
3. A prerelease or a release is edited and published a new, e.g. if you want to make a prerelease into a release, without needing to publish a new release.

I think that covers most bases, and for me, that feels like a good workflow supported by the options in GitHub releases. It allows a good workflow when having a development (on push), staging (on prerelease), and production (on release) environment.

Comment (lucascosti) [reply content/actions/using-workflows/events-that-trigger-workflows.md:1075]: Thanks for the extra detail, @devantler! I'll check in with the product and engineering people and report back.

Comment (lucascosti) [reply content/actions/using-workflows/events-that-trigger-workflows.md:1075]: Hi @devantler, sorry for the delay!

```
  release:
    types: [prereleased, released]
    tags:
      - "v*"
```

I've checked with engineering, and it looks like using `tags` under a `release` trigger is not supported; it doesn't seem to filter the tags, e.g. [here's a test workflow run](https://github.com/lucascosti/actions-public-playground/runs/8222129301). It's not something we currently document.

> I just tried the following:
> 
>  1. Create a draft release with a new tag, a title and a description.
>  2. Save the draft.
>  3. Leave the page.
>  4. Open the draft again.
>  5. Publish the draft as a prerelease.
>  6. The action starts and deploys the project.

I can't reproduce this with a draft pre-release. Before step 2, did you check the "This is a pre-release" check box **before** saving it as a draft? If not (i.e. the draft was saved as a 'normal' release, but when it was published it was published as a pre-release), then the workflow should run, and the behavior is consistent with the document wording.

devantler: You are right. The documentation is correct. The pre-releases are not triggered from draft releases.

It seems that the `prereleased` and `released` events only trigger on publish if they have changed between the draft and the published release. For me, this seems odd, as I would expect any draft release that is later published, should consider all its events at that point. Do you know why this is not the default behaviour? 

For now, I made my workflow work by using the `published` event with `if` conditions that checks if an action is triggered by a release, and what type of release. This is used to do rolling deployment from dev (on push) to staging (on release if github.event_name == 'release') to production (on release if github.event_name == 'release' && !github.event.release.prerelease).

lucascosti: Thanks for confirming, @devantler. I'll close this PR.

> Do you know why this is not the default behaviour?

I would suspect that it's probably due to the release webhooks that Actions integrates with. If you'd like to prove the product team feedback on this, I would suggest letting them know via GitHub support or posting in the https://github.community/ discussions. 🙇‍♂️

PR #20265, opened 2022-08-29
Fix example OIDC field repository_visibility

yogeshlonkar:
Created issue https://github.com/github/docs/issues/20266


yogeshlonkar:
Created issue #20266

PR #20572, opened 2022-09-13
Specify that GITHUB_REF is a full ref name

dgholz:
I can't tell the difference between the `GITHUB_REF` and `GITHUB_REF_NAME` environment variables available in Actions, and the docs describe them in the same terms.


cmwilson21:
@dgholz Thanks so much for opening a PR! I'll get this triaged for review :zap:

lucascosti (approved):
Thanks for this clarification, @dgholz! I've made some minor edits, and I'll get this emrged in. 🙇‍♂️

PR #20725, opened 2022-09-20
Adding larger hosted runners option

datanerd:
Adding the larger hosted runners with fixed IP address as an alternative to self-hosted runners when using allowlist.

### Why:

With larger hosted runners release, there is an option for customers to use a runner with a fixed IP address and add that address into allow list.

### Check off the following:


cmwilson21:
👋 @andrekolodochka Thanks so much for opening a PR! I'll get this triaged for review :zap:

lucascosti (approved):
Thanks for this, @andrekolodochka! I pushed up a small change to add versioning to the larger runners mentions just in case we decide to version it for other products in the future. I'll get this merged!

PR #20883, opened 2022-09-26
expressions.md: Provide an example of negation

cclauss:
### Why:

`! contains()` (with a space) invalidates a workflow file and its jobs will not run.  `!contains()` (no space) works as expected.  Unfortunately, the GitHub web UI editor suggests the _exact opposite_ of the correct syntax.  It marks the solution that works as an error and the solution that does not work as being correct.

### Check off the following:


cmwilson21:
@cclauss Thanks so much for opening a PR! I'll get this triaged for review :zap:

lucascosti:
👋 Hi @cclauss, thanks for this PR!

> Unfortunately, the GitHub web UI editor suggests the exact opposite of the correct syntax.

I'm not 100% what you're referring to here, but I suspect you might be using it with an `if` condition like:

```
if: !contains('Hello world', 'llo')
```

which, if that is the case, it is invalid because YAML doesn't allow an unquoted value to start with a special character. To escape the negation special character (`!`), you would need to use the expression syntax, such as:

```
if: ${{ !contains('Hello world', 'llo') }}
```

For your PR, I don't think a negated example is necessary here (for example, none of the other functions contain them).

Possibly a better change would be to mention the escaping requirement in the [About expressions](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions) section.

e.g instead of 

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. For more information about if conditionals, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/articles/workflow-syntax-for-github-actions/#jobsjob_idif)."

we could mention the special character requirement:

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) if the expression does not start with a special character, because GitHub automatically evaluates the if conditional as an expression. For more information about if conditionals, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/articles/workflow-syntax-for-github-actions/#jobsjob_idif)."

Let us know if you'd like to update this PR with that change :bow

cclauss:
@lucascosti Please try the `if` statements in a real GitHub Action.
The one that works is syntax highlighted as an error.
The one that does not work it has no syntax highlighting.

cclauss:
I have updated the pr with the requested change.

lucascosti:
> I have updated the pr with the requested change.

Hi, @cclauss, it looks like you might have misread my suggestion.

You're current changes (`if: {% raw %}${{ ! <expression> }}{% endraw %}`) isn't totally accurate because `!` is actually part of an `expression`.

Here's my suggestion again:

> Possibly a better change would be to mention the escaping requirement in the [About expressions](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions) section.
> 
> e.g instead of
> 
> > When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression. For more information about if conditionals, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/articles/workflow-syntax-for-github-actions/#jobsjob_idif)."
> 
> we could mention the special character requirement:
> 
> > When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) if the expression does not start with a special character, because GitHub automatically evaluates the if conditional as an expression. For more information about if conditionals, see "[Workflow syntax for GitHub Actions](https://docs.github.com/en/articles/workflow-syntax-for-github-actions/#jobsjob_idif)."
> 
> Let us know if you'd like to update this PR with that change 🙇‍♂️

PR #21034, opened 2022-10-02
chore: fix render syntax

schalkneethling:
It looks like `shell` is no longer a valid value from the `render` property. When I use it in a YAML template, I get a "value is not accepted" error in VSCode. I scrolled through the list of possible values, and the closest one seems to be `console`.

vscode error dialog showing the value is not accepted error

### Check off the following:


steveward:
Thanks for the PR @schalkneethling. Shell is still listed as valid (see https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L6045), and I just tested to confirm it still works in a test template. I'm not sure what's causing that error to get thrown in VSCode, but since this still works as expected I'm going to close this out.

PR #21351, opened 2022-10-14
Update searching-for-repositories.md

screig:
The search `jquery in:name` will only find repositories that **start** with the search term jquery, rather than repos that contain this search term **in** their name. 

https://github.com/search?q=jquery+in%3Aname   (122,150 repository results)

For example if one changes the search to `query in:name`, i.e. drop the j, then if the original description in this file was correct it should find more repos. However as one can see by running this query

https://github.com/search?q=query+in%3Aname 

you get less results, 81,119. This is because the examination is not actually searching if the word "query" is **in** the name but rather does the repo **start** with the word "query".

Just to prove the point lets search for repos that have "_uery_" in their names, `uery in:name`. If the code really was searching for this search term IN the name it should at least find as many results as in  `jquery in:name`.

https://github.com/search?q=uery+in%3Aname

We only get 8.

### Why:

Closes [issue link]

### Check off the following:

### DIFF
--- a/content/search-github/searching-on-github/searching-for-repositories.md
+++ b/content/search-github/searching-on-github/searching-for-repositories.md
@@ -27,7 +27,7 @@ With the `in` qualifier you can restrict your search to the repository name, rep
 
 | Qualifier  | Example
 | ------------- | -------------
-| `in:name` | [**jquery in:name**](https://github.com/search?q=jquery+in%3Aname&type=Repositories) matches repositories with "jquery" in the repository name.
+| `in:name` | [**jquery in:name**](https://github.com/search?q=jquery+in%3Aname&type=Repositories) matches repositories that start with "jquery" in the repository name.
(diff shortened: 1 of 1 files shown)


cmwilson21:
@screig Thanks so much for opening a PR and providing your examples! I'll get this triaged for review :zap:

sabrowning1:
Hi @screig, thanks again for bringing this to our attention and providing so much detail!

I followed your testing, and technically, it looks like searches for `query in:name` will return things like `blueimp/jQuery-File-Upload` and `camsong/You-Dont-Need-jQuery`, so it does successfully return some repositories, but you're right that the number of results decreases significantly from `jquery in:name`, and it isn't functioning as described. This is a little bit too detailed of an update to add to the docs as of now, but we have started work on it internally 👍🏼 thanks again for your help maintaining our docs! ✨

screig:
Bit of a disappointing response, you are leaving docs live that suggest features which in practise don't exist. I agree changing the underlying functionality is a big job, however by refusing to say in the docs that it does not work as described is very much like ignoring the problem.

PR #21705, opened 2022-10-31
Added OpenSSH service on Windows alternative

rdev32:
### Why:

Windows users have more options than using only git bash

I added content to [this](https://github.com/github/docs/blob/main/content/authentication/connecting-to-github-with-ssh/working-with-ssh-key-passphrases.md) file

image

### Check off the following:


mattpollard (commented):
Thanks for the contribution, @rdev32! 🙇🏻 Generally, we prefer to provide one method of accomplishing a goal on GitHub Docs. There are three reasons that I'm hesitant to incorporate your changes.

1. The reader may not have the required access on the system to enable a service, so the instructions for using `ssh-agent` within Git Bash are more likely to work for a larger number of readers.

2. It appears that the OpenSSH authentication service is only available in Windows 10 and later, which means we'd need to specify that prerequisite for readers, increasing complexity.

3. I did a bit of reading on this service and found https://github.com/PowerShell/Win32-OpenSSH/issues/1487, which seems to indicate that it's possible to reconstruct a private key from the registry entries that the service creates. I don't fully understand the risk associated with this attack vector, but this is another argument for providing only the most targeted, local instructions for Git Bash.

We'll close this pull request for now, but let us know if you have any questions or further context 🙏🏻

PR #21768, opened 2022-11-03
Add `strategy` to keywords for jobs that call workflows

qubitz:
### Why:

Buddy and I brainstormed using `matrix` values when it came to calling our reusable workflow and we 
were bummed to discover [it wasn't supported](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow). Low and behold [it is](https://github.com/qubitz/workflow-testbed/actions/runs/3388522235).

### What's being changed:

Mainly just added `strategy` to the list of [supported keywords for jobs that call reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow). Also, I figured I'd throw `strategy`'s use into the example as well, though it is a bit contrived, suggestions welcome. Or maybe leave it out of any example?

Here is how the [rendered changes](https://docs-21768-74a3d0.preview.ghdocs.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow) look:
image

### Check off the following:

- [x] I have reviewed my changes in staging


cmwilson21:
@qubitz Thanks so much for opening a PR! We appreciate the screenshoot, too! I'll get this triaged for review :zap:

lucascosti (approved):
Thanks a lot for this PR, @qubitz!

Yep, you are right; a recent addition to this feature was being able to use `strategy`, and we [currently have a section for it here](https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-a-matrix-strategy-with-a-reusable-workflow).

🤔 That section should probably be nested under the 'Calling a reusable workflow' heading (which is maybe why you missed it), so I've pushed up a commit that moves it there, and also adds the required versioning to your addition to the 'Supported keywords' section.

I'll get these changes merged! Thanks again 🚀

PR #22463, opened 2022-12-05
Use a random delimiter in examples

martinthomson:
It was a bit jarring seeing a strong warning about choosing random delimiters, then an example immediately afterwards that didn't follow that advice.  This fixes that.

I'm not 100% confident with `base64` here.  The inclusion of "/" and "+" characters should be fine, but sometimes they are treated specially in ways that might not produce good results.

Also, my powershell isn't that strong, but this seems to do the trick (with slightly lower entropy as I don't include "+" and "/").

FWIW, 120 bits of randomness (or close to that) should be plenty.  Using a multiple of three bytes ensures that `base64` doesn't add any awkward "=" characters.  I'd use hex, but then I'm less confident in the availability of `xxd`.

### Why:

Closes [issue link]

### Check off the following:


cmwilson21:
@martinthomson Thanks so much for opening a PR! I'll get this triaged for review :zap:

lucascosti (approved):
This is great, @martinthomson, thanks for putting the effort into this! 🙇‍♂️ 

This looks good to me from a docs POV, and I've tested that the examples work. About 10 runs each of the randomness commands seems not to include any characters that would break it.

I'll get someone from the engineering team to double check them before we merge this. 👍

thboop (approved):
LGTM

Jackpoot:
)

Jackpoot:
Binance id 304141410

martinthomson:
@soliton- , The random values here are 15 bytes in length.  That turns into 20 characters.  It's a little tricky to work out the exact odds of a collision, because that depends on what you are delimiting.  However, for any given 20 character sequence, the odds that you will produce a collision is 2<sup>-120</sup>.  Sure, with a large file you get many attempts, but even if you have random base64 data, you need to have in the order of 2<sup>60</sup> bytes before the chance of a collision is meaningful.  I don't know about you, but most people won't be including that much data.  If you are sending that much data, increase the length of the delimiter by a few bytes.

soliton-:
@martinthomson Normally you pick a delimiter that in principle cannot occur in the values in your specific workflow and you're done. If you absolutely must then pick the delimiter at random but there is no point in picking a different one each run. It just makes this example more complex and will make it all the more confusing if you get an error then and don't even know what the delimiter was.

Issue #22565, opened 2022-12-09
Transferring ownership article(s) should contain pointers to how/where transfer request can be approved


derisen:
### Code of Conduct

- [X] I have read and agree to the GitHub Docs project's [Code of Conduct](https://github.com/github/docs/blob/main/CODE_OF_CONDUCT.md)

### What article on docs.github.com is affected?

https://docs.github.com/en/developers/apps/managing-oauth-apps/transferring-ownership-of-an-oauth-app

### What part(s) of the article would you like to see updated?

The article should mention what happens when a transfer request is made and how/where the transfer request can be approved. 

Alternatively, it can contain a link to a resource that covers this information.

### Additional information

I'm trying to transfer the ownership of several OAuth apps to an organization. However, the organization admins are unsure where they can approve the request. It would be great if the article makes a mention of this.

----
_Maintainer update_

The proposed updates to fix this issue are in this comment: https://github.com/github/docs/issues/22565#issuecomment-1726051141.

cmwilson21:
@derisen Thanks so much for opening an issue! I'll triage this for the team to take a look :eyes:

RobertEllisClick55:
Thank you

skedwards88:
Thanks for opening this issue! I am going to leave it here in the open source repo for anyone to contribute to, but I do not currently have the information to document. I am also making an internal note in our backlog to research and document this.

derisen:
sounds great -thanks so much!

Venkatreddy17:
@derisen I am interested taking this issue. I found about docs through friend.

skedwards88:
Once you initiate the transfer, the new owner needs to navigate to their OAuth Apps page. From there, they should see the app listed under "Pending transfer requests". They need to click on the app name and then click **Complete transfer**.

CBID2:
Hi @cmwilson21 and @isaacmbrown, I made PR for this issue. It's #29464

PR #22606, opened 2022-12-10
Fix RateLimit.resetAt format

FiloSottile:
It's a DateTime not a UNIX timestamp.


Deesy31 (commented):
> #22565

skedwards88 (approved):
Thanks for this fix! We'll get it merged down for you.

PR #22918, opened 2023-01-01
format events-that-trigger-workflows.md

fyc09:
There should be a space between \- and \`. I added a space between them.


cmwilson21:
@Eric-fuyc Thanks so much for opening a PR! I'll get this triaged for review :zap:

felicitymay (approved):
Thanks for spotting and fixing this formatting issue ✨ 

It took me a while to see the difference your change made to formatting (typically, a missing spaces results in more obvious formatting problems). However, I can see that the column of "Activity types" is correctly aligned in the [preview](https://docs-22918-c190a1.preview.ghdocs.com/en/actions/using-workflows/events-that-trigger-workflows#check_run) 👍🏻

PR #23208, opened 2023-01-13
Update reusing-workflows.md

### LINKED ISSUE #23207: Improve Workflow Reuse Docs around Passing Values
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/reusing-workflows

The example for how values are passed.  It looks like maybe the wrong value name is used in the calling workflow.

### Additional information

_No response_

feanil:
### Why:
The caller is passing in  `repo-token` but the called workflow expects `envPAT`.  This is either incorrect or something is happening to get the right values passed and it should be further explained in the doc.

Closes https://github.com/github/docs/issues/23207

### Check off the following:

### DIFF
--- a/content/actions/using-workflows/reusing-workflows.md
+++ b/content/actions/using-workflows/reusing-workflows.md
@@ -142,7 +142,7 @@ You can define inputs and secrets, which can be passed from the caller workflow 
        steps:
        - uses: actions/labeler@v4
          with:
-           repo-token: ${{ secrets.envPAT }}
+           envPAT: ${{ secrets.envPAT }}
            configuration-path: ${{ inputs.config-path }}
    ```
    {% endraw %}


cmwilson21:
@feanil Thank you for opening a PR and linking it to your issue! ✨ 

lucascosti:
Hi @feanil, the doc is correct.

The secret specified in step 1 is used as `${{ secrets.envPAT }}` in step 2.

What it is being assigned to doesn't need to have the same name. It's being assigned to an input for an action used in the code example in step 2 (`uses: actions-labeler@v4`), and [that action expects an input with the name `repo-token`. ](https://github.com/actions/labeler#inputs)

It's the same convept as the reusable workflow input `config-path`, being used as another input for the action as `        configuration-path: ${{ inputs.config-path }}` in the next line.

feanil:
@lucascosti thank you for the clarification! That makes sense! I think my confusion came from not realizing that the code block above and this one are in the same file and referencing the reusable workflow that is in a third file,  I thought that when the environment variable is defined it is automatically passed through to the re-usable workflow but that is not the case.

PR #23456, opened 2023-01-24
improve steps: naming PAT is not optional

ayodejii:
Naming public access token is not optional, but the doc says it is.

This screenshot shows what happens if you do not name your PAT.

image


cmwilson21:
@ayodejii Thanks so much for opening a PR! I'll get this triaged for review :zap:

felicitymay (approved):
Many thanks for opening a pull request to fix this error in the docs 💖 

Your change looks spot on. I'll mark it as ready to merge.

PR #23824, opened 2023-02-09
Update reusing-workflows.md

### LINKED ISSUE #23823: Grammatical error in docs for reusable workflows
### Code of Conduct

[Reusing Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#access-to-reusable-workflows)

Under `Access to reusable workflows`, the line `A reusable workflow can be used by another workflow if either of the following is true:` references a list of three items. The logic in the source file attempts to use "either" when only two items are in the list, but "any" when there are more. This check doesn't account for `private-actions` being set.

### Additional information

_No response_

tvalenta:
Consider `version private-actions` when selecting "any" or "either" for the section on Access to reusable workflows

### Why:

Closes #23823 23823

This adds `private-actions` to the list of versions considered when creating a list of assertions for which one must be true in order to reuse workflows.

### Check off the following:


cmwilson21:
@tvalenta Thank you for opening a PR and linking it to your issue! ✨ 

content/actions/using-workflows/reusing-workflows.md
```diff
@@ -51,7 +51,7 @@ For more information, see "[Creating starter workflows for your organization](/a
 
 ## Access to reusable workflows
 
-A reusable workflow can be used by another workflow if {% ifversion ghes or ghec or ghae %}any{% else %}either{% endif %} of the following is true:
+A reusable workflow can be used by another workflow if {% ifversion private-actions or ghes or ghec or ghae %}any{% else %}either{% endif %} of the following is true:
```
felicitymay, line 54: It looks as if users will always see at least 3 bullets here so we can remove this condition altogether.

```suggestion
A reusable workflow can be used by another workflow if any of the following is true:
```

felicitymay (dismissed):
@tvalenta - many thanks for raising an issue for this grammar bug and for your suggested fix ✨ 

It looks as if users will now see at least 3 bullet points in all versions of the documentation, so I think we can use a simpler fix. I'm going to commit my suggestion, and then we should be able to merge this PR and close the issue that you raised 🚀

felicitymay (approved):

content/actions/using-workflows/reusing-workflows.md
tvalenta: Hi @felicitymay . Your change is definitely more reasonable. For my own edification, is `private-actions` now always true? Is it possible for ghes or ghec or ghae to be true, but private-actions disabled?

felicitymay: Hi @tvalenta - `private-actions` is an example of a docs "feature flag" that we use to simplify conditions for the different versions of GitHub. Any content within a `{% ifversion private-actions %}` condition is shown for these versions of the product: https://github.com/github/docs/blob/main/data/features/private-actions.yml.

This feature flag is used to control the display of the final bullet. We have a situation where:

- Bullets 1 and 2 are shown for all versions of GitHub.
- Bullet 3 is shown only for enterprise versions of GitHub.
- Bullet 4 is shown for Free, Pro, Team plans, GitHub Enterprise Cloud, and GitHub Enterprise Server 3.8+ (as defined by the `private-actions` flag.

Consequently, all versions of this article will include at least three bullets. However, that wasn't true for Free, Pro, Team plans until bullet 4 was added in December. So it looks as if we overlooked this in that update.

If you're interested, you can find out more about use of conditions and feature flags in our docs here: [liquid helpers](https://github.com/github/docs-internal/blob/main/contributing/liquid-helpers.md) and [featues/README](https://github.com/github/docs-internal/tree/main/data/features#readme).

erdenerrallas19:
``

Austinlynn1986 (commented):
I see

PR #24045, opened 2023-02-21
fix grammar in content/actions/learn-github-actions/understanding-github-actions.md

xanderchu:
### Why:

fix grammar 

original sentence: 
> A job is a set of steps in a workflow that execute on the same runner.

In this case, `that` is referring to `a set of steps` and a set is singular.

Hence, we have two potential fixes:
1. "A job is a set of steps in a workflow that executes on the same runner."
2. "A job is a set of steps in a workflow that is executed on the same runner."

but later in the paragraph, we have this sentence:
> Steps are executed ....

It would be good to keep it consistent, therefore, option 2 is proposed and committed. 

### Check off the following:


cmwilson21:
@kerrychu Thanks so much for opening a PR! I'll get this triaged for review :zap:

Jackpoot:
Binance id 304141410

felicitymay (approved):
Thanks for fixing this bug and for clearly explaining the options and why you chose this fix. ✨ 

I'll approve this change and we'll get it merged as soon as possible.

PR #24087, opened 2023-02-22
Update encrypted-secrets.md

highfi13g:
special characters parsing issue comment.

### Why:

Closes ISSUE

### Check off the following:

### DIFF
--- a/content/actions/security-guides/encrypted-secrets.md
+++ b/content/actions/security-guides/encrypted-secrets.md
@@ -48,6 +48,8 @@ The following rules apply to secret names:
 
 To help ensure that {% data variables.product.prodname_dotcom %} redacts your secret in logs, avoid using structured data as the values of secrets. For example, avoid creating secrets that contain JSON or encoded Git blobs.
 
+Lastly, if your secret contains a special character such as `$` ensure you utilize an escaped character beforehand to avoid any parsing issues.
+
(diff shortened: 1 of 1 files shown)


AR0011:
``

cmwilson21:
👋 @highfi13g Thanks so much for opening a PR! I'll get this triaged for review :zap:

lucascosti:
Hi, @highfi13g, thanks for this PR.

IIUC, we don't recommend that users use escaping characters in secrets. Secrets are stored as-is and adding meta characters like backslashes would cause more problems when secrets are passed to actions, as well as and other uses.

We do have this guidance further down in the article, in the [Using encrypted secrets in a workflow](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) section:

> Avoid passing secrets between processes from the command line, whenever possible. Command-line processes may be visible to other users (using the ps command) or captured by [security audit events](https://docs.microsoft.com/windows-server/identity/ad-ds/manage/component-updates/command-line-process-auditing). To help protect secrets, consider using environment variables, STDIN, or other mechanisms supported by the target process.
>
> If you must pass secrets within a command line, then enclose them within the proper quoting rules. Secrets often contain special characters that may unintentionally affect your shell. To escape these special characters, use quoting with your environment variables. For example:

PR #24726, opened 2023-03-28
Mention that "Allow specified actors to bypass required pull requests" only applies to repo admins

### LINKED ISSUE #24725: The "Allow specified actors to bypass required pull requests" setting only applies to repo admins
### Code of Conduct

[Managing a branch protection rule][1]

[1]: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule

The "Allow specified actors to bypass required pull requests" bullet

### Additional information

_No response_

dhermes:
### Why:

Closes: #24725

Phrasing a user reference as "repository admin users" instead of "actors"

### Check off the following:


dhermes:
I'm just trying to start a conversation and figured this PR would be a good place to start.

From **looking at** the feature, it appears the goal is for it to work for non-admin users. So maybe it's actually a bug in GitHub. Some potential bugs with the product that might be issued:

- (A) Non-admin users **should** be able to "bypass required pull requests"
- (B) Non-admin users **should not** be able to be added to the "Allow specified actors to bypass required pull requests" list
- (B) There should be a warning if an (invalid) non-admin users is added to / present in the "Allow specified actors to bypass required pull requests" list

cmwilson21:
@dhermes Thank you for opening a PR and linking it to your issue! ✨ 

We will take a look at the docs update, but if you have any suggestions or feedback for the feature, that conversation would best fit over at [community discussions ](https://github.com/orgs/community/discussions). 💖

elee-msft:
I would like to emphasize that the "Allow specified actors to bypass required pull requests" feature *appears* to be intended to grant bypassing privileges to accounts that aren't otherwise admins in the repo, particularly for apps like the Azure Pipelines app, which is the scenario I need and what I'm currently blocked on.  If the purpose was simply to grant bypass privileges to admins, they already have that by default unless the "Do not allow bypassing the above settings" setting is turned on.  So someone needs to check with the product team about what the design intent actually is.  :-)

isaacmbrown:
Hi @dhermes, thanks for the issue and linked PR!

Could you share more context about the issue you're experiencing and how we'd reproduce it?

From a quick test, I was able to add a user with just `write` access as an actor with bypass permissions. As I'd expect, I wasn't able to add a user with just `read` access.

Is the specific issue here:
- You've found you can't add users with just `write` access (this sounds like a bug if so)
- You are trying to add a Team or App, but you can't

dhermes:
> Could you share more context about the issue you're experiencing and how we'd reproduce it?

@isaacmbrown 

## Repo settings

Add a user to a repo with "Write" permissions (crucial: this user is not an org admin)

Screenshot 2023-05-25 at 11 55 40 AM

## Branch protection rule

Add that user to the "Allow specified actors to bypass required pull requests" list

Screenshot 2023-05-25 at 11 54 28 AM

## PR experience (from perspective of user)

The user **cannot** bypass on a given PR, for example:

Screenshot 2023-05-25 at 11 51 26 AM

This should say

> Merge without waiting for requirements to be met (bypass branch protections)

Screenshot 2023-05-25 at 12 01 19 PM

isaacmbrown:
Hi @dhermes, I will look closer, but the restriction there looks to be to do with the required code owner review? 

I believe selecting the bypass option should allow the selected user to push to the protected branch without creating a pull request at all. I don't believe it's meant to exempt them from the rest of the rules once a pull request is created.

isaacmbrown:
I got confirmation on this: https://github.com/github/docs/pull/24726#issuecomment-1563251229

@elee-msft, was the problem you were having the same thing `@dhermes` mentioned above? Or were you having a different issue that would require a docs update?

elee-msft:
@isaacmbrown My issue was a little different.  I'm trying to configure things so that the GitHub “Azure Pipelines” app can push directly to main, bypassing the branch rules that would normally forbid that.  (The pipeline is attempting to sync code between an ADO repo and a GitHub repo).  I [opened a feedback item on this](https://developercommunity.visualstudio.com/t/How-to-configure-permissions-for-GitHub-/10343826) and they pointed me to this docs issue, saying that the bypass option is applied to the “repository admin users” instead of “actors”.  That doesn't make sense to me from the plain reading of the UI, but if that's indeed supposed to be the desired design, then the docs should be updated to clarify that.  But I still think the current behavior is actually a bug and I'm hoping someone can confirm that with the product team.

isaacmbrown:
Hi @elee-msft, thanks for getting back to me! I got confirmation from product that the feature is intended to work as described. Therefore I'd suggest contacting GitHub support -- but first, just want to check one thing with you:

In the thread you linked, you listed your branch protection rules:

>a) Require a pull request before merging
b) Require approvals
c) Require review from Code Owners
d) Allow specified actors to bypass required pull requests, with the Azure Pipeline app selected (screenshotted in my previous reply)
e) Require status checks to pass before merging
f) Do not allow bypassing the above settings
g) Restrict who can push to matching branches

Is the Azure Pipeline app also given permission to push to the protected branch in `g)`? If not, that rule would also be preventing the app from pushing to the protected branch, irrespective of the pull request rules.

isaacmbrown:
Based on the discussion above I don't think a docs update is required here, so I'm closing this PR.

PR #24844, opened 2023-04-03
provide hint about other `gh` commands

joee:
to use after `gh auth login`

### Why:

Closes: 

### Check off the following:


cmwilson21:
@joee Thanks so much for opening a PR! I'll get this triaged for review :zap:

isaacmbrown:
Hi @joee, thanks for raising the PR. I'm not sure this addition is needed here, because the article is specifically about methods for caching your credentials. Authenticating to the CLI does allow you to use other CLI commands, but it also stores your credentials so you can, for example, access your private repos with a standard `git clone`. I worry the addition would take the focus away from the purpose of the article, and imply this method is only caching your credentials for other `gh cli` operations.

I'll close this PR, but feel free to come back if I've missed something or you'd like to make a new suggestion.

PR #24896, opened 2023-04-05
Specify scope of default environment variables

### LINKED ISSUE #24895: default environment variables - where can they be used?
### Code of Conduct

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

the section on default environment variables. it says

> The default environment variables that GitHub sets are available to every step in a workflow.

but when I try for instance to use such a variable as input to a step using a composite action, as in

```
      - name: Fetch job html
        uses: ./.github/actions/job_html_url
        with:
          job_name: ${{ env.GITHUB_JOB }}
```

the passed input is the empty string.

I would like to see more specifically where in a step these env vars can be used, as i would expect that they can be accessed using the `env` context in expressions, but apparently this is not the case. not even a step with
```
        run: echo ${{ env.GITHUB_JOB }}
```
works! it seems they are not accessible through the env context at all :thinking: 

### Additional information

_No response_

valentin-krasontovitsch:
It is not obvious that default environment variables will not be available in the `env` context. We add a note to that regard.

### Why:

Closes: #24895 

### Check off the following:

### DIFF
--- a/content/actions/learn-github-actions/variables.md
+++ b/content/actions/learn-github-actions/variables.md
@@ -274,6 +274,8 @@ Configuration variables can be accessed across the workflow using `vars` context
 
 The default environment variables that {% data variables.product.prodname_dotcom %} sets are available to every step in a workflow.
 
+Because default environment variables are set by GitHub and not defined in a workflow, they are not accessible through the env context. However, most of the default variables have a corresponding, and similarly named, context property. For example, the value of the GITHUB_REF variable can be read during workflow processing using the ${{ github.ref }} context property.
+
 {% data reusables.actions.environment-variables-are-fixed %} For more information about setting environment variables, see "[Defining environment variables for a single workflow](#defining-environment-variables-for-a-single-workflow)" and "[AUTOTITLE](/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable)."
 
 We strongly recommend that actions use variables to access the filesystem rather than using hardcoded file paths. {% data variables.product.prodname_dotcom %} sets variables for actions to use in all runner environments.
@@ -341,7 +343,6 @@ We strongly recommend that actions use variables to access the filesystem rather
 **Note:**
 
 - If you need to use a workflow run's URL from within a job, you can combine these variables: `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`
-- Most of the default variables have a corresponding, and similarly named, context property. For example, the value of the `GITHUB_REF` variable can be read during workflow processing using the {% raw %}`${{ github.ref }}`{% endraw %} context property.
 
 {% endnote %}
 
--- a/content/actions/learn-github-actions/variables.md
+++ b/content/actions/learn-github-actions/variables.md
@@ -340,9 +340,7 @@ We strongly recommend that actions use variables to access the filesystem rather
 
 {% note %}
 
-**Note:**
-
-- If you need to use a workflow run's URL from within a job, you can combine these variables: `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`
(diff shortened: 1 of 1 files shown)


cmwilson21:
@valentin-krasontovitsch Thank you for opening a PR and linking it to your issue! ✨ 

valentin-krasontovitsch:
four months for a simple, small docs update. I have my own project that I'm woefully neglecting, so I understand this can happen, but that one has just one maintainer, and is free-time based :sweat_smile:

cmwilson21:
@valentin-krasontovitsch Thanks for checking in on this. It is up on the board for review 👍 

We appreciate your patience as we are working through our backlog 💖

valentin-krasontovitsch:
Of course, thank you for your response, and godspeed :rainbow:

SiaraMist:
Hey @valentin-krasontovitsch, thank you so much for your patience with us as we work through our backlog. 🙇‍♀️ 

I do see how it would be confusing for users to know how to use default environment variables in workflows with the way the docs are currently written. There is [a note](https://github.com/github/docs/pull/24896/files#diff-e0e008dd4d4b5d933f0ed768ad05251d30e5da99be5b5a1af0f33e72aef52d51R344) underneath the table of default environment variables that says:

> Most of the default variables have a corresponding, and similarly named, context property. For example, the value of the `GITHUB_REF` variable can be read during workflow processing using the `${{ github.ref }}` context property.

What do you think of this idea:

We could move that out of the note and into its own paragraph after the first sentence in the introduction. We could also add a sentence before it so that we say something along the lines of:

> Because default environment variables are set by GitHub and not defined in a workflow, they are not accessed through the `env` context. However, most of the default variables have a corresponding, and similarly named, context property. For example, the value of the `GITHUB_REF` variable can be read during workflow processing using the `${{ github.ref }}` context property.

Would that make things more clear in this doc? Let me know! Thank you for helping to improve GitHub's docs! ⚡

valentin-krasontovitsch:
yes i think that would definitely help! thanks for the much better suggestion : )

should i change this PR to reflect your suggestions, or close this one and leave it to you / somebody else to create a new PR?

SiaraMist:
That's awesome. 😃 @valentin-krasontovitsch if you want to update this PR, then I can accept your changes and get them merged!

valentin-krasontovitsch:
it is done ^^

let me know if the new diff is OK, and if the commit message is up to par.

content/actions/learn-github-actions/variables.md
```diff
@@ -341,7 +343,6 @@ We strongly recommend that actions use variables to access the filesystem rather
 **Note:**
 
 - If you need to use a workflow run's URL from within a job, you can combine these variables: `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`
```
SiaraMist, line 345:
```suggestion
**Note:** If you need to use a workflow run's URL from within a job, you can combine these variables: `$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID`
```

content/actions/learn-github-actions/variables.md
```diff
@@ -274,6 +274,8 @@ Configuration variables can be accessed across the workflow using `vars` context
 
 The default environment variables that {% data variables.product.prodname_dotcom %} sets are available to every step in a workflow.
 
+Because default environment variables are set by GitHub and not defined in a workflow, they are not accessible through the env context. However, most of the default variables have a corresponding, and similarly named, context property. For example, the value of the GITHUB_REF variable can be read during workflow processing using the ${{ github.ref }} context property.
```
SiaraMist, line 277:
```suggestion
Because default environment variables are set by {% data variables.product.prodname_dotcom %} and not defined in a workflow, they are not accessible through the `env` context. However, most of the default variables have a corresponding, and similarly named, context property. For example, the value of the `GITHUB_REF` variable can be read during workflow processing using the {% raw %}`${{ github.ref }}`{% endraw %} context property.
```
Just adding some formatting!

SiaraMist (commented):
Your updates look great, as does your commit message! ⚡ I just added some docs formatting, and if it lets me commit my suggestions I'll merge this now.

SiaraMist (approved):

PR #25031, opened 2023-04-15
Describe that people receive an email when converted to outside collaborator.

trusktr:
### Why:

So people know the consequences of their actions. I'm not sure if an email is sent or not (please edit), but this should be clear. Some actions do send the target user an email, and we should document this so that people targeting actions on other users know what to expect.

### Check off the following:


cmwilson21:
@trusktr Thanks so much for opening a PR! I'll get this triaged for review :zap:

isaacmbrown:
Hi @trusktr, I just tested and it doesn't look like this action sends an email notification. 

To me it feels strange to need to specify when an action _won't_ send an email. I also think, if we were going down that route, we would need to update more places than this one specific example. 

> Some actions do send the target user an email, and we should document this so that people targeting actions on other users know what to expect.

If you feel strongly about this, can I recommend creating an issue so we can discuss and work from there? I'm going to close this PR, since this specific change isn't one we'll publish.

atifnimran (commented):
U

PR #25440, opened 2023-05-07
Fix workflow-syntax-for-github-actions.md

hrishikesh-kadam:
Correct the header level of `jobs.<job_id>.steps[*].run` so that it can be seen in the sidebar.


cmwilson21:
@hrishikesh-kadam Thanks so much for opening a PR! I'll get this triaged for review :zap:

skedwards88 (approved):
Thanks! We'll get this merged down for you.

PR #25519, opened 2023-05-12
Set permissions at the job level rather than the top-level

jeffwidman:
Best practice is to limit scopes, so here we limit the increased permissions to this particular job, not to the entire workflow.

Setting a good example here keeps folks from later copy/pasting bad practices to their own workflow files...

### Why:

Closes: 

### Check off the following:

- [ ] I have reviewed my changes in staging, available via the **View deployment** link in this PR's timeline.


data/reusables/actions/github_token-input-example.md
```diff
@@ -4,13 +4,14 @@ This example workflow uses the [labeler action](https://github.com/actions/label
...
 on: [ pull_request_target ]
 
-permissions:
-  contents: read
-  pull-requests: write
-
 jobs:
   triage:
     runs-on: ubuntu-latest
+
```
jeffwidman, line 10: I wasn't sure whether to include whitespace here...

1. I find the whitespace makes the example more readable, especially since it's trying to highlight the `permissions` stanza.
2. But it's inconsistent with the other example on one of the pages where this is used: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

Personally, I'd resolve this by adding whitespace to 2️⃣ but I'm fine either way, I defer to the expertise of the docs team, so do whatever is best and merge down.

cmwilson21:
👋 @jeffwidman Thanks so much for opening a PR! I'll get this triaged for review :zap:

SiaraMist (dismissed):
Hey @jeffwidman, thank you so much for updating this example to use the best practice to limit scopes! ✨ 

Re: the whitespace, it looks like all of our other examples only add an empty line between top level `keys`, so I'll remove it from this one for consistency. If it was just that one other example I'd update them both, but since it's all of our examples I think that's a bit large in scope. 😅 

Thank you again for opening this PR! I'll go ahead and get it merged.

SiaraMist (approved):

oysteinjaren:
@jeffwidman @SiaraMist I just came across this example being used here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input

A [paragraph](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) a bit further down refers to this example like this:

>The two workflow examples earlier in this article show the permissions key being used at the workflow level, and at the job level. In [Example 1](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input) the two permissions are specified for the entire workflow. In [Example 2](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#example-2-calling-the-rest-api) write access is granted for one scope for a single job.

This description is no longer true because of this change. Both examples now specify permissions at the job level.

This particular reusable example is currently used in two places. The other one is at https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-secrets-context. I agree it is best to show permissions set at the job level there, since that is arguably best practice from a least-privilage perspective. Maybe a separate example should be created instead of reusing the same one?

cmwilson21:
@oysteinjaren Thanks for pointing this out! ✨ 

Would you mind opening a [new issue](https://github.com/github/docs/issues/new/choose) with this information? That way we can get it triaged and put it up for review. 💖

PR #25628, opened 2023-05-18
Fix theme mismatch

### LINKED ISSUE #2229: Components do not update with programmatic changes in theme
**Describe the bug**

We use `<ActionMenu>` in a couple of places on docs.github.com and they get the wrong theme applied to them. 

In this screenshot you can see it twice. Once inside the modal (upper right-hand) and behind the modal too, where it has an action menu for "TYPE" and "TOPIC".

Screen Shot 2022-08-15 at 9 34 21 AM

On docs.github.com we use `auto` as the default theme on the server-side generated HTML. Then, after the page has loaded we use `useEffect` to correct the theme preference based on a cookie (`Cookies.get('color_mode')`). 

To reproduce it I think you need an OS that prefers dark, but [override your settings on github.com](https://github.com/settings/appearance) to light.

Visit: https://docs.github.com/en/issues/guides

Incidentally, when I test in my Firefox (where I currently don't have a cookie that prefers light mode) this rather odd behavior happens (when I refresh and refresh and refresh the page):

https://user-images.githubusercontent.com/26739/184646968-a77715df-d395-491a-8e9e-7857d212f2d7.mov

The two ActionMenu buttons this time starts out in light mode and then later become dark. 

**To Reproduce**
1. Set your OS to prefer dark
2. Go to https://github.com/settings/appearance and select that you prefer one of the light themes
3. Visit https://docs.github.com/en/issues/guides

**Expected behavior**

That the "TYPE" and "TOPIC" buttons appear in light mode. Like they do on the documentation:

Screen Shot 2022-08-15 at 9 46 05 AM

**Screenshots**

See above.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Chrome
 - Version latest stable

**Additional context**

On docs.github.com we use:
```
    "@primer/css": "^20.2.4",
    "@primer/octicons": "17.3.0",
    "@primer/octicons-react": "17.3.0",
    "@primer/react": "^35.2.2",
```
(from our [`package.json`](https://github.com/github/docs/blob/3314f8e6e4860c395d0ee3fd6e5a38327f66d75d/package.json#L14-L17))

colebemis:
### Why:

Fixes a theme mismatch issue that occurs when the OS color mode doesn't match the user's GitHub color mode (e.g., OS: `dark`, GitHub: `light`).

| Before | After |
| --- | --- |
| CleanShot 2023-05-18 at 16 34 36 | CleanShot 2023-05-18 at 16 35 01 |

Closes: https://github.com/primer/react/issues/2229

This bug was the result of a timing issue. After a page loads, the docs site switches the color mode to match the user's GitHub color mode. However, Primer React has a `useEffect` call that overrides this change, causing the site to ignore the user's GitHub color mode and revert to `auto`. 

As a temporary workaround, I've wrapped the code that fetches the user's GitHub color mode in a `setTimeout()` call to ensure that it's called after Primer React's `useEffect` call. The long term solution to this theming issue is to migrate to CSS variables under the hood, which Primer is planning to do in the next couple quarters.

### Check off the following:

- [ ] ~~For content changes, I have completed the [self-review checklist](https://github.com/github/docs/blob/main/contributing/self-review.md#self-review).~~

### DIFF
--- a/components/hooks/useTheme.ts
+++ b/components/hooks/useTheme.ts
@@ -105,14 +105,14 @@ export function useTheme() {
   })
 
   useEffect(() => {
-    // Using setTimeout with a default delay value of 0 interjects one 
-    // additional event cycle, which works around a bug that is the 
+    // Using setTimeout with a default delay value of 0 interjects one
+    // additional event cycle, which works around a bug that is the
     // result of a timing issue. Without the setTimeout function
-    // the page loads, then the docs site switches the color mode to 
-    // match the user's GitHub color mode. Primer React has a useEffect 
+    // the page loads, then the docs site switches the color mode to
+    // match the user's GitHub color mode. Primer React has a useEffect
     // call that overrides this change, causing the site to ignore the
     // user's GitHub color mode and revert to auto.
-    // As a temporary workaround, this code that fetches the user's GitHub 
+    // As a temporary workaround, this code that fetches the user's GitHub
     // color mode will be called after Primer React's useEffect call.
     // The long term solution to this theming issue is to migrate to CSS variables
     // under the hood, which Primer is planning to do in the next couple quarters.
--- a/components/hooks/useTheme.ts
+++ b/components/hooks/useTheme.ts
@@ -116,6 +116,7 @@ export function useTheme() {
     // color mode will be called after Primer React's useEffect call.
     // The long term solution to this theming issue is to migrate to CSS variables
     // under the hood, which Primer is planning to do in the next couple quarters.
+    // Reference: https://github.com/primer/react/issues/2229
     setTimeout(() => {
       const cookieValue = Cookies.get('color_mode')
       const css = getCssTheme(cookieValue)
--- a/components/hooks/useTheme.ts
+++ b/components/hooks/useTheme.ts
@@ -105,6 +105,17 @@ export function useTheme() {
   })
 
   useEffect(() => {
+    // Using setTimeout with a default delay value of 0 interjects one 
+    // additional event cycle, which works around a bug that is the 
+    // result of a timing issue. Without the setTimeout function
+    // the page loads, then the docs site switches the color mode to 
+    // match the user's GitHub color mode. Primer React has a useEffect 
+    // call that overrides this change, causing the site to ignore the
+    // user's GitHub color mode and revert to auto.
+    // As a temporary workaround, this code that fetches the user's GitHub 
+    // color mode will be called after Primer React's useEffect call.
+    // The long term solution to this theming issue is to migrate to CSS variables
+    // under the hood, which Primer is planning to do in the next couple quarters.
     setTimeout(() => {
       const cookieValue = Cookies.get('color_mode')
       const css = getCssTheme(cookieValue)
--- a/components/hooks/useTheme.ts
+++ b/components/hooks/useTheme.ts
@@ -105,10 +105,12 @@ export function useTheme() {
   })
 
   useEffect(() => {
-    const cookieValue = Cookies.get('color_mode')
-    const css = getCssTheme(cookieValue)
-    const component = getComponentTheme(cookieValue)
-    setTheme({ css, component })
+    setTimeout(() => {
+      const cookieValue = Cookies.get('color_mode')
+      const css = getCssTheme(cookieValue)
+      const component = getComponentTheme(cookieValue)
+      setTheme({ css, component })
+    })
   }, [])
 
   return { theme }


cmwilson21:
👋 @colebemis Thanks for the PR! 🎉 

I see this is a draft PR, but I'll go ahead and get this triaged. Feel free to ping me when it's ready for review! ✨

colebemis:
👋 Hi, @cmwilson21! This PR is ready for review now

rachmari (dismissed):
This looks great @colebemis! Thank you for the fix. ✨ 

To note, I modified my GitHub appearance preference to light mode, and my OS is already dark mode. I added a color_mode cookie to the preview site spun up by this PR and verified that the components were rendering correctly with the change.

peterbe:
@colebemis Thanks!!
What would be great is if that hack was accompanied with an explanation. Sooner or later someone's going to see that code and (fail to do a git blame lookup back to the (your) PR) wonder; what does this do?

peterbe:
Also, is it fair to say that this *closes* that bug in Primer? Seems like the bug is still present in a sense.

components/hooks/useTheme.ts
```diff
@@ -105,10 +105,12 @@ export function useTheme() {
   })
 
   useEffect(() => {
-    const cookieValue = Cookies.get('color_mode')
-    const css = getCssTheme(cookieValue)
-    const component = getComponentTheme(cookieValue)
-    setTheme({ css, component })
+    setTimeout(() => {
```
rachmari, line 108: @peterbe suggested adding a comment to summarize this workaround. @colebemis @peterbe here's a suggestion for a comment we can include. Feel free to suggest alternatives or edits.
```suggestion
    // Using setTimeout with a default delay value of 0 interjects one 
    // additional event cycle, which works around a bug that is the 
    // result of a timing issue. Without the setTimeout function
    // the page loads, then the docs site switches the color mode to 
    // match the user's GitHub color mode. Primer React has a useEffect 
    // call that overrides this change, causing the site to ignore the
    // user's GitHub color mode and revert to auto.
    // As a temporary workaround, this code that fetches the user's GitHub 
    // color mode will be called after Primer React's useEffect call.
    // The long term solution to this theming issue is to migrate to CSS variables
    // under the hood, which Primer is planning to do in the next couple quarters.
    setTimeout(() => {
```

peterbe (dismissed):
Tested locally and it solves the problem. Wo'hoo!

It's a bad sad that we collectively don't understand it better and that it's even necessary. 
I'm still not sure if it's fair to say this PR here closes that @primer/react bug, technically. But I'm also not sure if I care enough :) It definitely solves the bug for us. 

If you could think of a good code comment to inject just before the `setTimeout(...` that would be great. Perhaps a mention of the URL https://github.com/primer/react/issues/2229 (it's a public repo after all)

mudanzasIPROM:
aaa

mudanzasIPROM:
aaa

components/hooks/useTheme.ts
colebemis: Looks great. Thanks for adding this @rachmari!

components/hooks/useTheme.ts
```diff
@@ -105,10 +105,23 @@ export function useTheme() {
...
+    // additional event cycle, which works around a bug that is the 
+    // result of a timing issue. Without the setTimeout function
+    // the page loads, then the docs site switches the color mode to 
+    // match the user's GitHub color mode. Primer React has a useEffect 
+    // call that overrides this change, causing the site to ignore the
+    // user's GitHub color mode and revert to auto.
+    // As a temporary workaround, this code that fetches the user's GitHub 
+    // color mode will be called after Primer React's useEffect call.
+    // The long term solution to this theming issue is to migrate to CSS variables
+    // under the hood, which Primer is planning to do in the next couple quarters.
```
colebemis, line 118:
```suggestion
    // under the hood, which Primer is planning to do in the next couple quarters.
    // Reference: https://github.com/primer/react/issues/2229
```

colebemis:
> I'm still not sure if it's fair to say this PR here closes that @primer/react bug, technically. But I'm also not sure if I care enough :) It definitely solves the bug for us.

@peterbe Good point. I'm comfortable closing the primer/react issue since we're actively rewriting our styling architecture and theming system (https://github.com/github/primer/issues/1626). ThemeProvider will be completely rewritten to use CSS variables as part of that work.

heiskr (dismissed):
I vaguely remember proposing something similar at one point lol, but lost track of it. Looks great to me :)

peterbe (commented):
Would you mind running 
`./node_modules/.bin/prettier --write components/hooks/useTheme.ts`
on this branch. I think when Rachael suggested the code comment in the web UI, it injects excess whitespace at the end of some of the lines.

colebemis:
@peterbe Good catch. Done 👍

peterbe (approved):

peterbe:
FOR THE RECORD. 

Been chatting to @colebemis and the docs-engineering team in Slack. We're eager to see this land but **we won't merge it till after May 24th**. This is to be conservative and avoid any risk of causing any reliability issues on Docs when the monolith is observing a frozen tree.

str13tlife (commented):
Not approved.

str13tlife:
This is israel meza owner of this phone that is beung reconfigured without
my authorization. My account are hacked, so im kindly asking you and github
to return everything like it was. I am not authorizing any changes,please
contact me A.S.A.P AT MY NUMBER 686 161 1449 MEXI O

PR #26137, opened 2023-06-21
Avoid shell injection in examples

soliton-:
### Why:

Avoid shell injection in examples to follow suggestions from https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections.

Changed examples that used shell injection. Removed a paragraph that explained the injection poorly.
Clarified multi-line string syntax warning and revert to simple delimiter use in the example.

### Check off the following:


cmwilson21:
@soliton- Good to see another PR from you! 👍 

Thanks for the submission. I'll get it triaged for review. ⚡

cmwilson21:
Quick question as I'm triaging it, you mentioned removing a paragraph - do the updates you've made render that paragraph unnecessary? Or do you have a suggestion for improving the paragraph instead of removing it altogether?

soliton-:
@cmwilson21 Yes, the paragraph was about the shell injection that gets removed in this PR.

cmwilson21:
@soliton- I follow now, thank you for confirming!

soliton-:
Is there any update on this? Anything I can do?

cmwilson21:
@soliton- Thanks for checking in! It is up on the board for review. We appreciate your patience as we are working through our backlog 💛

content/actions/using-workflows/workflow-commands-for-github-actions.md
```diff
@@ -839,13 +833,13 @@ For multiline strings, you may use a delimiter with the following syntax.
 
 {% warning %}
 
-**Warning:** Make sure the delimiter you're using is randomly generated and unique for each run. For more information, see "[AUTOTITLE](/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections)".
+**Warning:** Make sure the delimiter you're using cannot occur on its own line in the value. If the value is completely arbitrary then don't use this format at all. Write to a separate file instead.
```
hubwriter, line 836: I think it would be clearer to say:
```suggestion
**Warning:** Make sure the delimiter you're using won't occur on a line of its own within the value. If the value is completely arbitrary then you shouldn't use this format. Write the value to a file instead.
```

content/actions/using-workflows/workflow-commands-for-github-actions.md
```diff
@@ -727,8 +727,6 @@ console.log("The running PID from the main action is: " +  process.env.STATE_pro
 
 During the execution of a workflow, the runner generates temporary files that can be used to perform certain actions. The path to these files are exposed via environment variables. You will need to use UTF-8 encoding when writing to these files to ensure proper processing of the commands. Multiple commands can be written to the same file, separated by newlines.
 
-Most commands in the following examples use double quotes for echoing strings, which will attempt to interpolate characters like `$` for shell variable names. To always use literal values in quoted strings, you can use single quotes instead.
```
hubwriter, line 730: Why have you removed the above? The seems to still apply.

hubwriter (changes requested):
@soliton- Many thanks for submitting this PR. 👍 

I have one suggestion for a change, and one question.

Thanks again for taking the time to improve the docs. Much appreciated.

content/actions/using-workflows/workflow-commands-for-github-actions.md
hubwriter: Thanks for the clarification. Let's take this out, as you suggest.

hubwriter (approved):

hubwriter:
@soliton- Thanks again.
We've got a merge freeze just now due to an upcoming release, but when that's done we'll get this merged and published.
🚀

PR #26698, opened 2023-07-14
'go vet' as part of Go starter workflow

SchulteDev:
### Why:

What is 'go vet'?[1]
'go vet' should be part of most Golang projects when building. Because of this, adding the command to the 'Go starter workflow' hoping more projects will include it in their pipeline.
 
[1] "https://pkg.go.dev/cmd/vet" Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string. Vet uses heuristics that do not guarantee all reports are genuine problems, but it can find errors not caught by the compilers.

### What's being changed
- 'go vet' command added as part of Go starter workflow

### Check off the following:

### DIFF
--- a/content/actions/automating-builds-and-tests/building-and-testing-go.md
+++ b/content/actions/automating-builds-and-tests/building-and-testing-go.md
@@ -55,6 +55,9 @@ jobs:
       - name: Build
         run: go build -v ./...
 
+      - name: Vet
+        run: go vet -v ./...
+
       - name: Test
         run: go test -v ./...
 ```


cmwilson21:
@SchulteMarkus Thanks so much for submitting a PR! I'll get this triaged for review :zap:

Facebookuz:
```python

``` Apteka

Facebookuz (commented):
Ad new workflows

hubwriter (changes requested):
@SchulteMarkus - Thanks for submitting this docs PR.

However, the code sample you are suggesting changing is intended to be the Go starter workflow (linked to from the article: https://github.com/actions/starter-workflows/blob/main/ci/go.yml).

I suggest you raise a PR against that repo to suggest a change there and then, if that's accepted, we can change the docs here. There are a couple of other little updates we need to make to keep the code sample up to date with that workflow, and we can make those changes at the same time.

Thanks again. 👍

PR #27277, opened 2023-08-06
docs(contexts): update inaccurate description of github.job

lillapetri:
### Why: Documentation is misleading, `github.job` actually gives the job's name and not its ID.

As a side note: [the community has expressed a strong demand](https://github.com/orgs/community/discussions/8945) for making the job's ID available in the context, which would be more useful than fetching its name.

Closes:  -

### Check off the following:

### DIFF
--- a/content/actions/learn-github-actions/contexts.md
+++ b/content/actions/learn-github-actions/contexts.md
@@ -178,7 +178,7 @@ The `github` context contains information about the workflow run and the event t
 | `github.event_path` | `string` | The path to the file on the runner that contains the full event webhook payload. |
 | `github.graphql_url` | `string` | The URL of the {% data variables.product.prodname_dotcom %} GraphQL API. |
 | `github.head_ref` | `string` | The `head_ref` or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either `pull_request` or `pull_request_target`. |
-| `github.job` | `string` | The [`job_id`](/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id) of the current job. <br /> Note: This context property is set by the Actions runner, and is only available within the execution `steps` of a job. Otherwise, the value of this property will be `null`. |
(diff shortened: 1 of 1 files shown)


SiaraMist:
Hi @lillapetri, thank you so much for opening this PR! ✨ 

Apologies if I'm misunderstanding the intent of your updates, but it looks like the current documentation is correct. the `github.jobs` context returns the `job_id`, and not the `job_name`. In [this example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-creating-jobs) in the docs, it shows that the `job_id` is `my_first_job`, and the `job_name` is "My first job". If someone were to use the `github.jobs` context in that example, the value would be the `job_id`, which is `my_first_job`, instead of the job name that shows in the UI. 

I'm going to close this PR in favor of keeping the current documentation since that's the behavior I'm seeing in workflow runs, but if I'm misunderstanding anything, please feel free to reopen this! Thank you again for contributing to GitHub Docs. 💛

lillapetri:
Dear @SiaraMist,

I guess the root of the misunderstanding here is the ambiguous meaning of the `job_id`, as I incorrectly assumed that the ID mentioned in the [API documentation](https://docs.github.com/en/rest/actions/workflow-jobs?apiVersion=2022-11-28#get-a-job-for-a-workflow-run) (which is an integer) and the ID mentioned in the context documentation (which is the string key) are the same thing, since they're both referred to as `job_id`. 
It would be extremely useful if there was a way to retrieve the value of the integer ID of the job.
To clarify that the job has two distinct properties called `job_id`, may I suggest the use of "string ID" and "integer ID"?

SiaraMist:
That's a good callout! While looking into your suggestion, I noticed [our table on this page](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) does list the data type. `github.job` has its "Type" listed as `string`, for example. In the [Workflow syntax article](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id), it looks like `job_id` is listed as a string value too. Is this what you’re looking for or are you asking about something else?

PR #27583, opened 2023-08-19
[GitHub Actions] Fail fast from the default shells (bash/sh) in Linux/macOS

### LINKED ISSUE #23853: [GitHub Actions] Fail fast from the default shells (bash/sh) in Linux/macOS
### Code of Conduct

The [`jobs.<job_id>.steps[*].shell`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell) lists the default shell commands for **Linux/macOS** under **Command run internally** column:

| shell             | command                                              |
|:-----------------:|:---------------------------------------------------|
| unspecified  | `bash -e {0}`                                           |
| `bash`          | `bash --noprofile --norc -eo pipefail {0}` |
| `sh`              | `sh -e {0}`                                               |

Screenshot of a test run from `ubuntu-latest`/`ubuntu-22.04` runner:

image

All these variants have `-e` and according to its description under  [The Set Builtin](https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin) section form the [Bash manual](https://www.gnu.org/software/bash/manual/bash.html):

> `-e`: Exit immediately if a pipeline (see [Pipelines](https://www.gnu.org/software/bash/manual/bash.html#Pipelines)), which may consist of a single simple command (see [Simple Commands](https://www.gnu.org/software/bash/manual/bash.html#Simple-Commands)), a list (see [Lists of Commands](https://www.gnu.org/software/bash/manual/bash.html#Lists)), or a compound command (see [Compound Commands](https://www.gnu.org/software/bash/manual/bash.html#Compound-Commands)) returns a non-zero status.

Which is equivalent to the [fail fast](https://en.wikipedia.org/wiki/Fail-fast) behavior.

But, according to the [Exit codes and error action preference](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference):

> `bash`/`sh`: Fail-fast behavior using `set -eo pipefail`: This option is set when `shell: bash` is _explicitly specified_. **It is not applied by default.**

And, that is not the case. This only applies to the `-o pipefail` part (Bash only) because `-e` is applied by default.

The `-o pipeline` only affects the returning of the exit status from [Pipelines](https://www.gnu.org/software/bash/manual/bash.html#Pipelines) in Bash:

> `pipefail`: If set, the return value of a pipeline is the value of the last (rightmost) command to exit with a non-zero status, or zero if all commands in the pipeline exit successfully.

The docs should be revised to avoid contradiction when both these sections are read in conjunction.

- `-e` is applied by default to `sh` and `bash` both.
- `-o pipefail` is applied to `bash` only (i.e. `bash -e -o pipefail`) when `shell: bash` is specified; and, not to `shell: sh`.
- ~~`sh` command should also have `pipefail` but at the moment it doesn't. Also, it has a different syntax i.e. `sh -e -pipefail`.~~

iamazeem:
### Why:

Closes: #23853

- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

### Check off the following:

- [ ] I have reviewed my changes in staging, available via the **View deployment** link in this PR's timeline.

### DIFF
--- a/content/actions/using-workflows/workflow-syntax-for-github-actions.md
+++ b/content/actions/using-workflows/workflow-syntax-for-github-actions.md
@@ -712,7 +712,7 @@ For information about the software included on GitHub-hosted runners, see "[AUTO
 For built-in shell keywords, we provide the following defaults that are executed by {% data variables.product.prodname_dotcom %}-hosted runners. You should use these guidelines when running shell scripts.
 
 - `bash`/`sh`:
-  - By default, fail-fast behavior is enfored using `set -e` for `sh` and `bash` both. When `shell: bash` is specified, `-o pipefail` is also appied to enforce early exit from pipelines in case of non-zero exit status.
+  - By default, fail-fast behavior is enforced using `set -e` for `sh` and `bash` both. When `shell: bash` is specified, `-o pipefail` is also applied to enforce early exit from pipelines in case of non-zero exit status.
   - You can take full control over shell parameters by providing a template string to the shell options. For example, `bash {0}`.
   - `sh`-like shells exit with the exit code of the last command executed in a script, which is also the default behavior for actions. The runner will report the status of the step as fail/succeed based on this exit code.
 
--- a/content/actions/using-workflows/workflow-syntax-for-github-actions.md
+++ b/content/actions/using-workflows/workflow-syntax-for-github-actions.md
@@ -687,7 +687,7 @@ steps:
 
 ### Custom shell
 
-You can set the `shell` value to a template string using `command […options] {0} [..more_options]`. {% data variables.product.prodname_dotcom %} interprets the first whitespace-delimited word of the string as the command, and inserts the file name for the temporary script at `{0}`.
+You can set the `shell` value to a template string using `command [options] {0} [more_options]`. {% data variables.product.prodname_dotcom %} interprets the first whitespace-delimited word of the string as the command, and inserts the file name for the temporary script at `{0}`.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


content/actions/using-workflows/workflow-syntax-for-github-actions.md
```diff
@@ -712,9 +712,9 @@ For information about the software included on GitHub-hosted runners, see "[AUTO
 For built-in shell keywords, we provide the following defaults that are executed by {% data variables.product.prodname_dotcom %}-hosted runners. You should use these guidelines when running shell scripts.
 
 - `bash`/`sh`:
-  - Fail-fast behavior using `set -eo pipefail`: This option is set when `shell: bash` is explicitly specified. It is not applied by default.
+  - By default, fail-fast behavior is enforced using `set -e` for `sh` and `bash` both. When `shell: bash` is specified, `-o pipefail` is also applied to enforce early exit from pipelines in case of non-zero exit status.
```
hubwriter, line 715:
```suggestion
  - By default, fail-fast behavior is enforced using `set -e` for both `sh` and `bash`. When `shell: bash` is specified, `-o pipefail` is also applied to enforce early exit from pipelines that generate a non-zero exit status.
```

hubwriter (approved):
@iamazeem - Many thanks for raising this pull request. 

I've made a very slight change to the wording, which I've applied so that we can get this merged and published.

Many thanks for your work to get this clarified in the documentation. Your assistance is much appreciated. 🙇 

🚀

PR #28379, opened 2023-09-20
Update error-ssh-add-illegal-option----k.md

guyyaffear:
### Why:
The -K and -A flags are deprecated and have been replaced with by the --apple-use-keychain and --apple-load-keychain flags
These commands will not prompt and warning message that the flag was deprecated.

Closes: 

the existing issue is that the -K flag prompt message of warning and it will be deprecated   

i change the command from `/usr/bin/ssh-add -K ~/.ssh/id_ed25519` to `  /usr/bin/ssh-add --apple-use-keychain ~/.ssh/id_ed25519`
the change happens in file content/authentication/troubleshooting-ssh/error-ssh-add-illegal-option----k.md
added screenshot
image

### Check off the following:


subatoi:
Hi @guyyaffear, thanks for opening this!

I agree this needs changing, but the best option would be to _also_ reuse the note that begins on the following line:

https://github.com/github/docs/blob/main/content/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md?plain=1#L159

So we'd make your change, but also turn the note I link to above into a reusable and then use it both in this article and in the one it's already in.

If you're happy to add those changes to your PR, that'd be great! Otherwise I can make an issue and follow up internally, and add you as co-committer.

guyyaffear:
> Hi @guyyaffear, thanks for opening this!
> 
> I agree this needs changing, but the best option would be to _also_ reuse the note that begins on the following line:
> 
> https://github.com/github/docs/blob/main/content/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md?plain=1#L159
> 
> So we'd make your change, but also turn the note I link to above into a reusable and then use it both in this article and in the one it's already in.
> 
> If you're happy to add those changes to your PR, that'd be great! Otherwise I can make an issue and follow up internally, and add you as co-committer.

i will add those changes in another commit and let you know

subatoi:
Thank you! That's close, but it's not quite what I was thinking. You can turn content into reusables (there's a [guide on doing so](https://docs.github.com/en/contributing/writing-for-github-docs/creating-reusable-content)): I think the most sensible thing to do is to turn the `note` content into one of those, and then make a reference to that content in both places. I'm happy to make those changes myself and add you as a co-committer (so you're still credited) if it's easier or if you don't have time 👍

guyyaffear:
> Thank you! That's close, but it's not quite what I was thinking. You can turn content into reusables (there's a [guide on doing so](https://docs.github.com/en/contributing/writing-for-github-docs/creating-reusable-content)): I think the most sensible thing to do is to turn the `note` content into one of those, and then make a reference to that reusable in both places. I'm happy to make those changes myself and add you as a co-committer (so you're still credited) if it's easier or if you don't have time 👍

@subatoi Okay, I think I didn't fully understand what you said, and I agree with the part about the reusable and make-it note. I prefer you make those changes and add me as a co-committer because I am not sure that I understand what you mean 100%
Thanks 👍

Ggggggghhry (commented):
`~/.zshrc`

subatoi:
No problem at all, we'll let you know when the new PR has been merged with you as co-committer, and close this one at that time 👍

subatoi:
@guyyaffear We've made the change in this commit, with you as co-contributor: 

https://github.com/github/docs/commit/9c9c0b11ed1695e4132eb261ef5c4d30915b63c3

You can see those changes live in for example https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent?platform=mac

I realise I wasn't very clear before so apologies for that, as there were a few moving parts to this change. Where you suggested changing the reference in the page https://docs.github.com/en/authentication/troubleshooting-ssh/error-ssh-add-illegal-option----k, we need to make a separate change for that to make the language clearer that it's deprecated but that that error will surface in a specific troubleshooting context. We'll track that change internally. For now, as you'll see, the change we've made with you credited effectively makes `--apple-use-keychain` the default in the 'main' documentation.

Thank you again for your help and interest in the docs!

PR #28551, opened 2023-09-27
Update explanation for the scope of `permissions` in the provided code examples for token authentication

adimitris:
### Why:
The original docs indicate that in Example 1 `permissions` are defined with scope at workflow level and Example 2 `permissions` at scoped at job level. In reality, for both examples `permissions` are defined at job level. See also [this comment](https://github.com/github/docs/pull/25519#issuecomment-1562532248).

~Since Example 1 is re-usable, I've chosen to modify Example 2, which now has permissions defined at the workflow level. I've also changed the explanation further down the docs to indicate that in Example 1 permissions are scoped at the job level while in Example 2 permissions are scoped at the workflow level.~

Edit: In order to keep best practice for limiting scope, I've updated the explanation for the use of `permissions` in the code snippets to clarify that in both examples `permissions` are defined at job scope rather than workflow scope.

### Check off the following:


meowius:
👋 @adimitris Thanks so much for noticing this discrepancy and opening this PR! We checked in with our engineering team and the code examples were reviewed earlier this year to change the scope from the workflow to the job, since best practice is to limit scopes (see https://github.com/github/docs/pull/25519). 

Because of this, we'd like to leave the code example as is, but update the contextualizing text so it matches what's now in the examples. Let us know if you'd be willing to update this PR to make that change!

AdityaSkhorne:
Great

adimitris:
> 👋 @adimitris Thanks so much for noticing this discrepancy and opening this PR! We checked in with our engineering team and the code examples were reviewed earlier this year to change the scope from the workflow to the job, since best practice is to limit scopes (see #25519).
> 
> Because of this, we'd like to leave the code example as is, but update the contextualizing text so it matches what's now in the examples. Let us know if you'd be willing to update this PR to make that change!

Thank you for the feedback @meowius. I've reverted my original changes and have updated the text to clarify that in both examples `permissions` are defined at the job's scope rather than the workflow's, for best practice.

skedwards88 (approved):
Thanks for this PR! We'll get this merged for you.

PR #28648, opened 2023-09-30
Inconsistency between description and code

LeMyst:
Rollback changes made in commit 370a10d6bfe8340ab253090b05ba6d2b6e5e3cd0

### Why:

Inconsistency

The description before the code is "This example uses the latest minor release of .NET 3."
There is two issues, the version is not the same, and the code doesn't show the usage of juste the version 6.

There is two choices here, correct the description and the example usage to use the version 6 correctly, as wrote in description.
The other choices, the one I made, is to rollback to a previous version who was right, but use a previous version of .NET, but I don't think it's the main point of the example.

EDIT: I've changed the commit and set the version to 6.x to have a more up to date example. I've updated the description.

Have a correct code

### Check off the following:


LeMyst:
> @LeMyst Thanks for your interest in improving the docs! However, would you mind providing a bit more information as to why this change is needed? What makes the current incorrect?

Hello,

The description before the code is "This example uses the latest minor release of .NET 3."
There is two issues, the version is not the same, and the code doesn't show the usage of juste the version 6.

There is two choices here, correct the description and the example usage to use the version 6 correctly, as wrote in description.
The other choices, the one I made, is to rollback to a previous version who was right, but use a previous version of .NET, but I don't think it's the main point of the example.

LeMyst:
I've changed the commit and set the version to 6.x to have a more up to date example. I've updated the description.

skedwards88 (approved):
Thanks for this PR! We'll get this merged for you.

PR #28668, opened 2023-09-30
Fix checkout version

alamirault:
On page https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions, explanation of `checkout/action` version is not good

image


SiaraMist (approved):
Thank you so much for catching this @alamirault! I'll go ahead and get this merged. 🚀

PR #28974, opened 2023-10-07
Making the readability better

Srijan-Baniyal:
I noticed that the readability of the docs was quite hard for the beginners. So, I changed some changes and explained it in a simpler manner


steveward:
Thank you for the work you but into this PR @Srijan-Baniyal, but at this time we won't be accepting these changes.

PR #29186, opened 2023-10-16
Update GitHub Actions Documentation for 'success' Expression

### LINKED ISSUE #28762: Clarify expressions and provide examples for workflows with skipped jobs
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif
https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

There is potential confusion when jobs are skipped in workflows with multiple jobs and with status checks.

> If a run contains a series of jobs that need each other, a failure or skip applies to all jobs in the dependency chain from the point of failure or skip onwards.

This is from https://github.com/github/docs/blob/ce6be8e078a2ab7f8462fb9de8c85b1272a6f7fc/data/reusables/actions/jobs/section-using-jobs-in-a-workflow-needs.md?plain=1#L1

This takes precedence over https://docs.github.com/en/actions/learn-github-actions/expressions#success. While this isn't _wrong_, it should be called out.

[`needs.<job_id>.result`](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context) is particularly interesting because it can be used in [`job.<job_id>.if`](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#overview). However, `needs.<job_id>.result == 'skipped'` will not matter because "a failure or skip applies to all jobs in the dependency chain from the point of failure or skip onwards."

There's additional nuance needed with https://docs.github.com/en/actions/learn-github-actions/expressions#always. While `always()` _always_ runs, it's needed to check conditionals on jobs in a dependency chain that contain a skipped job.

### Additional information

Multiple scenarios were set up and tested in https://github.com/Ubersmake/try-github-actions/actions/runs/6355991752 to try running jobs when a previous job was skipped.

## Maintainer note / Docs plan

To update the docs for this issue, we need to do the following.

1. In https://docs.github.com/en/actions/learn-github-actions/expressions#success, update the description to say:

    ```copy
    Returns `true` when all previous steps have succeeded.
    ```
2. In [this reusable](https://github.com/github/docs/blob/ce6be8e078a2ab7f8462fb9de8c85b1272a6f7fc/data/reusables/actions/jobs/section-using-jobs-in-a-workflow-needs.md?plain=1#L1), add a sentence at the end of line 1 that says something similar to:
 
    ```copy
   If you would like a job to run even if a job it is dependent on did not succeed, use the `always()` conditional expression in `jobs.<job_id>.if`.
    ```

SoundaryaKoutharapu:
### Why:

Closes: #28762

- Updates the description for the 'success' expression to clarify its behavior.
- Adds a sentence to inform users about the `always()` conditional expression for running jobs regardless of the success of dependent jobs.

### Check off the following:


jhosman (approved):
Thanks for opening a PR with this fix! We'll go ahead and get this merged for you 🚀

PR #29319, opened 2023-10-21
docs: Add note on skip-checks

### LINKED ISSUE #25562: GHA skip-checks:true recomendation and behaviour not consistent with git-commit --trailer behaviour
### Code of Conduct

https://github.com/github/docs/blob/main/content/actions/managing-workflow-runs/skipping-workflow-runs.md?plain=1#L29

This concern the use of `skip-checks: true` inserted into the end of a commit message to inhibit GHA workflows.

It is not (easily) possible to insert "TWO blank lines" as the `git commit` command automatically reformats the message based on de-facto rules.  This will remove the double blank lines and replace it with a single blank line, before adding the message to git.

Please try your own instructions to confirm this automatic reformat and please investigate to confirm no obvious git option to inhibit this behaviour exist.  If there is such an option, you should document it is needed to achieve the goal of the github documentation.

Furthermore there is an option to use `git commit --trailer skip-checks:true` which has a default behaviour (that also does not insert TWO blank lines, before the trailer).

So the 3 actions this issue is seeking to resolve:
 * GHA modified to accept a SINGLE blank line and allow multiple key=value and key:value items to be specified in the last paragraph of the git commit message.  Above/below the ones GHA is looking for.  The documentation indicates/implies it must be in isolation at the end, but this does not inter-operate with other mechanisms doing the same.
 * The documentation to be updated to remove the TWO blank lines rule and describe it as ONE blank line.
 * The documentation to provide an example command line demonstrating the --trailer feature to git-commit.
 
Now if GHA must really have its TWO blank lines, then maybe it could test and document how to achieve this, simple sounding action in a commit message in common environments.  Lets say "Linux Command Line" and "Windows Desktop" and "MacOS Desktop".

GHA should also explain how this might interact and work with other mechanisms using trailers, as I understand it, it should be possible to tail with:

```
Message Summary Line

Message Details Paragraph 1

Message Details Paragraph 2

Sign-Off: My Name <user@domain.com>
Another-Special-Thing: option123
key=value
skip-checks:true
My-Key=MyValue

```

The GHA mechanism should also work in a situation like the above.  Since it follows the rules that:
 * Trailing blank lines are removed, to expose the last paragraph block
 * Last paragraph is parsed for `Key:Value` or `key=value` format information, with spaces allow around the '=' and ':' that are disregarded.  Although maybe no space is allowed before ':' as that does not seem common usage.
 * It allows other items to be both before and after the `skip-checks` item.
 * The content is entirely in the last paragraph block of the commit log message.

Following these rules should ensure scope of interoperability with other mechanisms also working like this.

### Additional information

$ git --version
git version 2.39.1
$ grep PRETTY /etc/os-release 
PRETTY_NAME="CentOS Stream 8"

gonmmarques:
### Why:

Following the comment suggestion from https://github.com/github/docs/issues/25562#issuecomment-1764279845, this should clarify the behaviour on the skip-checks.

Closes:  https://github.com/github/docs/issues/25562

### Check off the following:


jc-clark:
Thanks for this contribution @gonmmarques! While I was digging into this, I found that we document the `skip-checks` behavior elsewhere too. I noticed we include what seems to be helpful information and a small example snippet here: "[Skipping and requesting checks for individual commits](https://docs-29319-d8d233.preview.ghdocs.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#skipping-and-requesting-checks-for-individual-commits)" 

> After your commit description, before the closing quotation, add two empty lines followed by `skip-checks: true:`

I'm curious to hear what you think about the closing quotation behavior specifically. Would including the quotes in this way mean that you don't need the `--cleanup=verbatim` option as you suggested?  

There are several ways we can go to improve these docs, and I think whatever we do, we can be sure to update both of these locations so that the `skip-check` documentation is consistent.

gonmmarques:
> Thanks for this contribution @gonmmarques! While I was digging into this, I found that we document the `skip-checks` behavior elsewhere too. I noticed we include what seems to be helpful information and a small example snippet here: "[Skipping and requesting checks for individual commits](https://docs-29319-d8d233.preview.ghdocs.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#skipping-and-requesting-checks-for-individual-commits)"
> 
> > After your commit description, before the closing quotation, add two empty lines followed by `skip-checks: true:`
> 
> I'm curious to hear what you think about the closing quotation behavior specifically. Would including the quotes in this way mean that you don't need the `--cleanup=verbatim` option as you suggested?
> 
> There are several ways we can go to improve these docs, and I think whatever we do, we can be sure to update both of these locations so that the `skip-check` documentation is consistent.

Hello, I was not aware of that documentation you mentioned. But given the fact that indeed Git's default cleanup strategy is 
either `strip` or `whitespace` (see https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupltmodegt), meaning the commit will not have the 2 consecutive empty lines unless you apply the `cleanup=verbatim`option. So I believe it would make sense to also update that documentation with a similar note. What do you think?

jc-clark:
Awesome! Agreed, I think that including it in both locations makes sense. 

To show the same information in multiple locations, we use reusables. Here is the reference documentation to make a reusable file and then reference it in multiple locations: https://docs.github.com/en/contributing/writing-for-github-docs/creating-reusable-content

Once you add the note as a reusable tag to both locations, I can finish up the review and help merge this. Thanks!

gonmmarques:
> Awesome! Agreed, I think that including it in both locations makes sense.
> 
> To show the same information in multiple locations, we use reusables. Here is the reference documentation to make a reusable file and then reference it in multiple locations: https://docs.github.com/en/contributing/writing-for-github-docs/creating-reusable-content
> 
> Let me know if you run into any questions with this method.
> 
> Once you add the note as a reusable tag to both locations, I can finish up the review and help merge this. Thanks!

Makes sense. I have added the reusable, let me know if anything else is missing. Thanks @jc-clark !

data/reusables/commits/about-commit-cleanup.md
```diff
@@ -0,0 +1 @@
+It is important to note that [Git's default commit behavior](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupltmodegt) removes consecutive newlines. As such, it is recommended including the `--cleanup=verbatim` option on your commit.
```
jc-clark, line 1:
```suggestion
By default, Git automatically removes consecutive newlines. To leave the commit message exactly as you entered it, use the `--cleanup=verbatim` option on your commit. For more information, see [`--cleanup=<mode>`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupltmodegt) in the Git documentation. 
```

content/actions/managing-workflow-runs/skipping-workflow-runs.md
```diff
@@ -25,10 +25,12 @@ Workflows that would otherwise be triggered using `on: push` or `on: pull_reques
 - `[skip actions]`
 - `[actions skip]`
 
-Alternatively, you can end the commit message with two empty lines followed by either:
+Alternatively, you can add a `skip-checks` trailer to your commit message . The trailers section should be included at the end of your commit message and be proceeded by two empty lines. If you already have other trailers in your commit message, `skip-checks` should be last. You can use either of the following:
```
jc-clark, line 28:
```suggestion
Alternatively, you can add a `skip-checks` trailer to your commit message. The trailers section should be included at the end of your commit message and be proceeded by two empty lines. If you already have other trailers in your commit message, `skip-checks` should be last. You can use either of the following:
```

data/reusables/commits/about-commit-cleanup.md
```diff
@@ -0,0 +1 @@
+By default, Git automatically removes consecutive newlines. To leave the commit message exactly as you entered it, use the `--cleanup=verbatim` option on your commit. For more information, see [`--cleanup=<mode>`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupltmodegt) in the Git documentation.
```
jc-clark, line 1:
```suggestion
By default, Git automatically removes consecutive newlines. To leave the commit message exactly as you entered it, use the `--cleanup=verbatim` option on your commit. For more information, see [`--cleanup=<mode>`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupltmodegt) in the Git documentation.
```

jc-clark (approved):

jc-clark:
Thanks for your help with this @gonmmarques! I added this to the merge queue so the changes should be live soon.

PR #29436, opened 2023-10-26
fix(docs): Updated the documentation on using local workflow

### LINKED ISSUE #19023: Clarify what path value to use when using an action in the same repository as the workflow
### Code of Conduct

This pertains to the [Using an action in the same repository as the workflow](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow) example in the [section on jobs.<job_id>.steps[*].uses](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses).

The text shows that the value of `uses` when one want to use an action that is in the same repository as the workflow should be (something of the form) `./path/to/dir` and then goes on to explain that this means "the path to the directory that contains the action in your workflow's repository."  To any user of any shell, `./path/to/dir` looks like a relative file path, but the question is then "relative to what?"  Relative to the file for the workflow we are currently running?  Relative to the root of the repository containing that workflow?  (Based on the example that follows, which uses a `uses` value of `./.github/actions/my-action`, this would seem a more likely guess.)  Relative to the value of `$GITHUB_WORKSPACE`?

In the example given, these last two are the same—that is, `$GITHUB_WORKSPACE` _is_ the root of the repository containing the workflow.  But suppose we checked out the repository somewhere else, e.g.
```
jobs:
  my_first_job:
    steps:
      - name: Check out repository
        uses: actions/checkout@v3
        with:
          path: source_files
      - name: Use local my-action
        uses: ./.github/actions/my-action
```
It turns out the `my-action` action will no longer be found unless we change the value of `uses` to be relative to the value of _`$GITHUB_WORKSPACE`_; that is, we must modify the last line to be
```
        uses: ./source_files/.github/actions/my-action
```
This requirement should be made clear.

Another thing that isn't made clear is that we apparently _must_ use the leading `./` when specifying the path.  If one is thinking of this the way one usually thinks of file system paths, one might wonder why the value of `uses` can't simply be `'.github/actions/my-action'` (or, for my modified version, `source_files/.github/actions/my-action`).  Presumably, the reason for the "leading `./`" requirement is that it affords a clear distinction from the case where we are using an action from some (other) public repository: if we write `uses: ./actions/heroku`, the workflow will look for a file called `action/heroku/actions.yml` in _our_ repository, whereas if we write `uses: actions/heroku`, it will look for a file called `actions.yml` in the root of the `actions/heroku` repository.

Could we use an absolute path with `uses`—for example, something like `/home/runner/work/.../.github/actions/my-action`?  This could equally well be thought of as "the path to the directory that contains the action in your workflow's repository."  I haven't tried this yet, but I suspect the answer is _No_.

### Additional information

_maintainer addition_
Anyone is welcome to open a PR following the instructions in https://github.com/github/docs/issues/19023#issuecomment-1269090748

Jegan-Kunniya:
### Why:

Closes: #19023

The documentation on referencing local workflow file to the repository is updated in more than one location in a resuable content snippet. (See modified files).

The changes emphasizes the relative path to the checkout location of the repo, while referencing the workflow as suggested in [comment](https://github.com/github/docs/issues/19023#issuecomment-1269090748)

### Check off the following:


Jegan-Kunniya:
@cmwilson21 - Any leads on the review? I see it is still in waiting stage. Just checking.

data/reusables/actions/workflows/section-referencing-an-action-from-the-same-repository.md
```diff
@@ -0,0 +1,28 @@
...
+|-- hello-world (repository)
+|   |__ .github
+|       └── workflows
+|           └── my-first-workflow.yml
+|       └── actions
+|           |__ hello-world-action
+|               └── action.yml
+```
+
+The path is relative (`./`) to the default working directory (`github.workspace`, `$GITHUB_WORKSPACE`), unless the repository is checked out to a different location in the checkout step.
```
rachaelrenk, line 13:
```suggestion
The path is relative (`./`) to the default working directory (`github.workspace`, `$GITHUB_WORKSPACE`). If the action checks out the repository to a location different than the workflow, the relative path used for local actions must be updated.
```

rachaelrenk (approved):
I tweaked the language in the reusable and reviewed the new example. Per the content design approach in [this comment,](https://github.com/github/docs/issues/19023#issuecomment-1269090748) I think this looks good! I'll make sure the tests are passing and get this added to the merge queue. ✨

PR #29443, opened 2023-10-26
docs: fix typo in content/webhooks/about-webhooks

henrique11varela:
Found this typo while studying so i'm fixing it.


mchammer01 (approved):
@henrique11varela - thanks for spotting this typo, and fixing it ✨

PR #30082, opened 2023-11-18
Fix concurrency groups example

garysassano:
### Why:

The description of the concurrency groups example states the following:
>[...] if a new commit is pushed to the main branch while a previous run is still in progress, the previous run will be cancelled and the new one will start:

However, in the given YAML configuration, the parameter `cancel-in-progress: true` is not included. This omission means that the default behavior will be followed, where the event is queued up for execution in the next run, instead of cancelling the ongoing run.

### Check off the following:


janiceilene:
Thanks so much for opening a PR! I'll get this triaged for review ✨

jc-clark (approved):
✨ Thanks for catching this and improving the docs @garysassano! Changes look good to me, I'll go ahead and merge this.

PR #30839, opened 2023-12-27
Update managing-a-merge-queue.md

aldeed:
Clarify when `merge_group` is needed

### Why:

The note mentioned both `pull_request` and `push` but seems to be relevant only to `pull_request`.

Clarify the language in Managing a Merge Queue article

### Check off the following:

### DIFF
--- a/content/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue.md
+++ b/content/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue.md
@@ -32,11 +32,11 @@ For more information on merging a pull request using a merge queue, see "[AUTOTI
 
 ### Triggering merge group checks with {% data variables.product.prodname_actions %}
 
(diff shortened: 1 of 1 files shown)


nguyenalex836:
@aldeed Thanks so much for opening a PR! I'll get this triaged for review ✨

jc-clark:
Hey @aldeed! Thanks for bringing this up. 

I had a chat with our engineering team about this change. While it is technically possible, we don't offer support for using the `push` event like this. The `merge_group` events are the explicit callbacks to execute required checks, which is why the docs emphasize using the `merge_group` event. For this reason, I'm going to close this PR and leave the docs as is.

aldeed:
@jc-clark The main problem with the current language is that when we turned on a merge queue and followed these instructions, all our checks ran twice because we already had `push` on that workflow. So adding `merge_group` trigger to a workflow that already runs on `push` is actually _incorrect_. Feel free to tweak the language, but I was trying to prevent others from spending hours trying to figure out this situation.

jc-clark:
Ah, that makes sense @aldeed! And thank you for the deeper context here. I can go back to our engineering team and see what they say.

PR #30855, opened 2023-12-29
Incorrect windows syntax & instructions fixed

Vortun:
Some of the syntax was incorrect in the instructions regarding the path to your ssh key on windows (/ used instead of \, no : after drive label)

Also corrected an error where the guide suggested to use CMD instead of PowerShell for commands that only work in PowerShell.

### Why: Some of the information regarding SSH connection was incorrect on Windows.

Closes: 

### Check off the following:

### DIFF
--- a/content/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
+++ b/content/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
@@ -83,7 +83,7 @@ If you are a site administrator for {% data variables.location.product_location 
    {% windows %}
 
    ```powershell
-   > Enter a file in which to save the key (/c/Users/YOU/.ssh/id_ALGORITHM):[Press enter]
+   > Enter a file in which to save the key (/c:\Users\YOU\.ssh\id_ALGORITHM):[Press enter]
    ```
 
    {% endwindows %}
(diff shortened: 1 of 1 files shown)


nguyenalex836:
@Vortun Thanks so much for opening a PR! I'll get this triaged for review ✨

stevecat:
Hey @Vortun! Thanks for opening this ✨  I just tested this on my own Windows machine and the prompt was a mind-bending combination of both slashes 😅 

```
Enter file in which to save the key (C:\Users\YOU/.ssh/id_ed25519):
```

Did you test this yourself and see `/c:\Users\YOU\.ssh\id_ALGORITHM`? I know Powershell supports both `/` and `\` in paths and it would be great to verify what this actually is. 

I'm curious about the first `/` too in your path too. Is that a typo or was that returned too?

Your other change, removing the mention of command prompt, looks great. Once we've figured this out, I'll approve and we can get this merged!

Vortun:
You're absolutely correct @stevecat! I missed removing the slash before C: (fixed in a new commit), however I can't reproduce this random combination of slashes you're seeing? It displays only backslashes on both my windows and ubuntu machine.

Screenshot for reference:
Screenshot from 2024-01-04 15-35-37

guntrip (approved):
Thanks for fixing that, @Vortun! ✨ 

Looking around, I think my Powershell is just spectacularly misconfigured, it looks like the format you're seeing is standard.

I'll get this merged now! Thanks again for contributing to our docs! :tada: ✨

PR #31191, opened 2024-01-21
Update generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md

charleskimbac:
### Why:
Fix some typos regarding what is actually shown in the terminal, and also fix a command which previously didn't work.
"ssh-add c:\Users\YOU\.ssh\id_ed25519" (with backslashes) leads to error "c:UsersYOU.sshid_ed25519: No such file or directory".

### Notes:
I had to run "eval \`ssh-agent -s\`" in bash before being able to run ssh-add [(link)](https://stackoverflow.com/questions/17846529/could-not-open-a-connection-to-your-authentication-agent), should maybe add this to the docs.

### Check off the following:

### DIFF
--- a/content/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
+++ b/content/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent.md
@@ -74,7 +74,7 @@ If you are a site administrator for {% data variables.location.product_location 
    {% windows %}
 
    ```powershell
-   > Enter a file in which to save the key (c:\Users\YOU\.ssh\id_ALGORITHM):[Press enter]
+   > Enter file in which to save the key (/c/Users/YOU/.ssh/id_ALGORITHM):[Press enter]
    ```
 
    {% endwindows %}
@@ -169,7 +169,7 @@ Before adding a new SSH key to the ssh-agent to manage your keys, you should hav
    {% data reusables.ssh.add-ssh-key-to-ssh-agent %}
 
    ```powershell
-   ssh-add c:\Users\YOU\.ssh\id_ed25519
+   ssh-add c:/Users/YOU/.ssh/id_ed25519
    ```
 
 {% data reusables.ssh.add-public-key-to-github %}


nguyenalex836:
@charleskimbac Thanks so much for opening a PR! I'll get this triaged for review ✨

stevecat:
Hi @charleskimbac, thanks so much for opening this. This is interesting because we've only [just changed the direction of the slashes](https://github.com/github/docs/pull/30855) based on previous feedback.

It was my understanding that the direction of the slashes wouldn't actually matter to Powershell, which can cope with both. To confirm this, I created a fresh Windows 11 VM to test. It looks like `ssh-add` really doesn't care how you provide the slashes:

```
PS C:\Users\adminadmin> ssh-add c:/Users/adminadmin/.ssh/id_ed25519
Identity added: c:/Users/adminadmin/.ssh/id_ed25519 (stevecat@github.com)

PS C:\Users\adminadmin> ssh-add c:/Users/adminadmin\.ssh\id_ed25519
Identity added: c:/Users/adminadmin\.ssh\id_ed25519 (stevecat@github.com)

PS C:\Users\adminadmin> ssh-add c:\Users\adminadmin\.ssh\id_ed25519
Identity added: c:\Users\adminadmin\.ssh\id_ed25519 (stevecat@github.com)
``` 

The same is true with `ssh-keygen`.

I'm curious if the errors you're seeing are due to a different version of `ssh-add` (for example, MinGW).

We can only really support a default installation but, as both slash-directions work on a default install, and it looks like switching back to forward slashes will help some folks such as yourself, let's do that!

I'll approve this and get your changes merged in. Thank you! ✨✨✨✨ 

>I had to run "eval `ssh-agent -s`" in bash before being able to run ssh-add [(link)](https://stackoverflow.com/questions/17846529/could-not-open-a-connection-to-your-authentication-agent), should maybe add this to the docs.

Did you run `Get-Service -Name ssh-agent | Set-Service -StartupType Manual` and `Start-Service ssh-agent` first? As I had the fresh VM, I tested this too, and was only able to use `ssh-add` once I'd ran those two commands from our docs. As I've confirmed this is working as expected with a default installation, I think we should stick to our current procedure.

guntrip (approved):

charleskimbac:
Hey @stevecat so I want to start by saying that I only have a bare knowledge of all of this, so apologies if I say something or am saying something incorrectly.
In-line with step 2 of "Adding your SSH key to the ssh-agent", it starts by saying "In a terminal window without elevated permissions..." so I went back (from powershell) to my bash terminal (mingw) in VSCode, and attempted `ssh-add`. I guess you don't need the `"eval ssh-agent -s"` if you do it in Powershell vs bash, and it also explains the forward/back-slash discrepancy.

stevecat:
Ooh, thanks @charleskimbac! I'm experimenting with this and it appears that running the `ssh-add` command in Powershell might set things up so the key works in Powershell *and* Git Bash. This surprised me tbh!

Launching `ssh-agent` in Git Bash (using the eval command) does then allow you to run `ssh-add` as you mentioned.

I'm currently working on a project to revamp our SSH docs, and am paying extra attention to the Windows procedures. I'll work with our team to decide on the best option for that second step; either sticking with Powershell or adding the eval command.

PR #31252, opened 2024-01-24
feat: describe conditional values

### LINKED ISSUE #23795: Document truthiness handling for expressions
### Code of Conduct

https://docs.github.com/en/actions/learn-github-actions/expressions

Some part of the page should explain how:

```
if: ${{ vars.FOO }}
```

works. Specifically, if `vars.FOO` isn't set, that's presumably treated as the equivalent of an empty something (string?).

if `vars.FOO` is set, it's presumably treated as a string.

Note that I'm using `vars` above, but the same applies for anything that would naturally be stringy (`env`, `inputs`(that are stringy), `secrets`, `outputs`,....)

Are all non-empty strings true?

Is the string `"0"` true?
Is the string `"false"` true?

There's some prose that says that when comparing two things, if they're of different types, they're converted to numbers.

But it isn't particularly clear if, `if: ${{ some_expression }}` is treated as `if: ${{ (some_expression) == true }}` which thus results in strings being converted to numbers and true being converted to a number.

### Additional information

[maintainer edit]
Content plan [here](https://github.com/github/docs/issues/23795#issuecomment-1647481101)

deemp:
### Why:

Closes: https://github.com/github/docs/issues/23795

I used this hidden knowledge in https://github.com/freckle/stack-action/pull/35 and decided to document it.

### Check off the following:


nguyenalex836:
@deemp Thanks so much for opening a PR! I'll get this triaged for review ✨

content/actions/learn-github-actions/expressions.md
```diff
@@ -15,7 +15,7 @@ versions:
 
 You can use expressions to programmatically set environment variables in workflow files and access contexts. An expression can be any combination of literal values, references to a context, or functions. You can combine literals, context references, and functions using operators. For more information about contexts, see "[AUTOTITLE](/actions/learn-github-actions/contexts)."
 
-Expressions are commonly used with the conditional `if` keyword in a workflow file to determine whether a step should run. When an `if` conditional is `true`, the step will run.
+Expressions are commonly used with the conditional `if` keyword in a workflow file to determine whether a step should run. When an `if` conditional is `truthy` (in [JavaScript sense](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)), the step will run. All values are `truthy` except for `false`, `0`, `-0`, `0n`, `""`, `''`, `null`, `undefined`, `NaN`. You can evaluate an `!!<expression>` (e.g., in your browser console) to check whether the `<expression>` is `truthy`. For example, `!!"false"` produces `true`, so `"false"` is `truthy`.
```
ferdnyc, line 19: This... isn't correct at all.

The workflow is parsed in a string context (except for numerals and booleans), so JavaScript truthiness is irrelevant. There's no way to define a true JavaScript `NaN` or `undefined` in YML, you end up with the strings `'NaN'` or `'undefined'`. Nearly any string will evaluate `true`.

In this workflow:
```yml
    strategy:
      matrix:
        java: [17]
        runme:
          - NaN
          - undefined
          - yes
          - no
          - false
          - 0
          - 1
          - -0

    steps:
      - name: true-path
        if: matrix.runme
        run: echo "${{ matrix.runme }} evaluates true"
      - name: false-path
        if: '! matrix.runme'
        run: echo "${{ matrix.runme }} evaluates false"
```

...The only `matrix.runme` value for which `false-path` will be executed are `false` and `0`. (And `-0`, technically, but that's only because it's interpreted as `0`. There's no distinction in YAML.) All of the others will evaluate `true`.

ferdnyc: Oh, sorry, `null` is also a valid YAML primitive, and it will also evaluate `false`.

Honestly, I think this is pretty well explained in [the Literals section of the Expressions docs](https://docs.github.com/en/actions/learn-github-actions/expressions#literals).

deemp: 1. It's weird (for me) that `'! matrix.runme'` evaluates to a condition. I'd expect that only `! matrix.runme` or `${{ matrix.runme }}` evaluates to a condition.
2. I ran the suggested job ([link](https://github.com/deemp/gists/actions/runs/7665774871)). I can confirm that the non-truthy values are `false`, `0`, `-0`, `""`, `''`, `null`. I can't see where these values are listed in [the Literals section](https://docs.github.com/en/actions/learn-github-actions/expressions#operators). What do you think about just extending the `boolean` docs?

Data type | Literal value
---------------|-------------------
`boolean`  | falsy (`false`, `0`, `-0`, `""`, `''`, `null`) or truthy (non-falsy)

deemp: Updated the PR

content/actions/learn-github-actions/expressions.md
```diff
@@ -126,7 +128,7 @@ env:
 {% endraw %}
 
 In this example, we're using a ternary operator to set the value of the `MY_ENV_VAR` environment variable based on whether the {% data variables.product.prodname_dotcom %} reference is set to `refs/heads/main` or not. If it is, the variable is set to `value_for_main_branch`. Otherwise, it is set to `value_for_other_branches`.
-It is important to note that the first value after the `&&` condition must be `truthy` otherwise the value after the `||` will always be returned.
+It is important to note that the first conditional after the `&&` must be `true`. Otherwise, the value after the `||` will always be returned.
```
SiaraMist, line 131:
```suggestion
It is important to note that the first value after the `&&` must be truthy. Otherwise, the value after the `||` will always be returned.
```
Just updating the language slightly!

SiaraMist (approved):
Thank you so much for updating the docs, @deemp! I'll go ahead and get this pull request merged. ✨

PR #31322, opened 2024-01-29
Fix cleanup example to use pull_request_target trigger

### LINKED ISSUE #31321: The workflow named "cleanup caches by a branch" is not working as expected
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

Modify the workflow example in [Managing caches > Force deleting cache entries](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries)

Why:
When triggered by the `pull_request` event, the `GITHUB_TOKEN` does not have permission to delete the cache of the base repository in a cross-repository pull request. This results in the error `Error: Resource not accessible by integration`.

### Additional information

This always happens for cross-repository pull requests.

Mogyuchi:
### Why:

Closes: #31321 

image

### Check off the following:

### DIFF
--- a/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
+++ b/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
@@ -343,7 +343,7 @@ This example workflow uses `gh-actions-cache` to delete up to 100 caches created
 ```yaml
 name: cleanup caches by a branch
 on:
-  pull_request:
+  pull_request_target:
     types:
       - closed
 


nguyenalex836:
@Mogyuchi Thanks so much for opening a PR! I'll get this triaged for review ✨

corneliusroemer (commented):
I tested this change and it does _not_ fix the issue unfortunately

I still get the same error, at least when triggering manually (having added `workflow_dispatch:` to do so)

Maybe that's as expected, maybe the permissions are only granted when you merge a PR, hard to tell/test

I see, the right permissions might be auto-granted due to this change:
>When a workflow is triggered by the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event, the GITHUB_TOKEN is granted read/write repository permission, even when it is triggered from a public fork. For more information, see "[Events that trigger workflows](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)."
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I've added explicit token permissions in #31582, this allowed me to test via `workflow_dispatch:`

Mogyuchi:
@jc-clark
https://github.com/github/docs/pull/31582#issuecomment-1979764206
> What would you think of keeping `pull_request` as the triggering event here but adding a note above the example that links to "[`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)" and describes the cross-repo pull request use case?

No… I think we should use "[`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)".
Especially in public repositories, I think there is a possibility that cross-repo pull requests will be made even if the repository owner does not assume it. Therefore, I think we should present the example that takes into account the existence of cross-repo pull requests.

Boozing1973 (commented):
Duplicate of #

PR #31582, opened 2024-02-12
fix: add necessary permissions to example workflow

### LINKED ISSUE #31321: The workflow named "cleanup caches by a branch" is not working as expected

corneliusroemer:
### Why:

Closes #31321

Uses the diff from https://github.com/github/docs/pull/31322/
but adds explicit permissions as well
which allows it to be tested with
`workflow_dispatch:` if users desire so

Also use `github.ref_name` instead of the pull_request.number to query the branch
I couldn't get the original to work in local tests
but with `ref_name` it works flawlessly.

### Check off the following:

### DIFF
--- a/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
+++ b/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
@@ -343,13 +343,15 @@ This example workflow uses `gh-actions-cache` to delete up to 100 caches created
 ```yaml
 name: cleanup caches by a branch
 on:
-  pull_request:
+  pull_request_target:
     types:
       - closed
 
 jobs:
   cleanup:
     runs-on: ubuntu-latest
+    permissions:
+      actions: write
     steps:
       - name: Cleanup
         run: |
@@ -369,7 +371,7 @@ jobs:
         env:
           GH_TOKEN: {% raw %}${{ secrets.GITHUB_TOKEN }}{% endraw %}
           REPO: {% raw %}${{ github.repository }}{% endraw %}
-          BRANCH: refs/pull/{% raw %}${{ github.event.pull_request.number }}{% endraw %}/merge
+          BRANCH: {% raw %}${{ github.ref_name }}{% endraw %} 
 ```
 
 Alternatively, you can use the API to automatically list or delete all caches on your own cadence. For more information, see "[AUTOTITLE](/rest/actions/cache#about-the-cache-in-github-actions)."
--- a/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
+++ b/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
@@ -371,7 +371,7 @@ jobs:
         env:
           GH_TOKEN: {% raw %}${{ secrets.GITHUB_TOKEN }}{% endraw %}
           REPO: {% raw %}${{ github.repository }}{% endraw %}
-          BRANCH: {% raw %}${{ github.ref_name }}{% endraw %} 
+          BRANCH: {% raw %}${{ github.ref_name }}{% endraw %}
 ```
 
 Alternatively, you can use the API to automatically list or delete all caches on your own cadence. For more information, see "[AUTOTITLE](/rest/actions/cache#about-the-cache-in-github-actions)."
--- a/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
+++ b/content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
@@ -371,7 +371,7 @@ jobs:
         env:
           GH_TOKEN: {% raw %}${{ secrets.GITHUB_TOKEN }}{% endraw %}
           REPO: {% raw %}${{ github.repository }}{% endraw %}
-          BRANCH: {% raw %}${{ github.ref_name }}{% endraw %}
+          BRANCH: {% raw %}${{ github.head_ref }}{% endraw %}
 ```
 
 Alternatively, you can use the API to automatically list or delete all caches on your own cadence. For more information, see "[AUTOTITLE](/rest/actions/cache#about-the-cache-in-github-actions)."


content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
```diff
@@ -369,7 +371,7 @@ jobs:
         env:
           GH_TOKEN: {% raw %}${{ secrets.GITHUB_TOKEN }}{% endraw %}
           REPO: {% raw %}${{ github.repository }}{% endraw %}
-          BRANCH: refs/pull/{% raw %}${{ github.event.pull_request.number }}{% endraw %}/merge
+          BRANCH: {% raw %}${{ github.ref_name }}{% endraw %}
```
corneliusroemer, line 374:
```suggestion
          BRANCH: {% raw %}${{ github.ref_name }}{% endraw %}
```

content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
```diff
@@ -369,7 +371,7 @@ jobs:
...
```
Mogyuchi, line 374: Indeed, using ref_name in workflow_dispatch is the correct approach.
IMG_0161
However, this approach is not correct in this example.
Because, [the cache is created for the merge ref (`refs/pull/.../merge`) when a cache is created by a workflow run triggered on a pull request.](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) Therefore, it will not work correctly with ref_name.
IMG_0162
IMG_0163

nguyenalex836:
@corneliusroemer Thanks so much for opening a PR! I'll get this triaged for review ✨

content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
corneliusroemer: I think I might have been wrong - `ref_name` ends up being main/master after merge. We'd need `github.head_ref` here instead.

corneliusroemer:
@Mogyuchi made a good point that `github.ref_name` might not be the right name here, instead one should probably use `github.head_ref`.

content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
corneliusroemer, line 374: Do you think this would work @Mogyuchi?

```suggestion
          BRANCH: {% raw %}${{ github.head_ref }}{% endraw %}
```

When I used the existing code, `gh actions-cache` didn't return anything for me

Mogyuchi:
> @Mogyuchi made a good point that `github.ref_name` might not be the right name here, instead one should probably use `github.head_ref`.

@corneliusroemer 
Sorry, I should have just said it straight.
In short, to clear the cache for a pull request, we need to stick to `refs/pull/${{ github.event.pull_request.number }}/merge`. Alternatively, if we want to clear the branch cache as well, we need to do it like [discordjs-japan/om # 90 (pull_request should be replaced by pull_request_target)](https://github.com/discordjs-japan/om/pull/90/files).

content/actions/using-workflows/caching-dependencies-to-speed-up-workflows.md
Mogyuchi: @corneliusroemer 
If you have tried with workflow_dispatch, then your perception that it does not work is correct. This is because `github.head_ref` is an empty string in the workflow_dispatch event.

Mogyuchi:
…I feel like I need to explain how what is written in [restrictions-for-accessing-a-cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) affects cache cleanup.
Simply put, for example, if you have a pull request 123 with head(current) branch `feature-b`, the cache created by the `push` event will be for `feature-b` and the cache created by the `pull_request` event will be for `refs/pull/123 /merge`.
So, if the cache you want to delete was created in the `push` event, you need to specify the cache for `feature-b`, and if it was created in the `pull_request` event, you need to specify `refs/pull/123/merge`.

Pelambrds (commented):
Aprovar

jc-clark:
Hello @corneliusroemer and @Mogyuchi - thank you for the suggestions here. We had a chance to dig into this on our team and we agreed that we shouldn't change this example to use `pull_request_target` vs `pull_request`. The majority of our docs readers will be using the `pull_request` trigger and using this as the default aligns with the rest of our documentation as well. 

That said, I didn't want to outright close this PR because I think it's a valid change to cover the cross-repository pull request use case. 

In the "Events that trigger workflows" article, we do have a warning box about the permissions: "[`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)"

Which looks like it has the context someone would need for the cross-repo pull requests. 

A few questions for you: 

- What would you think of keeping `pull_request` as the triggering event here but adding a note above the example that links to "[`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)" and describes the cross-repo pull request use case? 
- Based on [this comment](https://github.com/github/docs/pull/31582#discussion_r1486602743), it sounds like we should keep the `permissions` lines in this example even if we just use `pull_request`. Does that sound right? 
- Should this line remain the same: `BRANCH: refs/pull/{% raw %}${{ github.event.pull_request.number }}{% endraw %}/merge`

Mogyuchi:
@jc-clark 

https://github.com/github/docs/pull/31582#discussion_r1486602743
> Therefore, I believe it is more appropriate to make this a separate pull request from https://github.com/github/docs/pull/31322 .

What I wanted to say is that I don't want this pull request to be based on my pull request.

---
> What would you think of keeping `pull_request` as the triggering event here but adding a note above the example that links to "[`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)" and describes the cross-repo pull request use case?

I think we should talk about this in https://github.com/github/docs/pull/31322, so I'll answer [there](https://github.com/github/docs/pull/31322#issuecomment-1982698996).

> Based on https://github.com/github/docs/pull/31582#discussion_r1486602743, it sounds like we should keep the `permissions` lines in this example even if we just use `pull_request`. Does that sound right?

Yes. We should specify the permission. https://securitylab.github.com/research/github-actions-building-blocks/#following-the-principle-of-least-privilege

> Should this line remain the same: `BRANCH: refs/pull/{% raw %}${{ github.event.pull_request.number }}{% endraw %}/merge`

Yes. Related pull request: https://github.com/github/docs/pull/23612

jc-clark:
Thanks for the additional context @Mogyuchi and @corneliusroemer. We're deciding to make some updates to this article internally to account for using the `pull_request_target` event. 

Our engineers agreed that the example is good as is. So we are going to leave it using the default `pull_request` event and then add some information about using `pull_request_target` as an option. 

I'm going to close out this PR and our team will merge some improvements to this article soon.

PR #31761, opened 2024-02-21
Remove need for `KEYCHAIN_PASSWORD` secret in deploying-xcode-applications

cweider:
### Why:

Generating the ephemeral keychain's password in the script removes the need for the user to generate and store one manually.

Step describing `KEYCHAIN_PASSWORD` secret is removed. Script modified to generate random bytes to use as the keychain's password automatically (`openssl rand` – but it could be `$RANDOM`).

### Check off the following:


SiaraMist:
Hey @cweider, thanks so much for opening this PR to update the docs! ✨ 

The reason we recommend users create a secret and pass the secret to the workflow is because it is more secure than generating the password via the command line. You can see this documented in [Using secrets in a workflow](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow) with a little more information on why we make that recommendation. 

While your solution is simpler to implement 👍, it is less secure. For that reason, I am going to close out this PR. Thanks again for your interest in improving GitHub's docs!

PR #31779, opened 2024-02-22
Fix migrating-to-internal-repositories command pipe whitespace

janbrasna:
Extra whitespace in example migration command, differs between two ifversion ghes else blocks.

### Why:

Screen Shot 2024-02-22 at 15 29 26

Fixes whitespace in one migration command after pipe to make it consistent with the other.

### Check off the following:


janbrasna:
FYI I think the two distinct code blocks behind `ifversion ghes else` test make no sense today, as all currently published docs for available versions always satisfy the first condition?

The extra step was added in 3383ea7, and subsequently simplified to a point when with `ghae` removal in eb9571a the only logic left is `ifversion ghes` always matching, leaving the `else` code block with extra steps completely obsolete today.

I haven't removed the leftover code block myself to leave that to someone with mode knowledge about legacy version differences and if/when this might eventually be needed, so you can probably clean this up internally after some more examination why/how it got to this stage.

nguyenalex836 (approved):

PR #32544, opened 2024-04-16
Clarifies outputs in reusable workflows

### LINKED ISSUE #32543: Lack of clarity re: reusable workflow `outputs`
### Code of Conduct

https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow

The existing text is not clear that the workflow level `outputs:` may be mapped to job-level, but not step-level outputs.  The example workflow maps step-level outputs to job-level outputs, then to workflow-level outputs, but the text does not explain that this is needed and that attempting to map workflow-level outputs directly to step-level outputs won't work.  Attempts to use step-level outputs do not generate an error but pass an empty output to the caller, compounding the confusion.

### Additional information

A proposed clarification will follow momentarily in the form of a PR.

APCBoston:
### Why:

Closes: #32543

The existing text is not clear that the workflow level `outputs:` may be mapped to job-level, but not step-level outputs.  Attempts to use step-level outputs do not generate an error but pass an empty output.  I have added two sentences to clarify that only job-level outputs may be mapped to workflow outputs.

Two sentences at the end of the section https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-outputs-from-a-reusable-workflow.  The diff should be very clear.

### Check off the following:

- [ x ] I have reviewed my changes in staging, available via the **View deployment** link in this PR's timeline (this link will be available after opening the PR).

- [ x ] For content changes, I have completed the [self-review checklist](https://docs.github.com/en/contributing/collaborating-on-github-docs/self-review-checklist).


nguyenalex836:
@APCBoston Thanks so much for opening a PR! I'll get this triaged for review ✨

content/actions/using-workflows/reusing-workflows.md
```diff
@@ -363,7 +363,7 @@ A reusable workflow may generate data that you want to use in the caller workflo
...
 
-The following reusable workflow has a single job containing two steps. In each of these steps we set a single word as the output: "hello" and "world." In the `outputs` section of the job, we map these step outputs to job outputs called: `output1` and `output2`. In the `on.workflow_call.outputs` section we then define two outputs for the workflow itself, one called `firstword` which we map to `output1`, and one called `secondword` which we map to `output2`.
+The following reusable workflow has a single job containing two steps. In each of these steps we set a single word as the output: "hello" and "world." In the `outputs` section of the job, we map these step outputs to job outputs called: `output1` and `output2`. In the `on.workflow_call.outputs` section we then define two outputs for the workflow itself, one called `firstword` which we map to `output1`, and one called `secondword` which we map to `output2`.  Note that only job-level outputs can be mapped to workflow outputs.  Step-level outputs must first be mapped to job-level outputs as shown below.
```
jc-clark, line 366:
```suggestion
The following reusable workflow has a single job containing two steps. In each of these steps we set a single word as the output: "hello" and "world." In the `outputs` section of the job, we map these step outputs to job outputs called: `output1` and `output2`. In the `on.workflow_call.outputs` section we then define two outputs for the workflow itself, one called `firstword` which we map to `output1`, and one called `secondword` which we map to `output2`.  For more information about setting outputs with `jobs.<job_id>.outputs`, see "[AUTOTITLE](/actions/using-jobs/defining-outputs-for-jobs#overview)."
```

jc-clark (commented):
Hey @APCBoston, thank you for making this update! I dug around a little bit and noticed that "[Defining outputs for jobs](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs#overview)" specifically calls out setting outputs within a job: 

> You can use jobs.<job_id>.outputs to create a map of outputs for a job. 

So what do you think about this suggested change here that links directly to the defining outputs article?

content/actions/using-workflows/reusing-workflows.md
APCBoston: I am all for linking cross-references and I would support adding your final sentence (with the link) after my added sentences distinguishing between job- and step-level outputs.  Alternatively, a caveat such as what I drafted could be added to the documentation that you linked instead of (or in addition to) being added here.   

Here's what I found unclear: the example on this page references `${{ steps.step1.outputs.firstword }}` within the `jobs.<job_id>.outputs` mapping, and then uses that job output in the `on.workflow_call.outputs` mapping.  Why not simplify as follows:

```
on:
  workflow_call:
    # Map the workflow outputs to job outputs
    outputs:
      firstword:
        description: "The first output string"
        value: ${{ jobs.example_job.steps.step1.outputs.test }}
```

Clearly step-level outputs _exist_ as they are used in the example mapping to generate job-level outputs... one could imagine cases where the job-level output might involve some transformation or manipulation of step-level outputs, but since they are merely passed through here, why not reference them directly?  The answer is that you just can't do it that way -- the `workflow_call` trigger can see job outputs, but it cannot see the step outputs inside the job definition.  However, the current documentation does not make this clear.  The runner logs ALSO do not make this clear, as `${{ jobs.example_job.steps.step1.outputs.test }}` simply evaluates to a null value and gets substituted in as `''`.

My proposed text, "Note that only job-level outputs can be mapped to workflow outputs.  Step-level outputs must first be mapped to job-level outputs as shown below", is intended to clarify this specific issue.

jc-clark: I see what you're saying @APCBoston, thanks for the clarification! That was helpful. Your analysis makes sense to me, and I noticed that our [`on.workflow_call.outputs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_calloutputs) documentation does say: 

> The `value` must be set to the value of an output from a job within the called workflow.

This implies that step-level outputs are not available to the caller workflow, but it doesn't explicitly describe the dynamic you're describing. 

So, inspired by your original suggestion, I think we can go back to explicitly calling this out with the language from the Workflow syntax for GitHub Actions article. 

What do you think of something like this? 

```suggestion
The following reusable workflow has a single job containing two steps. In each of these steps we set a single word as the output: "hello" and "world." In the `outputs` section of the job, we map these step outputs to job outputs called: `output1` and `output2`. In the `on.workflow_call.outputs` section we then define two outputs for the workflow itself, one called `firstword` which we map to `output1`, and one called `secondword` which we map to `output2`.  

The `value` must be set to the value of a job-level output within the called workflow. Step-level outputs must first be mapped to job-level outputs as shown below.

For more information, see "[AUTOTITLE](/actions/using-jobs/defining-outputs-for-jobs#overview)" and "[For more information about setting outputs with `jobs.<job_id>.outputs`, see "[AUTOTITLE](/actions/using-jobs/defining-outputs-for-jobs#overview)."](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_calloutputs)"
```

APCBoston:
@jc-clark your latest suggestion works for me and I have accepted it!

jc-clark (approved):
Thanks for working with me on this on @APCBoston and for contributing to the docs! 🪄

PR #32599, opened 2024-04-19
Update expressions.md

### LINKED ISSUE #31708: Add clarity to fromJSON function for expressions in Actions
### Code of Conduct

GitHub Actions / Learn GitHub Actions / Expressions: https://docs.github.com/en/actions/learn-github-actions/expressions

In the section "Functions", under "fromJSON", the section "[Example returning a JSON data type](https://docs.github.com/en/actions/learn-github-actions/expressions#example-returning-a-json-data-type)" currently looks like this:

> This workflow uses `fromJSON` to convert environment variables from a string to a Boolean or integer.
> 
> ```yaml
> name: print
> on: push
> env:
>   continue: true
>   time: 3
> jobs:
>   job1:
>     runs-on: ubuntu-latest
>     steps:
>       - continue-on-error: {% raw %}${{ fromJSON(env.continue) }}{% endraw %}
>         timeout-minutes: {% raw %}${{ fromJSON(env.time) }}{% endraw %}
>         run: echo ...
> ```
> 
> This example workflow sets environment variables: `continue` is set to a boolean value `true`, `time` is set to an integer value `3`.
> 
> The workflow uses the `fromJSON()` function to convert the environment variable `continue` from a string to a boolean, allowing it to determine whether to continue-on-error or not. Similarly, it converts the `time` environment variable from a string to an integer, setting the timeout for the job in minutes.

The paragraph right below the code block is confusing, because it implies that `continue` and `time` have non-string types, when they're really strings. Both the sentence before the code block and the last sentence make that more clear.

Really, the type of everything set in `env` is a string, and similar for everything set as a step output. For example, if you set a value like `myval='true'`, with literal quotes, like this

```sh
echo "myval='true'" >> "$GITHUB_OUTPUT"
```

and then use `fromJSON` like this:

```yaml
${{ fromJSON(myval) == true }}
```

the expression evaluates as false, because it's comparing `'true' == true`, a string to a boolean.

These are considered true, though:

```yaml
${{ fromJSON(myval) == 'true' }}
${{ myval == 'true' }}
```

This can be very confusing when you set a boolean somewhere (but really, it's a string!), and then you have a condition based on that. For example:

```yaml
      - id: set
        run: |
          echo "myval=false"  >> "$GITHUB_OUTPUT"

      - name: Conditional step
        if: '!steps.set.outputs.myval'
        run: echo "I'd expect this to run"
```

I'd reasonably expect the conditional step to be run: `myval` is false, and I use logical NOT, so it should be true. What really happens is that I'm taking the *string* false, and a string seems to be considered truthy, so NOT a string is false, and the step gets skipped.

To add to the confusion, further up in the same document, at [Operators](https://docs.github.com/en/actions/learn-github-actions/expressions#operators) it says that for equality operations, strings are coerced to `NaN` if they're' not a legal JSON number format, but according to the linked [MDN doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN), `NaN` is falsy, so `!NaN` should be true?

I have created a [workflow to test these equalities](https://github.com/bewuethr/no-v-test/blob/05dbce938ff6c2839ee5860a798b7b966666a7bc/.github/workflows/boolean.yml) and better understand them ([example run](https://github.com/bewuethr/no-v-test/actions/runs/7944591045/job/21690284216)).

---

tl;dr

1. The paragraph below the code block in the "Example returning a JSON data type" is confusing, can it be updated to be more clear about environment values always being strings?
2. Coercions should mention how different types interact with `!`, because the behaviour seems to contradict the `NaN` coercion described for equality comparisons.

### Additional information

_No response_

mark-mxwl:
### Why:

Closes: #31708

Deleted unnecessary and misleading line following `fromJSON` data types example. Revised line in Operators section of article, as suggested in the issue comments, to clarify behavior of `NaN` in loose-equality comparisons.

### Check off the following:

- [X] I have reviewed my changes in staging, available via the **View deployment** link in this PR's timeline (this link will be available after opening the PR).


nguyenalex836:
@mark-mxwl Thanks so much for opening a PR! I'll get this triaged for review ✨

nguyenalex836 (approved):

nguyenalex836:
@mark-mxwl These changes look great! ✨ I'll get this merged 🍏

mark-mxwl:
Thx @nguyenalex836!

kaysiehubbard (commented):
Ok my boy I'm going to be real with you you shit don't help nobody your shit is doring

PR #32765, opened 2024-05-01
Update about-self-hosted-runners.md

diroussel:
Don't advise allowing access to *.blob.core.windows.net as it allows access to all azure blobs in the world, for any action.  Better to allow access to only specific blob stores.

### Why:

Relates to: https://github.com/actions/actions-runner-controller/issues/3406

Previously, the docs said to only allow specific azure blob stores, but now they say to allow `*.blob.core.windows.net` which is much less secure. Any-bucket controlled by anyone, for any operation in the whole of Azure, worldwide, isn't exactly controlled. Surely it's enterprises paying for github enterprise that want selfhosted runners, and don't want to give unfettered access to build agents that this exists for?

Just a list of hostnames

### Check off the following:

### DIFF
--- a/content/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners.md
+++ b/content/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners.md
@@ -181,7 +181,26 @@ codeload.github.com
 
 ```shell copy
 results-receiver.actions.githubusercontent.com
-*.blob.core.windows.net
+productionresultssa0.blob.core.windows.net
+productionresultssa1.blob.core.windows.net
+productionresultssa2.blob.core.windows.net
+productionresultssa3.blob.core.windows.net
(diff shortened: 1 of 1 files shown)


nguyenalex836:
@diroussel Thank you again for submitting this PR 💛 Our engineering team provided the following response upon review -

> We today already publish a list of URLs specifically as part of the https://api.github.com/meta if [the customer] wishes to allow specific access and that list is updated dynamically over time as we add more storage accounts to the system. This is on the customer though to be sure to evaluate their own firewall rules and what they do and don't want to allow for their own needs.

For this reason, we won't be accepting this contribution, but greatly appreciate the time you've taken on this PR! If you're looking for another issue to pick up, take a look at our [help wanted section](https://github.com/github/docs/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) to find open issues you can work on ✨

PR #33793, opened 2024-07-02
Fix regression in artifact-attestations-step-for-container-images.md

snorremd:
### Why:

Fixes regression where the reusable artifact attestations step example was accidentally broken by a commit that replaced dashes with asterisks.

Commit that broke example: https://github.com/github/docs/commit/0886a399f07aedae418d587873ffc6c0810ae7eb

Swapped an asterisk in a GitHub actions step example with a dash.

### Check off the following:


snorremd:
This seems to be a bit more complicated to solve than I initially thought.

1. The attestation step reusable file seems to be the only file that contains only YML.
2. The markdown linter configured for the project will throw a fit if you use a dash `-`
3. The templating system does not appear to recognise other extensions than `md`, so renaming the file to use the `yml` extension is no good

Not really sure how to solve this, so closing my PR and will add an issue instead so that the core contributors can fix it.

PR #33842, opened 2024-07-07
Update misleading note in section-defining-outputs-for-jobs.md

piotrekkr:
### Why:

Because note about handling outputs when using strategy matrix is misleading and wrong in multiple places. Here is my overview of what is wrong:

1. > `$GITHUB_OUTPUT` is shared between all steps in a job. If you use the same output name in multiple steps, the last step to write to the output will override the value
   
   This is simply not true. Each step has it's own unique outputs file that it can write to. Next steps cannot override previous step outputs which is even show in usage example.
2. > If your job uses a matrix and writes to `$GITHUB_OUTPUT`, the content will be overwritten for each matrix combination.
   
   Using `$GITHUB_OUTPUT` does not override any job outputs. It can only set or override step outputs. Job need to define its own outputs and it can use `steps` context for this but this is not required for job outputs definition. Author of this note probably mixed step outputs with job outputs. Each job outputs inside matrix are combined, this is true, but author forgot to mention that last finished job in matrix will add new and override matching job outputs from all previous jobs.

3. > You can use the `matrix` context to create unique output names for each job configuration. For more information, see "[AUTOTITLE](/actions/learn-github-actions/contexts#matrix-context)."

   There is no way to use `contexts` within job outputs names. I tested it and always had an error from GitHub when trying to run workflow. Something like this is not allowed:
   ```
   jobs:
     job1:
       runs-on: ubuntu-latest
       outputs:
         output_${{ matrix.job }}: ${{ steps.gen_output.outputs.output_1 }}
       strategy:
         matrix:
           job: [1, 2, 3]

   ```

I've changed the note content and added an example of how matrix job outputs are combined. 

### Check off the following:


nguyenalex836:
@piotrekkr Thanks so much for opening a PR! I'll get this triaged for review ✨

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
 ```
 
 {% endraw %}
+
+{% note %}
+
+**Note:** When using strategy matrix, job outputs will be combined from all jobs inside matrix. The last executed job outputs will override matching outputs from previous jobs within same matrix.
```
thyeggman, line 40:
```suggestion
Matrices can be used to generate multiple outputs of different names. When using a matrix, job outputs will be combined from all jobs inside the matrix.
```

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
 ```
 
 {% endraw %}
+
+{% note %}
```
thyeggman, line 38: This is a good example and deserves its own heading instead of just a note 👍 Perhaps "Using Job Outputs in a Matrix Job"

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
...
+jobs:
+  job1:
+    runs-on: ubuntu-latest
+    outputs:
+      output_1: ${{ steps.gen_output.outputs.output_1 }}
+      output_2: ${{ steps.gen_output.outputs.output_2 }}
+      output_3: ${{ steps.gen_output.outputs.output_3 }}
+    strategy:
+      matrix:
+        job: [1, 2, 3]
```
thyeggman, line 56: We should avoid overloading the term "job" to avoid confusion. Perhaps a practical example would be useful, maybe using strings?  I'll leave it up to you

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
...
+        id: gen_output
+        run: |
+          job="${{ matrix.job }}"
+          echo "output_${job}=${job}" >> "$GITHUB_OUTPUT"
+          if [[ $job -eq 3 ]]; then
+            # slow down job so it will hopefully finish executing last
+            sleep 10
+            # override output_1 param
+            echo "output_1=${job}" >> "$GITHUB_OUTPUT"
+          fi
```
thyeggman, line 68: I think this part is not needed in the example, as we would prefer for the example to be the base use case.

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
...
+      # Will show
+      # {
+      #   "output_1": "3",
+      #   "output_2": "2",
+      #   "output_3": "3"
+      # }
+      - run: echo '${{ toJSON(needs.job1.outputs) }}'
+```
+
+{% endraw %}
```
thyeggman, line 82: Add a `{% warning %}` block at the bottom with the message about overriding jobs in later steps.  Something like this would work:

```
Actions does not guarantee the order that matrix jobs will run in. Ensure that the output name is unique, otherwise the last matrix job that runs will override the output value.
```

thyeggman (commented):
Good catch, thanks for the contribution! I think it's worth adding, if only for the fact that it's removing incorrect information.  The example should be the base use case though, as users might inadvertently copy it without knowing what it's supposed to do. The warning can be added separately 👍

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
piotrekkr: You mean something like this?
```
    strategy:
      matrix:
        tasks: ["generate-quote", "generate-order", "generate-email"]
```

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
piotrekkr: Not sure what should I do here. Should I remove `steps` section completely?

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
piotrekkr: Not sure if I understand it correctly. You suggest to create new separate heading to illustrate how using job outputs works with matrix and put almost all my change in there?

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
thyeggman: Something like we use in the [main matrix docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-a-matrix-for-your-jobs#using-a-matrix-strategy) would be good:

```yml
strategy:
  matrix:
    version: [10, 12, 14]
```

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
thyeggman: No, just remove the `if` block - I understand what you want to demonstrate, but it would be better to describe this in a warning block.

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
thyeggman: Yep, just create a new heading and then everything you added can go under it

piotrekkr: Done

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
piotrekkr: Added at the end as a warning.

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
piotrekkr: Condition removed.

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
piotrekkr: Done

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
...
+      output_3: ${{ steps.gen_output.outputs.output_3 }}
+    strategy:
+      matrix:
+        version: [10, 12, 14]
+    steps:
+      - name: Generate output
+        id: gen_output
+        run: |
+          job="${{ matrix.job }}"
+          echo "output_${job}=${job}" >> "$GITHUB_OUTPUT"
```
thyeggman, line 64:
```suggestion
          version="${{ matrix.version }}"
          echo "output_${version}=${version}" >> "$GITHUB_OUTPUT"
```

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
...
+jobs:
+  job1:
+    runs-on: ubuntu-latest
+    outputs:
+      output_1: ${{ steps.gen_output.outputs.output_1 }}
+      output_2: ${{ steps.gen_output.outputs.output_2 }}
+      output_3: ${{ steps.gen_output.outputs.output_3 }}
+    strategy:
+      matrix:
+        version: [10, 12, 14]
```
thyeggman, line 58: Changing this to match the output names

```suggestion
        version: [1, 2, 3]
```

data/reusables/actions/jobs/section-defining-outputs-for-jobs.md
```diff
@@ -40,3 +34,49 @@ jobs:
...
 
 {% endraw %}
+
+### Using Job Outputs in a Matrix Job
+
+{% note %}
+
+Matrices can be used to generate multiple outputs of different names. When using a matrix, job outputs will be combined from all jobs inside the matrix.
+
+{% endnote %}
```
thyeggman, line 44: No note block needed

```suggestion
Matrices can be used to generate multiple outputs of different names. When using a matrix, job outputs will be combined from all jobs inside the matrix.
```

thyeggman (commented):
Just a few more corrections to the workflow to match the new matrix strategy 👍

piotrekkr:
> Just a few more corrections to the workflow to match the new matrix strategy 👍

@thyeggman All is be done I think. Should I resolve discussions or should it be you?

thyeggman (approved):

nguyenalex836 (approved):

nguyenalex836:
@piotrekkr Thank you so much for working with @thyeggman on those suggestions! 💛 I'll update the branch and get this merged once tests are passing 🍏

PR #34338, opened 2024-08-18
Explicitly state that at least one job is required to run and that without, a workflow will fail

### LINKED ISSUE #34337: Explicitly state that at least one job is required for a workflow
### Code of Conduct

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-jobs-in-a-workflow

Slightly altering to something like:

A workflow run is made up of at least one or more `jobs`, which run in parallel by default. Without any jobs to run, a workflow will fail. To run jobs sequentially, you can define dependencies on other jobs using the `jobs.<job_id>.needs` keyword.

### Additional information

_No response_

MizouziE:
### Why:

Closes: #34337

### What's being changed:

The opening paragraph of the overview to:

A workflow run is made up of at least one or more `jobs`, which run in parallel by default. Without any jobs to run, a workflow will fail. To run jobs sequentially, you can define dependencies on other jobs using the `jobs.<job_id>.needs` keyword.

### Check off the following:

### DIFF
--- a/data/reusables/actions/jobs/section-using-jobs-in-a-workflow.md
+++ b/data/reusables/actions/jobs/section-using-jobs-in-a-workflow.md
@@ -1,4 +1,4 @@
-A workflow run is made up of one or more `jobs`, which run in parallel by default. To run jobs sequentially, you can define dependencies on other jobs using the `jobs.<job_id>.needs` keyword.
+A workflow run is made up of at least one or more `jobs`, which run in parallel by default. Without any jobs to run, a workflow will fail. To run jobs sequentially, you can define dependencies on other jobs using the `jobs.<job_id>.needs` keyword.
 
 Each job runs in a runner environment specified by `runs-on`.
 


nguyenalex836:
@MizouziE Thanks so much for opening a PR! I'll get this triaged for review ✨

subatoi:
Hi @MizouziE, and thanks for your PR—

I'm afraid we won't be accepting this submission on this occasion:

* The reusable you're proposing to update is only used in articles after where we'd expect this kind of disclaimer would be useful. For example, you can see under ["Workflow basics" in "About workflows"](https://docs.github.com/en/actions/writing-workflows/about-workflows#workflow-basics) that we say that a workflow "must contain" one or more jobs
* The language "one or more", in any case, already implies that a minimum of one should be present

Thank you for your interest in the GitHub docs!

MizouziE:
Thanks for the reply and explanation @subatoi I appreciate your points. Just wanted to mention that it's possible that people land directly on the pages which are beyond the points addressed in the `Workflow basics` chapters (as was my case) and, as is always the case with documentation, that leaves the reader with a little less context than ideal.

Anyway, lesson learned and hopefully I'll form a better habit of gathering context beyond the scope of what it is I'm attempting to do.

subatoi:
Thanks for your engagement @MizouziE—I've reached out to the internal team in question with your feedback, and they may choose to take it further and make some amendments.

Please don't be discouraged from raising further contributions—just as a gentle point, it may help to raise an issue first, so we can have this kind of discussion there first.

PR #34369, opened 2024-08-21
Update evaluate-expressions-in-workflows-and-actions.md

dmail:
### Why:

I want to have a dynamic path depending on `runner.os` in a GitHub workflow. See https://github.com/microsoft/playwright/issues/7249 for the context.

The documentation at https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions helped me. However even with all that documentation it took me a long time to make it work. I would like to add an example to help others that would face a similar use case.

Closes: Nothing

This pull request add an example where a value is a path dynamic per OS.

This example uses `format` and `env.HOME` to build the path on windows, something as follow:

```yml
path: format('{0}{1}', env.HOME, '\AppData\Local\ms-playwright')
```

This means something before have to set `env.HOME` which can be achieved like this:

```yml
- name: Put $HOME in env
  if: runner.os == 'windows'
  run: echo "HOME=$HOME" | Out-File -FilePath $env:GITHUB_ENV -Append
```

I don't know if this should be added to the documentation but if we don't add it people will have trouble to make it work.

I hope you'll find this useful, thank you.

### Check off the following:

### DIFF
--- a/content/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions.md
+++ b/content/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions.md
@@ -132,6 +132,17 @@ env:
 In this example, we're using a ternary operator to set the value of the `MY_ENV_VAR` environment variable based on whether the {% data variables.product.prodname_dotcom %} reference is set to `refs/heads/main` or not. If it is, the variable is set to `value_for_main_branch`. Otherwise, it is set to `value_for_other_branches`.
 It is important to note that the first value after the `&&` must be truthy. Otherwise, the value after the `||` will always be returned.
 
+An other example where a path is set depending on `runner.os`:
+
+{% raw %}
+
+```yaml
+env:
+  MY_ENV_VAR: ${{ runner.os == 'Windows' && format('{0}{1}', env.HOME, '\AppData\Local\ms-playwright') || runner.os == 'Linux' && '~/.cache/ms-playwright' || '~/Library/Caches/ms-playwright' }}
+```
+
+{% endraw %}
+
 ## Functions
 
 {% data variables.product.prodname_dotcom %} offers a set of built-in functions that you can use in expressions. Some functions cast values to a string to perform comparisons. {% data variables.product.prodname_dotcom %} casts data types to a string using these conversions:


nguyenalex836:
@dmail Thanks so much for opening a PR! I'll get this triaged for review ✨

nguyenalex836:
@dmail Thank you very much for your patience while our SME team reviewed your PR! ✨ Our SME team provided the following feedback on why we are unfortunately unable to accept this PR:

> > I would like to add an example to help others that would face a similar use case.
> 
> I feel that this document is intentionally devoid of specific use-cases unless those use-cases are illustrative and educational in some way. I came to this conclusion by comparing to other examples on this page:
> - [Immediately preceding example for operators](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example)
>   - The example uses placeholder names and GitHub/Git specific concepts like refs that are applicable and understandable (hopefully) to everyone working with GitHub writing actions workflows.
> - [Object filters](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#object-filters)
>   - The example presented is abstract using fruits as example values to illustrate the concepts using common object data names and qualities that we all understand well.
> 
> In both examples, the use-case of the example is not the focus; instead, the focus to demonstrate the workflow syntax using common ground the reader is likely familiar with.
> 
> Comparing these examples to the proposed changes, I see these differences:
> - The example does not use placeholder names and involves subject matter very specific to the use-case (playwrite paths & checking the runner.os context variable).
> - The example is not exclusively illustrative of actions workflow syntax concepts; instead it demonstrates this user’s use case.
> 
> What could be changed to make this PR acceptable?
> - Remove over-specific subject matter (playwrite) and replace with generic or (mostly) universally subject matter.
> - Add a written explanation, walking the reader through what is happening in the example. See how the [immediately preceding example for operators](https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example) does this. If this isn’t possible or seems unnecessary (if we are just repeating the explanation from the previous example), then it suggests that this example isn’t actually bringing any new learning to the table beyond the introduction of third-party/external subject matter like playwrite and the overall use-case.

We greatly appreciate your time putting together this PR, and if you have any feedback regarding the response our SME team provided, please let us know! 💛

PR #34401, opened 2024-08-26
Update telling-git-about-your-signing-key.md

alexmazaltov:
add alternative command for fish shell

### Why: cause it is not obvious how to deal with gpg signature using fish shell

Closes: 

### Check off the following:


nguyenalex836:
@alexmazaltov Thanks so much for opening a PR! I'll get this triaged for review ✨

alexmazaltov:
@nguyenalex836 thank you for your time

alexmazaltov:
@nguyenalex836 please review

subatoi:
Hi @alexmazaltov, and thanks for your submission and interest in the docs—

As a rule, we target Bash. I think we can make that fact clearer in our contributing documentation and I'll discuss that internally. There are potentially many more places other than this one where we could add examples that are specific to other shells, and unless there's a very specific reason to call one out in particular, we wouldn't make an exception, so for that reason I'm afraid we won't be accepting this submission right now.

I'd recommend opening an issue in future first for a better experience all round, just so we can offer feedback on how worthwhile a change or addition is before you go ahead and raise a PR. Alternatively, you can take one of the issues that are already marked as help wanted.

Thank you again!

SonuG56 (commented):
SonuG56

Issue #34529, opened 2024-09-07
Mini-TOC shows inline instead of on the side


hsnabszhdn:
### Code of Conduct

- [X] I have read and agree to the GitHub Docs project's [Code of Conduct](https://github.com/github/docs/blob/main/.github/CODE_OF_CONDUCT.md)

### What article on docs.github.com is affected?

[About Actions Runner Controller](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller)

### What part(s) of the article would you like to see updated?

The mini-TOC on this page is shown inline:

![image](https://github.com/user-attachments/assets/08df3e4d-994a-4e68-8444-5101d6a542b4)

But, in other pages of the site, it's on the right:

![image](https://github.com/user-attachments/assets/e9606dc5-f1dd-4ea4-b9f8-adcd70c7e766)

### Additional information

It's caused by `layout: inline` [here](https://github.com/github/docs/blob/25bf7a4d6ecea16ad1ba67109a165f8ba99b1c96/content/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller.md?plain=1#L5), and it will be fixed if removed.

However, it seems to have been added [here](https://github.com/github/docs/commit/50e20f57b49dcba22aea0a3d8c4d9ff83cadb26c) explicitly for a reason, so wasn't sure if I should raise a PR without discussing it first.

nguyenalex836:
@hsnabszhdn Thank you for raising this issue! ✨ We checked in with the original authors of this doc, and they stated `layout: inline` was deliberate to avoid a table rendering poorly 💛 

Thank you very much for discussing with us before raising the PR! We appreciate your contributions to our community 🙇

PR #34451, opened 2024-08-30
Update_access_control

Swastik-Swarup-Dash:
updated docs for easier understanding About access control

PR #34533, opened 2024-09-07
Update inline#34529

Swastik-Swarup-Dash:
Removed layers=inline maybe that raised that issue


nguyenalex836:
Hi @Swastik-Swarup-Dash 👋 

- As discussed [here](https://github.com/github/docs/issues/34529#issuecomment-2338414781), `layout: inline` was deliberate to avoid a table rendering poorly 💛

- As we discussed [here](https://github.com/github/docs/pull/34451#issuecomment-2324808550), unfortunately since your remaining changes are mainly personal preferences on how to phrase something, rather than corrections or improvements, we won't be able to accept this contribution.

Swastik-Swarup-Dash:
hi @nguyenalex836  thats cool i totally understand thank u for ur response

Issue #32988, opened 2024-05-16
Update docs for "Organizing information with collapsed sections" to include GitHub Pages support


jrzyshr:
### Code of Conduct

- [X] I have read and agree to the GitHub Docs project's [Code of Conduct](https://github.com/github/docs/blob/main/.github/CODE_OF_CONDUCT.md)

### What article on docs.github.com is affected?

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections

### What part(s) of the article would you like to see updated?

The sample Markdown displayed in the article will fail to render properly on a GitHub Pages website generated with Jekyll:

````markdown
<details>

<summary>Tips for collapsed sections</summary>

### You can add a header

You can add text within a collapsed section. 

You can add an image or a code block, too.

```ruby
   puts "Hello World"
```

</details>
````
**Problem:** This Markdown will result in the Jekyll renderer (Kramdown or GHMarkdown) treating all content in the Markdown file after the first `<summary/>` tag being rendered as plain text.  The webpage will hide the plain text in a collapsible section.  If the user clicks on the text in the `<summary>` tag on the GH Pages webpage, it will expand revealing the plain text from the source Markdown file in the GitHub repo. 

**Solution:** Adding a `markdown="1"` attribute to the `<details>` tag and a `markdown="span"` attribute to the `<summary>` tag will solve this issue.

I discovered the solution in this issue from the Kramdown repo: https://github.com/gettalong/kramdown/issues/155#issuecomment-1024896918

I propose either the sample in the document be updated as per the block below, OR a note added afterwards calling out this scenario when GH Pages is used with your Markdown.

````markdown
<details markdown="1">

<summary markdown="span">Tips for collapsed sections</summary>

### You can add a header

You can add text within a collapsed section. 

You can add an image or a code block, too.

```ruby
   puts "Hello World"
```

</details>
````

### Additional information

To reproduce the issue:
1. Create a Markdown file with the content from the sample snippet from this page in the docs
2. Enable GH Pages to render a site from the branch where your Markdown file is.
3. View the corresponding HTML page in the GH Pages site.

nguyenalex836:
@jrzyshr Thanks so much for opening this issue! I'll get this triaged for review ✨

subatoi:
Hi @jrzyshr—many thanks for opening an issue and explaining in such detail.

[...]

PR #34567, opened 2024-09-12
Add note on gh-pages for formatting within `<details>` block

nisbet-hubbard:
### Why:

Markdown formatting doesn’t render properly for gh-pages.

### Check off the following:


subatoi:
Hi @nisbet-hubbard, could I ask which renderer you're using? Is it Jekyll? Many thanks.

subatoi:
Thank you—I think this might cover the same issue: https://github.com/github/docs/issues/32988

Because of 32988, we've got an issue in our backlog that we'd like to cover internally. The solution we agreed on (that is, adding product callouts) will affect quite a few different pages when we implement it. 

I'll copy @nguyenalex836 in case I've missed something, but I'm afraid we'll need to close this in favour of our internal issue (I'll raise its priority). Thank you for your interest in the Docs and helping other users 🙏

PR #34635, opened 2024-09-18
docs: add alternative to bfg on about-large-files-on-github.md

MathiasBaumgartinger:
With all due respect, I found BFG really clunky to work with. Hence, I searched for an alternative. Credits to ZelluX (https://stackoverflow.com/a/1186549).

### Why:

BFG is a mighty tool, but may have to much overhead for a simple solution concerning big files in a previous commit. 

Closes: 

Just the documentation on large files and how to handle them in order to be able to push to github.

### Check off the following:

### DIFF
--- a/content/repositories/working-with-files/managing-large-files/about-large-files-on-github.md
+++ b/content/repositories/working-with-files/managing-large-files/about-large-files-on-github.md
@@ -104,7 +104,37 @@ If the file was added with your most recent commit, and you have not pushed to {
 
 ### Removing a file that was added in an earlier commit
 
+#### BFG Repo-Cleaner
+
 If you added a file in an earlier commit, you need to remove it from the repository's history. To remove files from the repository's history, you can use the BFG Repo-Cleaner or the `git filter-repo` command. For more information see "[AUTOTITLE](/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository)."
+
+#### git rebase 
+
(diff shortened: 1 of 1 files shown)


subatoi:
Hi @MathiasBaumgartinger and thanks for your interest in the GitHub docs 👋 

I'm afraid we won't be accepting this contribution on this occasion—I'll outline our reasoning below:

* There are a few places where we'd need to make this change, and it requires a more considered approach than only adding the content you've proposed to one article.

* Generally, we take an "opinionated" approach to content where there are multiple possible solutions, and pick one method and stick to it. That's something that we agree with stakeholders internally at time of publication, and once the content is published, unless there's something actually wrong about using that method and that feedback is surfaced, then we won't change it. 

For cases like this, where there isn't an immediately straightforward solution, I'd really recommend opening an issue first before taking time to write a PR. That way we can have a discussion first, and everybody involved—most importantly, you—has a better experience.

I will pass on feedback to the content team in question for any future updates they may make to this article.

Thanks again for your interest in the GitHub docs.

MathiasBaumgartinger:
Hey there!

I accidentally got caught in the same trap today *again*. I really cannot understand why you would recommend BFG for a solution that can honestly be solved in a matter so much simpler. For the sake of all other clumsy fellows out there, I would really appreciate you consider exposing this option in your documentation.

newren:
Hi,

The alternative text suggested here might work great for some users, but it doesn't provide warnings about when it is applicable (only if the history is linear, only if the large files are contained to one branch, and as the instructions are written, only if the large file was only touched in a single commit) which could be very deleterious for other users.  See #34950 for more details.  Also, as promised there, we do have a pending change that will mention interactive rebase, but with the caveats listed.

Docs team: It's a little

PR #34828, opened 2024-10-03
Update changing-your-github-username.md

dewantawsif:
I tried using a repository name that existed on the previous username owner and got the below message which leads me to believe that this isn't possible anymore.

image

Since the previous username owner has actually deleted the repository I wasn't able to confirm if the redirection still works.

### DIFF
--- a/content/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-user-account-settings/changing-your-github-username.md
+++ b/content/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-user-account-settings/changing-your-github-username.md
@@ -72,7 +72,7 @@ After you change your username, {% data variables.product.product_name %} will a
 * Web links to your existing repositories will continue to work. This can take a few minutes to complete after you make the change.
 * Command line pushes from your local repository clones to the old remote tracking URLs will continue to work.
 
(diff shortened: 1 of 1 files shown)


nguyenalex836:
@dewantawsif Thanks so much for opening a PR! I'll get this triaged for review ✨

nguyenalex836:
@dewantawsif Thank you for your patience while our SME team reviewed! They explained the following - 

> We are unable to accept this change. The part being removed is still accurate. 
> 
> If the new owner of the username creates a repo with a name that the old user used, it will mask the redirect, as the docs say. Yes, we now remove the redirect, but that is still functionality equivalent to "masking" it. 
> 
> This user may be making assumptions about the redirect behavior based on their inability to actually create a repo with a previously used name, but that restriction is based on how used the previous repo was, which is an entirely orthogonal concern. 
> 
> So, no they can't create a repo with the name they want, but if they could (based on the previous repo having little to no activity) the statement about the redirect would still hold.

For this reason, we will be closing this PR 💛 

Thank you very much for your willingness to contribute to GitHub Docs! If you're looking other issues to pick up, take a look at our [help wanted section](https://github.com/github/docs/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) to find open issues you can work on 🙇

PR #34833, opened 2024-10-04
Add context information for `jobs.<job_id>.uses`

lachieh:
### Why:

As far as I can tell, this property does not have access to any context and needs to be a statically defined string. If this is incorrect, it'd be great to have some clarification here.

Add row for `jobs.<job_id>.uses` into context availability table. Alternatively, a note at the top of this section could be used instead saying indicating that if the key is not listed in the table, it does not/may not support expression evaluation from context data.

As a third option, a note about this behavior could be added to the [relevant section on the workflow syntax page](https://docs-34833-5eac20.preview.ghdocs.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses).

### Check off the following:

### DIFF
--- a/content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
+++ b/content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
@@ -88,7 +88,7 @@ Different contexts are available throughout a workflow run. For example, the `se
 
 In addition, some functions may only be used in certain places. For example, the `hashFiles` function is not available everywhere.
 
-The following table indicates where each context and special function can be used within a workflow. Unless listed below, a function can be used anywhere.
+The following table lists the restrictions on where each context and special function can be used within a workflow. The listed contexts are only available for the given workflow key, and may not be used anywhere else. Unless listed below, a function can be used anywhere.
 
 | Workflow key | Context | Special functions |
 | ---- | ------- | ----------------- |
--- a/data/reusables/actions/reusable-workflow-calling-syntax.md
+++ b/data/reusables/actions/reusable-workflow-calling-syntax.md
@@ -3,4 +3,4 @@
 
 In the first option, `{ref}` can be a SHA, a release tag, or a branch name. If a release tag and a branch have the same name, the release tag takes precedence over the branch name. Using the commit SHA is the safest option for stability and security. For more information, see "[AUTOTITLE](/actions/security-guides/security-hardening-for-github-actions#reusing-third-party-workflows)."
 
-If you use the second syntax option (without `{owner}/{repo}` and `@{ref}`) the called workflow is from the same commit as the caller workflow. Ref prefixes such as `refs/heads` and `refs/tags` are not allowed.
+If you use the second syntax option (without `{owner}/{repo}` and `@{ref}`) the called workflow is from the same commit as the caller workflow. Ref prefixes such as `refs/heads` and `refs/tags` are not allowed. You cannot use contexts or expressions in this keyword.


nguyenalex836:
@lachieh Thanks so much for opening a PR! I'll get this triaged for review ✨

nguyenalex836:
@lachieh Thank you for your patience while our SME team reviewed! ✨ They responded with the following:

--------

I don't think this is the right place to convey this information.  We don't list where contexts cannot be used in the availability table, we only list where they can be used.  Right above that table we state:

> The following table indicates where each context and special function can be used within a workflow. Unless listed below, a function can be used anywhere.

So this is the part that could be clearer to me.  We have two things: contexts and functions.  In the table, the contexts column lists the only places where the given contexts can be used.  However, the functions column only lists the places where either 1) Functions cannot be used at all or 2) Only a specific set of functions can be used.  So unless that's clear to the reader going in they might have the wrong expectation.
Perhaps something like this to replace that would be a better option:

> The following table lists the restrictions on where each context and special function can be used within a workflow. The listed contexts are only available for the given workflow key, and may not be used anywhere else. Unless listed below, a function can be used anywhere.

--------

Would you be open to updating your PR to make the changes our SME team suggested instead? We'll be happy to merge this PR once that is finished 💛

LucifeMORNINGSA84:
As far as I can tell, this property does not have access to any context and needs to be a statically defined string. If this is incorrect, it'd be great to have some clarification here.
---
 .../accessing-contextual-information-about-workflow-runs.md

LucifeMORNINGSA84 (commented):
As far as I can tell, this property does not have access to any context and needs to be a statically defined string. If this is incorrect, it'd be great to have some clarification here.
  jobs.<job_id>.uses #7356

lachieh:
@nguyenalex836 Thanks for the review. I made the change to the text before the table, and removed the added row. I do feel that it is still not quite clear, as neither functions nor context appear to be usable in this key. The same restrictions apply in many other places, such as `defaults.run`. Because of that, I've copied [the sentence from that section](https://docs-34833-5eac20.preview.ghdocs.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrun:~:text=defaults.run.-,You%20cannot%20use%20contexts%20or%20expressions%20in%20this%20keyword.,-When%20more%20than) into [the section for `jobs.<job-id>.uses`](https://docs-34833-5eac20.preview.ghdocs.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses:~:text=are%20not%20allowed.-,You%20cannot%20use%20contexts%20or%20expressions%20in%20this%20keyword.,-Example%20of%20jobs). Happy to remove this change if needed.

nguyenalex836 (approved):

nguyenalex836:
@lachieh I had a messaged queued up, but forgot to send before I merged your PR!

> Because of that, I've copied [the sentence from that section](https://docs-34833-5eac20.preview.ghdocs.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrun:~:text=defaults.run.-,You%20cannot%20use%20contexts%20or%20expressions%20in%20this%20keyword.,-When%20more%20than) into [the section for jobs.<job-id>.uses](https://docs-34833-5eac20.preview.ghdocs.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iduses:~:text=are%20not%20allowed.-,You%20cannot%20use%20contexts%20or%20expressions%20in%20this%20keyword.,-Example%20of%20jobs).

I think this is a fair addition 💛 thanks again for your contribution! ✨

lachieh:
Thank you!

PR #35234, opened 2024-11-08
Informing a regex syntax should include required slash separators

nbenitez:
If we inform the user about how to make a case sensitive search using a regex syntax, do it with the regex delimitators in place, so the user:

- Don't gets an error when copying and pasting the regex syntax to the github code search textbox. Error meaning it didn't find results because it lacks the regex slash separators.

- Avoid the user to have to also find in the documentation how to enter a regex in the code search textbox. Have to find out it needs to use the slash separators.

So this MR is about removing burden from the user when trying to find out how to do a case sensitive search in Github. 

### Why:

Because a user checking docs of GitHub code search syntax to know how to search with case-sensitiveness enabled, would have to do so in two separate places (one where the required regex syntax is informed (i.e. `(?-i)`) and when pasting that and see it does not work, he would then have to find again in the documentation how a regex syntax is inputted (i.e. using two slash separators `/(?-i)/`).

### Check off the following:

### DIFF
--- a/content/search-github/github-code-search/understanding-github-code-search-syntax.md
+++ b/content/search-github/github-code-search/understanding-github-code-search-syntax.md
@@ -317,4 +317,4 @@ If code search guesses wrong, you can always get the search you wanted by using 
 
 ## Case sensitivity
 
-By default, code search is case-insensitive. Searching for `True` will include results for _uppercase_ `TRUE` and _lowercase_ `true`. You can do case-sensitive searches by using a regular expression with case insensitivity turned off, for example `(?-i)True`.
+By default, code search is case-insensitive. Searching for `True` will include results for _uppercase_ `TRUE` and _lowercase_ `true`. You can do case-sensitive searches by using a regular expression with case insensitivity turned off, for example `/(?-i)True/`.
--- a/content/search-github/github-code-search/understanding-github-code-search-syntax.md
+++ b/content/search-github/github-code-search/understanding-github-code-search-syntax.md
@@ -317,7 +317,7 @@ If code search guesses wrong, you can always get the search you wanted by using 
 
 ## Case sensitivity
 
-By default, code search is case-insensitive, and results will include both uppercase and lowercase results. You can do case-sensitive searches by using a regular expression with case insensitivity turned off. For example, to search for the string "True",  you would use:
+By default, code search is case-insensitive, and results will include both uppercase and lowercase results. You can do case-sensitive searches by using a regular expression with case insensitivity turned off. For example, to search for the string "True", you would use:
 
 ```text
 /(?-i)True/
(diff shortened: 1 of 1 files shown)


nguyenalex836:
@nbenitez Thanks so much for opening a PR! I'll get this triaged for review ✨

content/search-github/github-code-search/understanding-github-code-search-syntax.md
```diff
@@ -317,4 +317,4 @@ If code search guesses wrong, you can always get the search you wanted by using
 
 ## Case sensitivity
 
-By default, code search is case-insensitive. Searching for `True` will include results for _uppercase_ `TRUE` and _lowercase_ `true`. You can do case-sensitive searches by using a regular expression with case insensitivity turned off, for example `(?-i)True`.
+By default, code search is case-insensitive. Searching for `True` will include results for _uppercase_ `TRUE` and _lowercase_ `true`. You can do case-sensitive searches by using a regular expression with case insensitivity turned off, for example `/(?-i)True/`.
```
subatoi, line 320:
````suggestion
By default, code search is case-insensitive, and results will include both uppercase and lowercase results. You can do case-sensitive searches by using a regular expression with case insensitivity turned off. For example, to search for the string "True",  you would use:

```text
/(?-i)True/
```
````

**Review by subatoi — CHANGES_REQUESTED** [review]
Hi @nbenitez 👋 I agree we can make improvements here. We should use the opportunity to make this more consistent with the rest of the article, so I've made a suggestion to amend your PR accordingly. 

Thanks for your interest in the GitHub docs!

File: content/search-github/github-code-search/understanding-github-code-search-syntax.md [hunk 2]
```diff
@@ -317,4 +317,8 @@ If code search guesses wrong, you can always get the search you wanted by using
 
 ## Case sensitivity
 
-By default, code search is case-insensitive. Searching for `True` will include results for _uppercase_ `TRUE` and _lowercase_ `true`. You can do case-sensitive searches by using a regular expression with case insensitivity turned off, for example `(?-i)True`.
+By default, code search is case-insensitive, and results will include both uppercase and lowercase results. You can do case-sensitive searches by using a regular expression with case insensitivity turned off. For example, to search for the string "True",  you would use:
```
subatoi, line 320:
```suggestion
By default, code search is case-insensitive, and results will include both uppercase and lowercase results. You can do case-sensitive searches by using a regular expression with case insensitivity turned off. For example, to search for the string "True", you would use:
```

Just removing a bit of extra whitespace introduced by mistake

subatoi (dismissed):
Thank you! I'll get this merged now 👍

subatoi (approved):

PR #35276, opened 2024-11-13
Add search syntax tip for whole word searches

nbenitez:
Whole word searches are very useful in situations when a normal search gives two many false positives. Let's document how to do that in Github code search, so people don't have to look that up in regext tutorials.

### Why:

Because some times when receiving too many false positives matches, using whole word matching is a popular and effective way to narrow results down to your sought target.

### Check off the following:

### DIFF
--- a/content/search-github/github-code-search/understanding-github-code-search-syntax.md
+++ b/content/search-github/github-code-search/understanding-github-code-search-syntax.md
@@ -316,3 +316,11 @@ By default, code search is case-insensitive, and results will include both upper
 ```text
 /(?-i)True/
 ```
+
+## Whole Words
+
+If you want your search string to be matched as a whole word, you need to use a regular expression that includes the word boundary token `\b`. For example, to search for the string "True" as a whole word, you would use:
+
+```text
+/\bTrue\b/
+```


subatoi:
Hi @nbenitez and thanks for your submission 👋 

However, I'm afraid we won't be accepting this on this occasion: I'm not convinced about the particular benefit this new section would have, as we already describe the use of regular expressions in the same article.

Many thanks for your interest in the GitHub Docs and please feel free to take a look at our issues marked `help wanted`!

nbenitez:
Much like case sensitiveness, `Whole word` is a search feature present in most search dialogs (eg. try Adobe Reader, Firefox, etc..) for a reason, because it is both useful and used commonly to limit false positives.

Try eg. search for the [matio library](https://github.com/tbeu/matio) inside Flathub organization -> ` org:flathub matio` will crash for too many results, then try with case sensitiveness activated -> `org:flathub /(?-i)matio/` will not crash but show and error box with message `Results are incomplete due to an internal timeout, consider retrying or refining your query.` that is cause because the word infor**matio**n contains **matio** and information appears in a lot of places including every code license.

So the solution is to use `Whole Word` -> `org:flathub /(?-i)\bmatio\b/`

And is not a coincidence that most search dialogs include that option instead of letting the user to learn regex to use that.

subatoi:
I understand your points; however, in this repo we're only able to help with issues in the context of the documentation, not the product(s) themselves. In this case, I'd recommend opening a discussion in `community` to provide your feedback: https://github.com/orgs/community/discussions

Thank you again!

PR #35327, opened 2024-11-18
Fix `Code security` feature descriptions

### LINKED ISSUE #35325: Wrong description of the `Settings / Security / Code security` feature
### Code of Conduct

Multiple articles.

Regarding the issue #35324 I noticed that in the GitHub docs the **Settings / Security / Code security** feature is refered as "Code security and analysis". This term occurs 32 times over 19 files in the docs. This should be updated to the correct name "Code security".

### Additional information

I could provide a PR to fix this issues.

akordowski:
### Why:
Fix and unify the `Code security` feature description matching the current descriptions in the GitHub menu.

Closes: #35325

### Check off the following:


akordowski:
**NOTE**: There are still muliple occurrences of the text "code security and analysis" (mostly in the metadata), but I was unsure if to change them accordingly.

akordowski:
Reopen PR

subatoi:
Hi @akordowski 👋 this turned out to be simpler than I initially thought. It transpires that all current GHES versions still use the "Code security and analysis" wording, but Dotcom (FPT and GHEC) versions use the new wording.

With that said, I'd be happy to accept this PR if the conflicts are fixed and versioning is applied appropriately. Before I fixed it myself internally, I wanted to give you that option.

The best way to handle versioning in this case would be to create a [feature-based versioning file](https://docs.github.com/en/contributing/writing-for-github-docs/versioning-documentation) versioned for `fpt` and `ghec` only, so that we have the option to edit it later when it is changed for a future version of GHES. Then, all instances that you've changed here would need to be changed to say `{% ifversion <FBV> %} "Code security" {% else %} "Code security and analysis" {% endif %}.

Let me know what you think—if you'd prefer not to do this, just let me know and I'll fix it internally and add you as co-committer.

Many thanks!

akordowski:
Hi @subatoi 👋 

> The best way to handle versioning in this case would be to create a [feature-based versioning file](https://docs.github.com/en/contributing/writing-for-github-docs/versioning-documentation) versioned for fpt and ghec only, so that we have the option to edit it later when it is changed for a future version of GHES.

I am not familiar with the porcess how to do it, so I would leave it up to you.

As there are a lot of occurrences of `Code security and analysis` and this might be a subject to change, I would also suggest to create a reusable and replace all occurrences with it. That would make future changes even simpler as the change must be done in one file. What do you think?

subatoi:
No problem @akordowski — I'll handle this internally 👍 

> As there are a lot of occurrences of Code security and analysis and this might be a subject to change, I would also suggest to create a reusable and replace all occurrences with it. That would make future changes even simpler as the change must be done in one file. What do you think?

We could certainly use a reusable if the language were the same, yes. For the moment, I'll handle the replacement without using a reusable, but we can revisit that idea afterwards (I'll come back to you when it's merged, though it's possible it won't be this week as we're going on a soft freeze for the US Thanksgiving week). 

In the meantime I'm afraid we'll have to close this PR, but thank you again for your help! I'm still happy to add you as a co-committer, since this branch will be useful as a starting point for me.

Issue #35714, opened 2024-12-19
Documentation confuses Runner Scale Set Controller with Actions Runner Controller


Nuru:
### Code of Conduct

- [X] I have read and agree to the GitHub Docs project's [Code of Conduct](https://github.com/github/docs/blob/main/.github/CODE_OF_CONDUCT.md)

### What article on docs.github.com is affected?

Pretty much everything under `actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller`.

### What part(s) of the article would you like to see updated?

There are numerous references to "Actions Runner Controller" (e.g. [About Actions Runner Controller](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller)). These references are frequently paired with information that does not apply to the ARC or the runners it controls, leading to massive confusion.

The statement "Actions Runner Controller (ARC) is a Kubernetes operator that orchestrates and scales self-hosted runners for GitHub Actions" is true and correct. However, there is, additionally, a [GHA Runner Scale Set Controller](https://github.com/actions/actions-runner-controller/tree/master/charts/gha-runner-scale-set-controller) (RSSC) which, while it also is a Kubernetes operator that orchestrates and scales self-hosted runners for GitHub Actions, operates differently and on a different kind of runner configuration, a Runner Scale Set. 

Much of the documentation purportedly about ARC is actually about RSSC and the Runner Scale Sets they control. The RSSC manages [Runner Scale Sets](), while the [ARC](https://github.com/actions/actions-runner-controller/tree/master/charts/actions-runner-controller) manages Runner Deployments and Runner Replica Sets. See [About ARC](https://github.com/actions/actions-runner-controller/blob/master/docs/about-arc.md), which is correct, versus [About ARC](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller), which is actually about RSSC.

[...]

PR #35715, opened 2024-12-19
Clarify use of labels and groups together

Nuru:
### Why:

- Saying "and/or" when in fact the requirement is "and" is misleading.
- Saying "ARC does not support multiple labels" in an article that explains how to use multiple labels with ARC is extra confusing. See #35714 for more details.

As correctly explained under [Using labels and groups to route jobs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow#using-labels-and-groups-to-route-jobs), when you combine groups and labels, the runner must meet both requirements to be eligible to run the job. Therefore, a statement like "GitHub looks for a runner that matches the job's runs-on labels and/or groups" is confusing and wrong. The runner must match the job's labels ***and*** groups and will not be selected if it only matches the job's labels ***or*** groups but not both. So all the "and/or" instances have been changed to "and".

Reference to ARC at the top of the article has been changed to a reference to Runner Scale Sets for accuracy. The Note has been upgraded to Important because the entire article does not apply if you are using Runner Scale Sets. 

### Check off the following:

- [x] A subject matter expert (SME) has reviewed the technical accuracy of the content in this PR. In most cases, the author can be the SME. Open source contributions may require a SME review from GitHub staff.
- [x] All CI checks are passing.


nguyenalex836:
@Nuru Thanks so much for opening a PR! I'll get this triaged for review ✨

jc-clark (approved):
This looks good to me! Thank you @Nuru for digging into this and improving the docs. 

While we are in a repo freeze right now. I think this would be good to merge once the freeze is lifted. 

cc @nguyenalex836

PR #35780, opened 2024-12-28
Update accessing-contextual-information-about-workflow-runs.md

lachieh:
### Why:

I needed to pass the repository URL to a 3rd party action and used the `github.repositoryUrl` property. It took some digging to find out why the operation was timing out, but I eventually stumbled across a blog post which says to basically just [stop using the `git://` protocol.](https://github.blog/security/application-security/improving-git-protocol-security-github/#git-protocol-troubleshooting)

Adding a note below the table to mention this restriction.

### Check off the following:

- [ ] A subject matter expert (SME) has reviewed the technical accuracy of the content in this PR. In most cases, the author can be the SME. Open source contributions may require a SME review from GitHub staff.
- [ ] All CI checks are passing.

### DIFF
--- a/content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
+++ b/content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
@@ -206,7 +206,7 @@ The `github` context contains information about the workflow run and the event t
 | `github.repository_id` | `string` | {% data reusables.actions.repository_id-description %} |
 | `github.repository_owner` | `string` | The repository owner's username. For example, `octocat`. |
 | `github.repository_owner_id` | `string` | {% data reusables.actions.repository_owner_id-description %} |
-| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. [(See note below)](/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#note-on-githubrepositoryurl-property) |
+| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. Note: The `git://` protocol has been retired for security reasons. This value should only be used for identification and cannot be used to perform any git operations. Replace with {% raw %}`${{ github.serverUrl }}${{ github.repository }}`{% endraw %} to give you the HTTPS URL for the current repository. |
 | `github.retention_days` | `string` | The number of days that workflow run logs and artifacts are kept. |
 | `github.run_id` | `string` | {% data reusables.actions.run_id_description %} |
 | `github.run_number` | `string` | {% data reusables.actions.run_number_description %} |
@@ -287,10 +287,6 @@ jobs:
       - name: Run PR CI
         run: echo "Running PR only CI"
 ```
-
-### Note on `github.repositoryUrl` property
-
-The `git://` protocol is disabled for security reasons. You will not be able to use the value of `github.repositoryUrl` to perform any git operations. Instead, you can use `${{ github.serverUrl }}${{ github.repository }}` to give you the HTTP URL for the current repository.
 
 ## `env` context
 
--- a/content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
+++ b/content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
@@ -206,7 +206,7 @@ The `github` context contains information about the workflow run and the event t
 | `github.repository_id` | `string` | {% data reusables.actions.repository_id-description %} |
 | `github.repository_owner` | `string` | The repository owner's username. For example, `octocat`. |
 | `github.repository_owner_id` | `string` | {% data reusables.actions.repository_owner_id-description %} |
-| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. |
+| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. [(See note below)](./#note-on-github-repositoryUrl-property) |
 | `github.retention_days` | `string` | The number of days that workflow run logs and artifacts are kept. |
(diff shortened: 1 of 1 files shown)


content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
```diff
@@ -206,7 +206,7 @@ The `github` context contains information about the workflow run and the event t
 | `github.repository_id` | `string` | {% data reusables.actions.repository_id-description %} |
 | `github.repository_owner` | `string` | The repository owner's username. For example, `octocat`. |
 | `github.repository_owner_id` | `string` | {% data reusables.actions.repository_owner_id-description %} |
-| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. |
+| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. [(See note below)](./#note-on-github-repositoryUrl-property) |
```
lachieh, line 209:
```suggestion
| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. [(See note below)](/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#note-on-github-repositoryUrl-property) |
```

nguyenalex836:
@lachieh Thanks so much for opening a PR! I'll get this triaged for review ✨

content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
```diff
@@ -206,7 +206,7 @@ The `github` context contains information about the workflow run and the event t
 | `github.repository_id` | `string` | {% data reusables.actions.repository_id-description %} |
 | `github.repository_owner` | `string` | The repository owner's username. For example, `octocat`. |
 | `github.repository_owner_id` | `string` | {% data reusables.actions.repository_owner_id-description %} |
-| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. |
+| `github.repositoryUrl` | `string` | The Git URL to the repository. For example, `git://github.com/octocat/hello-world.git`. [(See note below)](/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#note-on-githubrepositoryurl-property) |
```
felicitymay, line 209: Suggest that we put the information from your note here directly.

```suggestion
| `github.repositoryUrl` | `string` | **Retired for security reasons.** The URL for the repository with the `git://` protocol, for example, `git://github.com/octocat/hello-world.git`. Replace with {% raw %}`${{ github.serverUrl }}${{ github.repository }}`{% endraw %} to give you the HTTPS URL for the current repository. |
```

content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
```diff
@@ -288,6 +288,10 @@ jobs:
         run: echo "Running PR only CI"
 ```
 
+### Note on `github.repositoryUrl` property
+
+The `git://` protocol is disabled for security reasons. You will not be able to use the value of `github.repositoryUrl` to perform any git operations. Instead, you can use `${{ github.serverUrl }}${{ github.repository }}` to give you the HTTP URL for the current repository.
+
```
felicitymay, line 294: Thanks for the clear note. In this case, I think that users will find the information more easily if we put the information directly in the table, as suggested above.

felicitymay (commented):
Many thanks for reporting back on your discoveries about the `git://` protocol and proposing a useful alternative for users. It's much appreciated 💖 

I'm suggesting that we move the information directly into the table. However, before we merge any updates, we should follow up internally because it feels as if the output of `github.repositoryUrl` should have been updated when we turned off the `git://` protocol.

content/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs.md
lachieh: Thanks, I moved the comment inline with some changes.

lachieh:
> it feels as if the output of `github.repositoryUrl` should have been updated when we turned off the `git://` protocol.

I'm going to assume that the output was not modified because it can still be used for string comparison despite it not working for git operations. It seems like it is [still used a fair bit](https://github.com/search?q=%22github.repositoryUrl%22+lang:yaml+path:/%5E%5C.github%5C/workflows%5C//&type=code). 

In most cases, the `actions/checkout` action is enough as it configures the runner with the correct remote URL anyway. I needed the https URL for a different 3rd-party action and had to dig a bit to work out why the `github.repositoryUrl` wasn't working specifically for git operations. I just didn't want others to waste time following the same path.

lachieh:
Thanks, @felicitymay!

> it feels as if the output of `github.repositoryUrl` should have been updated when we turned off the `git://` protocol.

I'm going to assume that the output was not modified because it can still be used for string comparison despite it not working for git operations. It seems like it is [still used a fair bit](https://github.com/search?q=%22github.repositoryUrl%22+lang:yaml+path:/%5E%5C.github%5C/workflows%5C//&type=code). 

In most cases, the `actions/checkout` action is enough as it configures the runner with the correct remote URL anyway. I needed the https URL for a different 3rd-party action and had to dig a bit to work out why the `github.repositoryUrl` wasn't working specifically for git operations. I just didn't want others to waste time following the same path.

nguyenalex836:
@lachieh Thank you so much for your patience while our SME team reviewed! ✨ After review, our SME team wanted to relay the following:

> I don't think we need a note here, we can just change the url to `https://github.com/octocat/Hello-World.git`, instead of explaining that `git` is no longer favored.

Would you be willing to update your PR to match the guidance from our SMEs? 💛

lachieh:
No, to change the URL in the docs means that they will be incorrect since the current `github.repositoryUrl` still uses the `git://` protocol. If the output of `github.repositoryUrl` is going to change then I suggest that the GitHub team handle the documentation update with whatever the usual breaking change process is.

PR #36538, opened 2025-02-26
Warn that `runs.pre` doesn't work from local actions

### LINKED ISSUE #34029: `runs.pre` should mention that it doesn't work from local actions or when called from other actions called from local actions
### Code of Conduct

https://github.com/github/docs/blob/main/content/actions/creating-actions/metadata-syntax-for-github-actions.md#runspre

`runs.pre` should include a paragraph explaining availability caveats.

It should mention:
1. `pre` doesn't work from local actions
2. If you create an action that will be used by other actions that might in turn be used by local actions, you'll trigger warnings in all workflows that use _that_ action.
3. that better practice involves creating a secondary action containing the `pre` phase so it can _optionally_ be used at the discretion of consumer actions/workflows.

### Additional information

* @zijchen opened an issue in https://github.com/actions/typescript-action/issues/564 and referenced this documentation.
* @ncalteen struggled to explain the problem to the various people who complained about the confusing behavior.

Documentation here would have helped since at least @zijchen would have known that it wasn't going to work.

Documentation here might have also discouraged people from adding `pre` to actions.

Instead, the right behavior which should be recommended in resolving this ticket is to provide a _second_ action in the same repository responsible for the `pre` behavior.

Thus, instead of @MoChilia in https://github.com/Azure/login/pull/384 changing `azure/login` to have a `pre` thing which triggers warnings as noted in https://github.com/actions/runner/issues/3397, @MoChilia could have created `azure/login/clear-credentials` which does the work that's currently included in `pre`. Then users could do:

- uses: azure/login/clear-credentials@...
- uses: azure/login@...
- uses: azure/login@...

And groups like microsoft/playwright could `uses: azure/login@...` without triggering dozens of warnings.

Fwiw. I'm pretty sure I've run across other repositories causing similar headaches in the past. It's really frustrating, and given that it appears some actions authors really do refer to documentation, documentation is the right place to put advice to avoid this problem.

jsoref:
### Why:

- Closes: #34029

### Check off the following:

### DIFF
--- a/content/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions.md
+++ b/content/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions.md
@@ -172,6 +172,9 @@ runs:
 
 **Optional** Allows you to run a script at the start of a job, before the `main:` action begins. For example, you can use `pre:` to run a prerequisite setup script. The runtime specified with the [`using`](#runsusing-for-javascript-actions) syntax will execute this file. The `pre:` action always runs by default but you can override this using [`runs.pre-if`](#runspre-if).
 
+> [!NOTE]
+> `runs.pre` is not supported for local actions.
+
 In this example, the `pre:` action runs a script called `setup.js`:
 
 ```yaml


subatoi (approved):
Thank you, @jsoref—and thanks for your patience with this one. I'll get this merged for you now 👍

PR #36735, opened 2025-03-10
Link to choosing the runner for a job

jsoref:
### Why:

- Closes: #36649

### Check off the following:

- [ ] A subject matter expert (SME) has reviewed the technical accuracy of the content in this PR. In most cases, the author can be the SME. Open source contributions may require an SME review from GitHub staff.
- [ ] The changes in this PR meet [the docs fundamentals that are required for all content](http://docs.github.com/en/contributing/writing-for-github-docs/about-githubs-documentation-fundamentals).
- [ ] All CI checks are passing and the changes look good in the review environment.

Issue #36649, opened 2025-03-06
`About billing for GitHub Actions` should link to `Standard GitHub-hosted runners for public repositories`


jsoref:
### Code of Conduct

- [x] I have read and agree to the GitHub Docs project's [Code of Conduct](https://github.com/github/docs/blob/main/.github/CODE_OF_CONDUCT.md)

### What article on docs.github.com is affected?

https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions

### What part(s) of the article would you like to see updated?

Possibly have each of `Linux`, `Windows`, and `macOS` in https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates-for-standard-runners link to the various sections of [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories)

### Additional information

Once I've decided that I _can_ do a thing based on pricing, the next logical step is figuring out _how_ to do that thing.

- Here, I've decided, after #36648 that I _can_ run a Linux arm runner, but I have no idea how.

A link to [Choosing the runner for a job: Per-minute rates for standard runners](https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories) would tell me that the answer is one of: [ubuntu-24.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md), [ubuntu-22.04-arm](https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-22-image.md)

jc-clark:
I think the connection here could be clearer. However, I don't think we need to add links throughout the tables. What do you think about updating: 

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the account's plan. Any usage beyond the included amounts is controlled by spending limits.

to 

[...]

PR #36757, opened 2025-03-11
Replace GitHub with data variable

jsoref:
### Why:

- In #36735 I copied some text from a recommendation and unfortunately I missed restoring a data variable reference

### Check off the following:


subatoi (dismissed):
Thanks @jsoref, I've also added a missing full stop. I also think that given this is in a reusable, it would be better with an abbreviated linking sentence as in my suggestion, and removing the new blank line between the paragraphs. Otherwise, the risk is that there's too much text on some pages where this reusable is used, before the user gets to the "good stuff" so to speak.

Would you be happy to make that further change (deleting the blank line)?

jsoref:
@subatoi I was just copying content from https://github.com/github/docs/issues/36649#issuecomment-2712062135 and _trying_ to adjust the text to match the github data stuff.

I have no opinion. But, please feel free to apply whichever changes and merge.

subatoi (approved):

PR #37421, opened 2025-04-09
Fix typo in URL

MatthewSteeples:
The original URL returns a 404, the new one goes to the correct page


Sharra-writes:
Thanks for finding and fixing that! I'll get this merged shortly. 💛

Sharra-writes (approved):

PR #38445, opened 2025-05-21
Correct duplicate branch-paths-filter.md paragraph

### LINKED ISSUE #38444: There is a duplicate paragraph in `events-that-trigger-workflows.md`
### Code of Conduct

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#push

The duplicated portions are the two "Note" blocks:

https://github.com/github/docs/blob/b64f820ddb0bd5c8a07f59201cc1930933b83277/content/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows.md?plain=1#L865-L875

https://github.com/github/docs/blob/b64f820ddb0bd5c8a07f59201cc1930933b83277/content/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows.md?plain=1#L903-L913

What I believe is the original intent (based on the surrounding context) is that the second block was originally going to say that "If you use both the *`tags`* filter and the `paths` filter, the workflow will only run when both filters are satisfied.", as opposed to mentioning the `branches` twice. Assuming that this is in fact correct, I've created a PR that updates the documentation to say so in this [PR](https://github.com/github/docs/pull/38445).

### Additional information

> Are you able to reliably reproduce the problem?

Yes, just go to https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#push

> How often does it occur?

Everytime.

> How many users are affected?

All GHA users that are reading the documentation for the `push` event.

> Add any other context or screenshots about the feature here.

The documentation for the `push` event in `events-that-trigger-workflows.md` has a duplicate paragraph. I've tried to fit both duplicates in one screen shot, hopefully it's clear.

Image

*Edit: add the PR link*

norwd:
### Why:

Closes: https://github.com/github/docs/issues/38444

I've reworded the second of two duplicate paragraphs to match the surrounding context. Namely, instead of saying *specifying both `branches: [...]` and `paths: [...]` means both conditions must be satisfied* twice, the second time it talks about `tags: [...]` instead (since it's under the `Running your workflow only when a push of specific tags occurs` heading).

### Check off the following:


Sharra-writes:
@norwd Looking at this more closely, isn't the duplicate note under "[Running your workflow only when a push affects specific files](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files)"? The tags section is the one above. I'm going to check in with our SMEs about the necessity for having a note there at all, just to make sure it's needed and to keep this from getting held up more, but I thought I would point that out in case they agree that the note in that section needs to be included and updated. 💛

Sharra-writes:
@norwd Okay, word from the SMEs is: only one note is necessary. Simplest fix would be to delete the second one. If, however, you would rather create an example with tag, path, and branch, update the reusable, and delete the first note instead, that was also a solution they proposed as acceptable.

norwd:
> ...Simplest fix would be to delete the second one...

Seems, fair enough, I'll make that update now 👍

Sharra-writes:
Perfect! I'll get this added to the merge queue once checks are passing.

Sharra-writes (approved):

PR #38522, opened 2025-05-24
Update using-your-github-profile-to-enhance-your-resume.md

Zearin:
- Distinguish “résumé” from “resume”
- Add link break before unordered list for consistency with other lists in the document

### Why:

- Typographical obsession
- Consistency

Closes: N/A

(This is a _very_ minimal edit. I don’t think it needs its own Issue.)

Spelling and whitespace.

### Check off the following:


Sharra-writes:
Thank for opening a PR! I appreciate the attention to detail, and "resume" as opposed to "résumé" is one that always makes my eye twitch slightly when I'm trying to decide how to write it. Here's the issue in this case: the title has to match the file name, which also affects the url. Since the accented e isn't a standard character, that creates problems. I'm guessing that since "resume" had to be used in the title, they went ahead and used it throughout the document for consistency.

I can check in about it and see what the team thinks about changing it in the body, but I'm going to defer to your preferences here, because I know I also personally get twitchy when things aren't used consistently. 😅 That blank line you added is fine. If you would rather open a separate PR for that, I can get it approved while I'm looking into everything else on this one. Again, up to you.

PR #38685, opened 2025-06-02
docs: fixes in Actions runner documentation

Rahul-Lashkari:
### Why: 
Fixes in the GitHub Actions documentation related to self-hosted runners and the Actions Runner Controller (ARC).

Closes: 

### Check off the following:


Sharra-writes:
Thanks for opening a PR! The use of binary vs. binaries seems really inconsistent to me in the original. "Binaries" is used once more in that same paragraph and once in the following paragraph, but then on line 106 it switches to "binary." 😵‍💫 I'm going to check in with the rest of the team about that, but if you can make it make sense, I would welcome the explanation from any source. 😅

Rahul-Lashkari:
> Thanks for opening a PR! The use of binary vs. binaries seems really inconsistent to me in the original. "Binaries" is used once more in that same paragraph and once in the following paragraph, but then on line 106 it switches to "binary." 😵‍💫 I'm going to check in with the rest of the team about that, but if you can make it make sense, I would welcome the explanation from any source. 😅

thanks for the feedback, I felt the same and waited for your saying. I also felt the word `scale set` would be much appropriate. I've updated the pr, hope this makes it! PTAL.

content/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller.md
```diff
@@ -18,21 +18,21 @@ topics:
 
 {% data reusables.actions.actions-runner-controller-about-arc %}
 
-The following diagram illustrates the architecture of ARC's autoscaling runner scaleset mode.
+The following diagram illustrates the architecture of ARC's autoscaling runner scale set mode.
```
Sharra-writes, line 21: Okay, this is also complicated. I think "ScaleSet" is correct, even though it's written as "Scale Sets" in one place, because I see names like "AutoScalingRunnerSet" and "RunnerSet" that follow the same format. The "scaleset" is more ambiguous, though, because "scale set" is used in the reusable on line 19. Lowercase "scaleset" might be the name of the mode, but it's not used that way consistently.

Here's what I would recommend: "ScaleSet" should always be one word, because it matches the other capitalized terms. Lowercase "scaleset" should always actually be "scale set." That *should* (🤞) make it consistent and more readable.

Rahul-Lashkari: > Okay, this is also complicated. I think "ScaleSet" is correct, even though it's written as "Scale Sets" in one place, because I see names like "AutoScalingRunnerSet" and "RunnerSet" that follow the same format. The "scaleset" is more ambiguous, though, because "scale set" is used in the reusable on line 19. Lowercase "scaleset" might be the name of the mode, but it's not used that way consistently.
> 
> Here's what I would recommend: "ScaleSet" should always be one word, because it matches the other capitalized terms. Lowercase "scaleset" should always actually be "scale set." That _should_ (🤞) make it consistent and more readable.

PTAL :)

Rahul-Lashkari: Thanks for the recommendation, please review~~

content/actions/hosting-your-own-runners/managing-self-hosted-runners-with-actions-runner-controller/about-actions-runner-controller.md
```diff
@@ -91,7 +91,7 @@ Once the custom resources are installed, you can deploy ARC into your Kubernetes
 
 ### About the runner container image
 
-{% data variables.product.company_short %} maintains a [minimal runner container image](https://github.com/actions/runner/pkgs/container/actions-runner). A new image will be published with every runner binaries release. The most recent image will have the runner binaries version and `latest` as tags.
+{% data variables.product.company_short %} maintains a [minimal runner container image](https://github.com/actions/runner/pkgs/container/actions-runner). A new image will be published with every release of the runner binaries. The most recent image will have the runner binaries version and `latest` as tags.
```
Sharra-writes:
```suggestion
{% data variables.product.company_short %} maintains a [minimal runner container image](https://github.com/actions/runner/pkgs/container/actions-runner). A new image will be published with every runner binaries release. The most recent image will have the runner binaries version and `latest` as tags.
```

I don't think rephrasing this adds much, it's just using more words to say the same thing. I didn't catch it in my last review, so I'll just commit the suggestion so we can move forward.

Sharra-writes (approved):

Sharra-writes:
Okay, I think this is good! I'll get it added to the merge queue once checks are passing.

PR #38735, opened 2025-06-04
doc: remove prefix "Premium" of GPT-4.1 to reduce confusion

satomic:
### Why:

- While other models do not have the "Premium" prefix, the prefix "Premium" of GPT-4.1 has caused a lot of misunderstanding, like:
  -  Most people who look at this table will understand that there are 2 different models: the (Base) GPT-4.1 as the Base Model and the Premium GPT-4.1 as the Premium Model.
  - Therefore, some customers have asked me "Where can I find the setting to enable the Premium version of GPT-4.1?" 
 
So, please remove this prefix so that it is easier to understand.

- `about-premium-requests.md`

### Check off the following:

### DIFF
--- a/content/copilot/managing-copilot/monitoring-usage-and-entitlements/about-premium-requests.md
+++ b/content/copilot/managing-copilot/monitoring-usage-and-entitlements/about-premium-requests.md
@@ -47,7 +47,7 @@ Each model has a premium request multiplier, based on its complexity and resourc
 | Model                                                                   | Premium requests                                                             |
 |-------------------------------------------------------------------------|------------------------------------------------------------------------------|
 | Base model (currently {% data variables.copilot.copilot_gpt_41 %}) [^2] | 0 (paid users), 1 ({% data variables.copilot.copilot_free_short %}) |
-| Premium {% data variables.copilot.copilot_gpt_41 %}                     | 1                                                                            |
+| {% data variables.copilot.copilot_gpt_41 %}                             | 1                                                                            |
 | {% data variables.copilot.copilot_gpt_4o %}                             | 1                                                                            |
 | {% data variables.copilot.copilot_gpt_45 %}                             | 50                                                                           |
 | {% data variables.copilot.copilot_claude_sonnet_35 %}                   | 1                                                                            |


Sharra-writes:
Thanks for opening a PR! This does seem weird. Let me check and see if there's a reason for "premium" to be there, or if it was just an oversight when the table last got updated.

Sharra-writes:
@satomic Okay, I looked into this and that "premium" is there very deliberately. It's not that the model is premium, it's that the requests can be either basic or premium, as the article discusses. If users want to enable the "premium version of GPT 4.1," they're fundamentally misunderstanding what premium means in this context, and haven't understood any of the rest of the article. There's a reason the table isn't at the top. The rest of the article is important context. If there's something that you think would make the article clearer, that would be worth talking over, but I'm afraid taking away the word "premium" will do the opposite of reducing confusion for anyone who understands the context.

satomic:
@Sharra-writes I think t“Your first instinct tells you that this is a bit Wired” shows that this place is indeed prone to misunderstanding😂

Because we are all Microsoft insiders, we know the answer and can make the right judgment after some thinking times. But assuming we are ordinary customers, according to this table, the Model column in the table is obviously the name of the model that is actually optional in Copilot (in fact, only Premium GPT-4.1 is not), so naturally, customers will think that "Premium GPT-4.1" is also an optional model, which causes misunderstanding. that's why some customers asked me about that.

Of course, I agree that it is worth adding a little explanation here, but I think a better way is to add a superscript `[3]` and explain it in more detailed text in small print below. Instead of directly adding a prefix like "Premium".

image

felicitymay:
Hi @satomic - thanks for raising this PR and opening a discussion on how to make this clearer ✨ 

This is an area of the product and docs that are undergoing rapid change at the moment so I'm going to add a subject matter expert (SME) label which will trigger our internal review process. We'll get back to you as soon as we're able because this is important to get right.

sophietheking:
@satomic Thanks for your patience and contributing to our docs, we really appreciate it. 🙏 We've made some changes to that section of that article and the word "premium" has been removed.

PR #38913, opened 2025-06-14
Update about-forks.md: move a non-difference out of the list of diffe…

### LINKED ISSUE #38912: about-forks: move a non-difference out of the list of differences
### Code of Conduct

https://docs.github.com/en/contributing/collaborating-on-github-docs/about-contributing-to-github-docs

In the list of differences between forks and duplicate repositories, there is one bullet point that doesn't belong there because it holds for both, forks and duplicates. Listing it as a difference suggests that duplicates didn't have their own members, branches, tags, labels, policies, issues, pull requests, discussions, actions, projects, and wikis - which is inaccurate and may be confusing to new or unknowing users.

The point should be separated from the list of differences. See PR https://github.com/github/docs/issues/38913.

### Additional information

_No response_

GfEW:
…rences.

Listing that set of properties of forks as a difference to duplicate repositories suggests that the latter didn't have their own members, branches, tags, labels, policies, issues, pull requests, discussions, actions, projects, and wikis - which is inaccurate and may be confusing to unknowing users.

### Why:

Closes: https://github.com/github/docs/issues/38912

Rearranged existing content to resolve a confusing inconsistency.

### Check off the following:


Sharra-writes:
Thanks for opening a PR! That list item definitely looks out of place, but I would advise just deleting it rather than moving it. We're trying to move toward a more streamlined style in our "about" pages, and that just isn't essential information. If we're listing the differences between forked and duplicated repositories, then it can probably be assumed that things not listed aren't differences, and users can go read more in depth about forks to confirm, if necessary.

GfEW:
For a beginner audience seeking to learn the basics about forks, that sentence is full of essential information -- it's just not a difference. I think the clarity it provides about forks **(*)** should be given _somewhere_ at "About forks", since that's the obvious choice out of all pages listed at [Working with forks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks) . Where else would users expect to find those basics?

Also, strictly forks aside, hinting at those properties being shared with dupes is a shortcut to understanding both. It prevents users from having to search and learn about those properties of forks and dupes separately, to only find out then by manual comparison how much they have in common.
That's the point of the adverbial phrase "Like duplicated repositories".

How about moving the sentence further up on the page?

_**(*)** (N. b.: If you haven't worked with forks (or even git) before and wonder how their members, branches, tags, labels, policies, issues, pull requests, discussions, actions, projects, and wikis, relate to upstream, those are real questions.)_

Sharra-writes:
@GfEW Okay, I looked a little more and this is apparently the only place this information is stated explicitly, so...yeah, you're right. Maybe more of the articles about forks should be revised to put it in an article where it makes more sense, but that's not going to happen today. This is fine. 👍 I'll get it added to the merge queue once checks are passing.

Sharra-writes (approved):

PR #39342, opened 2025-07-16
Update using-copilot-code-review.md

kaushganguly:
Visual Studio supports GitHub Copilot Code Review as stated in the docs.

### Why:

Closes: 

### Check off the following:


Sharra-writes:
@kgangulyvibe Okay, so...this one is weird because the versioning is complicated (maybe even overly complicated). I think because it's under "web browser" it's just trying to tell you what you can and can't do in a web browser, not trying to give you a comprehensive list of other places you can do that thing. If you want to know where you can do that thing, there are other versions available at the top of the page. 

I do think it's confusing, though, (*only* available in VS Code? is that really worth specifying when, even if true, it could change at any moment?) and my preference would be to just have it labeled "not available on the GitHub website" or "available on the GitHub website." I don't know why we're bringing VS/VS Code into it at all.

Let me check in and make sure there's not a reason for doing it this way, though, and also that this article isn't already being completely changed. We're doing a bunch of reorganization in the docs, and Copilot is one of the first sections getting picked apart.

Sharra-writes:
@kgangulyvibe I've confirmed that this article has been completely rewritten and is in the process of being approved right now, so I'm going to close this. Thank you for bringing it to our attention, though. 💛

PR #39695, opened 2025-08-06
Update using-yaml-frontmatter.md

vit100-trader:
### Why:

**title** tag is required according to this code:

https://github.com/github/docs/blob/f584b8f16867a48d2268deb733acce68484299d4/src/frame/lib/frontmatter.js#L40

* The `title` field in YAML frontmatter is now marked as **required** instead of optional, making it clear that every page must specify a human-friendly title.


Sharra-writes:
@vit100-trader Looks good! Thanks for catching that. I'll get this added to the merge queue once checks are passing.

Sharra-writes (approved):

PR #39702, opened 2025-08-06
Corrected search index limitations explanation

ruchira-net:
Updated the limitations section to remove references to public repositories deleted before the index was created as it's wrong to say deleted code before the index was created will be NOT be included. It will be included is the right thing to say and it's already explained in the sentence after that.

### Why:

Closes: 

### Check off the following:

### DIFF
--- a/content/copilot/concepts/completions/code-referencing.md
+++ b/content/copilot/concepts/completions/code-referencing.md
@@ -78,7 +78,7 @@ Code in private {% data variables.product.prodname_dotcom %} repositories, or co
 
 ## Limitations
 
-The search index is refreshed every few months. As a result, newly committed code, and code from public repositories deleted before the index was created, may not be included in the search. For the same reason, the search may return matches to code that has been deleted or moved since the index was created.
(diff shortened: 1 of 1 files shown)


Sharra-writes:
@ruchira-net Thanks for opening a PR! This reads to me more like hedging than a declaration (note the "*may* not be included"), just in case some results are missing for whatever reason. The next sentence also uses hedging language ("*may* return matches"). I will check in about the phrasing, but the way it's punctuated definitely reads like it was added after the fact. The sentence is actually pretty awkward, and should probably be revised even if that clause is left in.

Sharra-writes:
@ruchira-net I've confirmed with the Copilot team that the information is accurate, but also written in a way that confuses them briefly, too. Essentially, the index takes a snapshot of things as they are at that moment. If you search for something deleted before that snapshot was taken, it's probably not going to appear in your search results, because the snapshot shows that it was deleted. If, however, you search for something deleted after the snapshot was taken, the snapshot still shows it's there, and so it probably will appear in your search results. 

It is honestly a weird case to include, and they way it's inserted haphazardly makes me think that someone complained at some point about deleted code or repositories not showing up, and so that clause was inserted to explain why. I'm going to close this PR and instead try to make the paragraph clearer.

PR #40296, opened 2025-09-08
It didn't really work when the README.md file was linked like that, hopefully now it works?

HibouDevv:


Sharra-writes:
@HibouDevv Thanks for opening a PR! Unfortunately this isn't a file we can accept contributions for, but I would like to know if you got a 404 when you tried the link, because it works fine for me in its current form.

hiboudev1:
Ok

PR #40367, opened 2025-09-13
Make blocked user info more consice

SollyBunny:
The issue for this would be the same as the PR and take longer

Combined `you no longer x's on users` and `users on longer x's on a` into one phrase using `eachother`. This happens to contribution counting and colaborator status. I then put these next to eachother.

I'm not sure if I can use `collaborator status` since it's not used elsewhere, it would make it easier to read in my opion rather than `You are removed as collaborators from`

I am also not sure if I can change `counted as contributions` to just `counted`, I initally wanted to change it to `counted on eachothers' profiles` since I wasn't aware of anywhere else contributions were counted

### Check off the following:

~~- [ ] All CI checks are passing and the changes look good in the review environment.~~ Wording change


rajpoot538:
shoaib

Sharra-writes:
@SollyBunny Thanks for opening a PR! I'm afraid, though, that you're sacrificing clarity for being concise, and that is absolutely not what we want to do in any documentation about safety. Clarity is the top priority. We're not going to accept this PR, but we do have a [help wanted](https://github.com/github/docs/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) tag in the issues if you're looking to make a contribution. A lot of articles beginning with "about" ([About projects](https://docs.github.com/en/issues/planning-and-tracking-with-projects/learning-about-projects/about-projects?platform=mac), for example) could also be much more concise, if that's the kind of editing you're specifically looking to do. ✨

PR #40399, opened 2025-09-16
Update authorizing-oauth-apps.md

lopo12123:
### Why:
I found that this request should be GET instead of POST, so I made this change

Closes: 

### Check off the following:

### DIFF
--- a/content/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps.md
+++ b/content/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps.md
@@ -79,7 +79,7 @@ If the user accepts your request, {% data variables.product.github %} redirects 
 
 Exchange this `code` for an access token:
 
-    POST {% data variables.product.oauth_host_code %}/login/oauth/access_token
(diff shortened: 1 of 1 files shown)


Sharra-writes:
@lopo12123 Thanks for opening a PR! Could you do me a favor and give me a little more context on why this change should be made? I need to be able to accurately summarize the problem for our SMEs when I go to ask for verification, and this is not my area of expertise. Any error messages you might have received would be helpful!

lopo12123:
This request needs to use GET, not POST. The documentation says POST, which is misleading, so I hope to correct the error in the description in a timely manner.
By the way, it was only after trying POST several times that I finally figured out how to use GET that it worked.

Sharra-writes:
@lopo12123 Our SMEs tell me that `GET` was supported in the OAuth 1.0a spec, but is no longer supported in the 2.0 spec. It may work for you because you're using an older spec, but `POST` is correct for the current spec.

m-balghonaim:
👋🏽 @lopo12123, thank you for helping us keep our docs up to date!

You should be able to use `POST` for this request. There might be a configuration issue with your `POST` request if it's not succeeding.

PR #40522, opened 2025-09-23
Update Dependabot options reference for clarity

cluse-heartflow:
Clarify behavior of 'increase-if-necessary' option in Dependabot.

### Why:

Originally I misread this. I thought this setting would make Dependabot ignore minor versions. I didn't understand that with this setting enabled, Dependabot will still make a PR, but it will only update the lockfile, not the package.json.

Closes: 

Just changing the description of the `increase-if-necessary` option for Dependabot so that it is more obvious that this does not cause Dependabot to ignore minor versions.

### Check off the following:


Sharra-writes:
@cluse-heartflow Thanks for opening a PR! This wording definitely looks opaque as-is. I checked with the Dependabot team, and they proposed this wording to be more concise:

> Leave the version requirement unchanged if it already allows the new release (Dependabot still updates the resolved version). Otherwise widen the requirement.

Do you think this is clearer, or is there a change you would like to make for better clarity? We're trying to keep it short since it appears in a table and we don't want a big block of text. That's unfortunately also a barrier to understanding. We could probably make it a little longer than this, though.

cluse-heartflow:
Sure, sounds good to me.

content/code-security/dependabot/working-with-dependabot/dependabot-options-reference.md
```diff
@@ -782,7 +782,7 @@ When `versioning-strategy` is defined, {% data variables.product.prodname_depend
 |--------|--------|
 | `auto` | Default behavior.|
 | `increase`| Always increase the minimum version requirement to match the new version. If a range already exists, typically this only increases the lower bound. |
-| `increase-if-necessary` | Leave the constraint if the original constraint allows the new version, otherwise, bump the constraint. |
+| `increase-if-necessary` | If the existing rule already allows the new version, Dependabot leaves the manifest file unchanged, but will still update the lockfile to use the latest version. If the existing rule does not allow the new version (for example, when a new major release falls outside the range), Dependabot will also update the manifest constraint. |
```
Sharra-writes:
```suggestion
| `increase-if-necessary` | Leave the version requirement unchanged if it already allows the new release (Dependabot still updates the resolved version). Otherwise widen the requirement. |
```

Sharra-writes (approved):

PR #40550, opened 2025-09-25
Replace outdated hardcoded GitHub SHA links for SARIF schemas with HTTP links

TWiStErRob:
### Why:

1. The links pointing to the specific SHA and file path are outdated. The files have been moved around. See https://github.com/oasis-tcs/sarif-spec/issues/646 for more details. Note: because the links were using specific SHA they were still working, they are just pointing to an old version. There is one actual meaningful difference (language regex is wrong).

Only the pointed links changed:
 * The link `SARIF specfication documentation` now points to the docs, not the schema. This is the same URL to where the `OASIS Standard` link points.
 * The schema links using SHA has been replaced to point to the link used in the spec.

### Check off the following:


Sharra-writes:
@TWiStErRob I've confirmed the links work, and I'm going to trust the community that the content in them is correct, since you know better than I do.

Sharra-writes (approved):

PR #40796, opened 2025-10-10
Update instructions for changing the default branch

abhiabhi94:
- relevant part was under the 'General' option on settings.

### Why:

Closes: 

### Check off the following:


Sharra-writes:
@abhiabhi94 Thanks for opening a PR! The reason "general" isn't listed is because it's the default option. Settings opens on "general," and all the user needs to do is scroll down. I don't think this one needs to be clarified, because adding extra detail just makes it more complicated to update if the UI changes slightly. 

We're not going to accept this change, but if you would like to find something to help with, our [help wanted](https://github.com/github/docs/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) label is a great place to start!

PR #40824, opened 2025-10-12
run-on-specific-branches-or-tags1: terminology fix

wh0:
### Why:

In this line, there isn't really such thing as "the undefined Git ref." Rather "defined" and "undefined" describe the `branches`/`tags` _filters_ when you define only one of them. 

Closes: 

### Check off the following:


Sharra-writes:
@wh0 Thanks for opening a PR! Copilot tells me this is correct, and I would be inclined to trust it since this is a small change, if it hadn't just completely made up a citation for me. So. I think I'm going to verify with an actual person.

data/reusables/actions/workflows/run-on-specific-branches-or-tags1.md
```diff
@@ -5,6 +5,6 @@ Use the `branches` filter when you want to include branch name patterns or when
...
 
-If you define only `tags`/`tags-ignore` or only `branches`/`branches-ignore`, the workflow won't run for events affecting the undefined Git ref. If you define neither `tags`/`tags-ignore` or `branches`/`branches-ignore`, the workflow will run for events affecting either branches or tags. If you define both `branches`/`branches-ignore` and [`paths`/`paths-ignore`](/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), the workflow will only run when both filters are satisfied.
+If you define only `tags`/`tags-ignore` or only `branches`/`branches-ignore`, the workflow won't run for events affecting the undefined filter. If you define neither `tags`/`tags-ignore` or `branches`/`branches-ignore`, the workflow will run for events affecting either branches or tags. If you define both `branches`/`branches-ignore` and [`paths`/`paths-ignore`](/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), the workflow will only run when both filters are satisfied.
```
larsxschneider, line 8: I believe `filter` is not right... but `Git ref` is not right either.

Because if `branches-ignore` is set, then the workflow is indeed affecting the undefined Git refs AFAIK.

> If you define only `tags`/`tags-ignore` or only `branches`/`branches-ignore`, the workflow won't run for events affecting the undefined filter.

Maybe we can remove that ☝️ sentence all together because we already discuss the filters in the previous paragraphs. I am not sure.

larsxschneider (changes requested):

data/reusables/actions/workflows/run-on-specific-branches-or-tags1.md
wh0: you mean if you set

```yml
on:
  push: 
    # tags not set
    # tags-ignore not set
    branches-ignore: nobuild
```

it will still run on all pushes to tags?

Sharra-writes:
@wh0 I couldn't find anyone to review this, unfortunately, so I'm going to close this for now. It's possible we'll be able to revisit it in the future.

PR #40863, opened 2025-10-15
Remove duplicate section on fail-fast

RobKenis:
The mention on fail-fast applying to the entire matrix is also included in the reusable block. This renders to the block being repeated in the documentation.

### Why:

This Pull Request removes duplication of the same documentation on the same page, even the same paragraph.


Sharra-writes:
@RobKenis Thanks for updating this article! I'll get it approved and added to the merge queue once checks are passing.

Sharra-writes (approved):

PR #40865, opened 2025-10-15
Use curly braces in if condition

glenthomas:
### Why:

GitHub lint says it's strongly recommended to always wrap the full `if:` condition in `${{ }}`. I think it would be best to include this in the example code to avoid leading people into unexpected issues.

See the [GH-Lint documentation](https://ghlint.twisterrob.net/issues/default/DoubleCurlyIf/) for an explanation.

### What's being changed (if available, include any code snippets, screenshots, or gifs):

Add braces to if.

### Check off the following:

- [x] A subject matter expert (SME) has reviewed the technical accuracy of the content in this PR. In most cases, the author can be the SME. Open source contributions may require an SME review from GitHub staff.
- [x] The changes in this PR meet [the docs fundamentals that are required for all content](http://docs.github.com/en/contributing/writing-for-github-docs/about-githubs-documentation-fundamentals).
- [x] All CI checks are passing and the changes look good in the review environment.

PR #40878, opened 2025-10-16
Fix example in control-jobs-with-conditions

glenthomas:
### Why:

My previous PR (https://github.com/github/docs/pull/40865) broke the example. The condition is now displaying as `$false`.

I didn't realise that the expression needs to be escaped.

Use the `{% raw %}` template directive (is this the right way to handle it?

### Check off the following:


Sharra-writes (approved):

Sharra-writes:
@glenthomas Thanks for coming back to fix it!

PR #41035, opened 2025-10-28
docs: clarify definition of GitHub Actions

DanielDerefaka:
## What
Clarified the definition of "action" in the Understanding GitHub Actions guide.

## Why
The previous definition incorrectly described actions as "a set of jobs or code," which is misleading. Actions are custom applications used as steps within jobs, not collections of jobs themselves.

## Changes
- Updated line 82 to accurately define actions as "custom applications for the GitHub Actions platform"
- Removed the inaccurate "set of jobs" phrasing

### DIFF
--- a/content/actions/get-started/understand-github-actions.md
+++ b/content/actions/get-started/understand-github-actions.md
@@ -79,7 +79,7 @@ For more information, see [AUTOTITLE](/actions/using-jobs).
 
 ### Actions
 
-An **action** is a custom application for the GitHub Actions platform that performs a complex but frequently repeated task. Actions can perform tasks such as:
+An **action** is a custom application for the GitHub Actions platform that performs a specific, repeated task in your workflow. Actions can perform tasks such as:
 
 * Pulling your Git repository from {% data variables.product.prodname_dotcom %}
 * Setting up the correct toolchain for your build environment
--- a/content/actions/get-started/understand-github-actions.md
+++ b/content/actions/get-started/understand-github-actions.md
@@ -79,7 +79,7 @@ For more information, see [AUTOTITLE](/actions/using-jobs).
 
 ### Actions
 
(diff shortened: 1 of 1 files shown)


content/actions/get-started/understand-github-actions.md
```diff
@@ -79,7 +79,7 @@ For more information, see [AUTOTITLE](/actions/using-jobs).
 
 ### Actions
 
-An **action** is a pre-defined, reusable set of jobs or code that performs specific tasks within a **workflow**, reducing the amount of repetitive code you write in your workflow files. Actions can perform tasks such as:
+An **action** is a custom application for the GitHub Actions platform that performs a complex but frequently repeated task. Actions can perform tasks such as:
```
Sharra-writes:
```suggestion
An **action** is a custom application for the GitHub Actions platform that performs a specific, repeated task in your workflow. Actions can perform tasks such as:
```

Updating the language slightly to be more general. The tasks don't *have* to be complex just because they usually are.

Sharra-writes:
@DanielDerefaka Thanks for the PR! This looks like a good update, with one little tweak to make it more general. We have a deploy freeze for GitHub Universe this week, but I'll get this merged once that's lifted.

rajlanjewar01 (commented):
Test

Sharra-writes:
@DanielDerefaka An engineer came and weighed in on this, and they aren't in favor of the change in definition, feeling that it would be confusing to use "application" when we also have GitHub and OAuth apps, and also that the term itself was too broad. It's possible the definition could use an update, but after their breakdown of what they didn't like about this definition, I can see why it is what it is. We would need one of our engineers to work with a writer to rewrite it so they could help pick and choose specific terminology that's not being used extensively elsewhere to mean other things.

All that is to say, I'm afraid I'm going to have to reject this one after all. I'm sorry about that! Some of the other people on my team are very familiar with Actions, so I thought it would be enough to ask them, but our actual engineers had the opposite opinion.

PR #41207, opened 2025-11-11
Update terminology from 'transient' to 'transitive'

yeikel:
### Why:

While both terms describe the same concept, “transitive dependency” is more commonly used, and the term “transient dependency” may lead to confusion for some audiences.

See https://en.wikipedia.org/wiki/Transitive_dependency and https://github.com/dependabot/dependabot-core/issues/13456

### Check off the following:


Sharra-writes (approved):

Sharra-writes:
@yeikel This is a pretty small change and Copilot agrees with you, so I'm going to accept it without trying to get a third opinion.

PR #41371, opened 2025-11-21
Normalize use of string in git command.

kakkeshyor:
"gpgSign" instead of "gpgsign" for both invocations: (a) avoids confusion, (b) matches use of camel case as in Git documentation.

### Why:

The invocation `git config --global x true` spells the string `x` in two different ways. The case is probably not important but to maintain parity with Git documentation the suffix of `x` should use camel case: `gpgSign`. Also, having two versions of the same suffix in the same block of text may confuse readers.

Closes: NA

Text formatting of this "reusable text block," located at `data/reusables/gpg/set-auto-sign.md`, specifying a git invocation that sets Git's global configuration to enable automatic commit-signing, is being changed.

### Check off the following:

### DIFF
--- a/data/reusables/gpg/set-auto-sign.md
+++ b/data/reusables/gpg/set-auto-sign.md
@@ -1,7 +1,7 @@
 1. Optionally, to configure Git to sign all commits and tags by default, enter the following command:
 
    ```shell
-   git config --global commit.gpgsign true
+   git config --global commit.gpgSign true
    git config --global tag.gpgSign true
    ```
 


Sharra-writes:
@kakkeshyor Thanks for opening a PR! This is a small change and I don't think the case matters, but I should point out that all the documentation I can find on the site writes it as `gpgsign`, so if your goal is consistency and avoiding user confusion, I don't think this will achieve either of those.

kakkeshyor:
As you wish.

PR #41726, opened 2025-12-07
Fix job definition syntax in migration tutorial

NenoLoje:
Fixed 1 minor issue in a code sample that would result in invalid YAML.

- No functional changes are introduced. 
- Fixed by removing an unnecessary colon from a job definition.


Sharra-writes:
@NenoLoje This is a pretty minor change, and Copilot says it's right, so I'm going to accept it.

Sharra-writes (approved):

PR #41868, opened 2025-12-15
docs: clarify member-only profile instructions

mbainter:
### Why:

The docs do specify a private repository already, but this adds a callout to emphasize that this doesn't mean "not public" but rather that it has to have visibility of private -- internal is not enough. It can be easy to think of internal as a different level of visibility for a private repository instead of a whole other type that affects the way other features interact with it.

Because of how we structure our organization I wanted our .github-private repo to be internal and for a while I could not figure out why the org profile wasn't showing. Since we had also moved to Enterprise Managed Users I defaulted to assuming it was connected to EMU and glossed over the visibility as a factor. Trying to do searches when troubleshooting this problem also didn't yield helpful results, partly because I assumed it was connected to EMUs. I almost moved on once I figured it out, but I suspect maybe I'm not the first or the last to be bitten by this.  I figure worst case people searching on it might land here.

I also considered rewording the documentation itself, but I think this is more a problem of ensuring people have the right mental modal for private vs internal. I don't know if that has the same impact anywhere else, so a small callout to help people avoid making my mistake seemed like the best solution to me.

Adding a callout to the member-only profile instructions

### Check off the following:


Sharra-writes:
@mbainter I think this is probably a good change, but I'm going to check in to make sure we want it here and dig in a bit to make sure the wording is what we would want it to be.

content/organizations/collaborating-with-groups-in-organizations/customizing-your-organizations-profile.md
```diff
@@ -56,6 +56,8 @@ The content of public `README.md` will appear on your organization's public prof
 
 ## Adding a member-only organization profile README
 
+>[!NOTE] Private organization profiles must have "Private" visibility, not "Internal".
```
guntrip:
```suggestion
>[!NOTE] Member-only organization profile READMEs will only load from a `.github-private` repository with its visibility set to private. Repositories in enterprises set to internal visibility will not work.
```

guntrip (approved):
Hi @mbainter! Thanks for opening this. I can see this being an easy point of confusion and agree we should add a note.

I've made a few changes, committed them, and I'll now get this merged 🎉

PR #42621, opened 2026-01-29
Update Docker actions to use tags instead hashes

szabgab:
### Why:

Those hashes were very old and in general it recommended (by other parts of the GitHub documentation) that people use tags to specify version of actions.

The versions used in the example.

### Check off the following:

### DIFF
--- a/content/actions/tutorials/publish-packages/publish-docker-images.md
+++ b/content/actions/tutorials/publish-packages/publish-docker-images.md
@@ -196,13 +196,13 @@ jobs:
         uses: {% data reusables.actions.action-checkout %}
 
       - name: Log in to Docker Hub
-        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
+        uses: docker/login-action@v3
         with:
           username: {% raw %}${{ secrets.DOCKER_USERNAME }}{% endraw %}
           password: {% raw %}${{ secrets.DOCKER_PASSWORD }}{% endraw %}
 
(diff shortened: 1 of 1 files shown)


Sharra-writes:
@szabgab Can you cite sources on parts of the documentation that recommend versions instead of SHA? I'm probably going to have to check in anyway since Copilot disagrees and it may be a matter of the advice being conditional rather than blanket, but if you have a source it might be useful if that needs to be updated.

szabgab:
I am sorry I don't have the link, I just vaguely remember I read it. Years ago.

szabgab:
OK, so this is not in the documentation, but the sample GitHub workflows seem to almost exclusively use tags.

I went to an arbitrary repository, tried to add a GitHub Action and picked 5 different workflows. I only found one case where a hash was used and even there only one action used a hash. The others used tags.

Most of them also used out-of date tags (`actions/checkout@v4` while it is now (`actions/checkout@v6`). Those should be also updated I guess.

Sharra-writes:
@szabgab Our team doesn't have control over what workflows do. We're writers, not engineers. Our docs are guided by engineers and what they tell us best practices are, or sometimes if they feel two approaches are more or less equal, we will choose one and use it for the sake of simplicity. I can take this to an engineer to discuss the merits of hashes vs. tags, but since changing all the docs to be consistent would probably be a considerable undertaking, I don't know that it would receive any priority, unless it includes concerns about security or functionality.

PR #42746, opened 2026-02-03
Fix grammar: pluralize user accounts

ritik35k:
### Why:

Closes: 

### Check off the following:


Sharra-writes:
@ritik35k I think the way it is might be technically correct, because the "are" refers to "types" rather than "account," but I also don't think it matters enough to go on a grammatical hunt for, so I'll just accept this because I think it reads more comfortably.

Sharra-writes (approved):

PR #42794, opened 2026-02-06
docs: add missing GHES versioning for Copilot IDE

saisravan909:
Closes #38826. Added GHES versioning tags to reflect compatibility with Enterprise Server.

### Why:

<[https://github.com/github/docs/issues/38826]>
Closes: 

<Updated the Liquid versioning tags in the Eclipse quickstart to include ghes visibility.>

### Check off the following:


Sharra-writes:
@saisravan909 Sorry, this isn't the problem we're having with this article. The versioning is fine as it is now. The problem is that trying to update the links so that they go directly where they're supposed to instead of redirecting breaks the versioning, and we don't know why. Honestly, though, we have redirects for a reason, it's fine, it works, there's no pressing need to update, so I'm just going to close the issue. I'm also going to close your PR, though, because it's not fixing what the issue is about. 

Please feel free to look around more, though. We get small chores that would be perfect updates for a new user pretty frequently. This one is just a mess.

PR #42917, opened 2026-02-12
Correct instructions to avoid 403 when Copilot CLI auths via PAT

DanielJonesEB:
You need to give the token "Models" permissions, otherwise when authing non-interactively with COPILOT_GITHUB_TOKEN you get a 403. There are a few open issues in the Copilot CLI repo about this, eg https://github.com/github/copilot-cli/issues/1316

### Why:

Closes: 

### Check off the following:


Sharra-writes:
@DanielJonesEB According to the engineer who responded to that issue, this is not intended behavior. We don't generally document behavior that is, in essence, a bug that needs to be fixed.

Sharra-writes:
@DanielJonesEB I talked to one of our engineers who confirmed that the docs should not be updated with this. I'm sorry, but I'm going to close the PR now.

PR #42934, opened 2026-02-13
docs: clarify custom domain with user/org site

jamietanna:
As a follow-up to support ticket 4079946 to make it more explicit.

### Why:

Closes: 

### Check off the following:


Sharra-writes:
@jamietanna Thanks for opening a PR! I don't have access to support tickets, and support usually won't tell me any specifics about them, so understand going in that I have no context for the change you're making. The problem here is that the sentence you're adding is not new information, it's just adding emphasis to what the sentence before already said, which isn't good writing practice. The emphasis should be added in the preceding sentence. Here are two ideas for doing that, and you can tell me if you think one of them is clear enough since you're the one who has context for all of this:

> By default, if you set a custom domain for a user site or organization site, that same custom domain will be used for all project sites owned by the same account.

I think having "by default" middle of the sentence breaks up the flow of ideas and makes it more difficult to understand, so moving it to the opening may help with clarity. Using the full terms "user site or organization site" adds a little more emphasis to those terms, hopefully drawing attention to them. It would be possible to emphasize them further, though, by saying:

> By default, if you set a custom domain for a **user site** or **organization site**, that same custom domain will be used for all project sites owned by the same account.

Do you think one of these options would answer your purpose?

jamietanna:
That's very reasonable, thanks! How does this proposed wording look?

Sharra-writes (approved):

Sharra-writes:
@jamietanna Looks good! Unfortunately we're now in the midst of a deploy freeze that's going to last an unspecified amount of time. I'll put a ready to merge label on this and merge it when I'm allowed to.

janellecanon01-lab (commented):
Fixes #44435 ###

Sharra-writes:
@jamietanna I'm so sorry, I can't believe I somehow missed this when I was going through the "ready to merge" label after the deploy freeze. I will get it merged as soon as all the checks are passing.

jamietanna:
Np, thank you!

PR #43082, opened 2026-02-22
Update events-that-trigger-workflows.md

### LINKED ISSUE #43081: workflow_dispatch Note should be amended regarding default branch
### Code of Conduct

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch

The workflow_dispatch section, specifically the first Note section. 

The Note states "This event will only trigger a workflow run if the workflow file exists on the default branch."

This is false--workflow_dispatch is absolutely able to be triggered when the workflow file is not in the default branch. The issue lies in HOW it can be triggered: when it is in the default branch, the Github UI can be used. When it is not, then the Github UI cannot be used. The "Note" regarding this should be updated, as it misleads the reader into thinking that workflow_dispatch is more limited than it is.

This should be changed to be more accurate. There are scenarios where a developer may not be able to merge to the repository default branch for whatever reason, yet they still need to be able to trigger/test a workflow. This change makes it clear that only the Github UI is unavailable to activate workflow_dispatch, not that the trigger itself cannot be activated.

### Additional information

Image

mcummings128:
### Why:

Closes: https://github.com/github/docs/issues/43081

workflow_dispatch is absolutely able to be triggered when the workflow file is not in the default branch. The issue lies in HOW it can be triggered: when it is in the default branch, the Github UI can be used. When it is not, then the Github UI cannot be used. The "Note" regarding this should be updated, as it misleads the reader into thinking that workflow_dispatch is more limited than it is.

Please also refer to the Issue linked above. 

### Check off the following:

### DIFF
--- a/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md
+++ b/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md
@@ -1107,7 +1107,7 @@ on: workflow_call
 | [workflow_dispatch](/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_dispatch) | Not applicable | Last commit on the `GITHUB_REF` branch or tag | Branch or tag that received dispatch |
 
 > [!NOTE]
-> {% data reusables.actions.branch-requirement %}
+> To trigger the workflow manually using the Github UI, the workflow file must exist on the default branch.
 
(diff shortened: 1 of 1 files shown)


subatoi:
Closing out per my comment in https://github.com/github/docs/issues/43081#issuecomment-5330726928 🙇

PR #43138, opened 2026-02-26
Enhance REST API documentation with conditional requests

JustinGrote:
Added information about conditional requests in the REST API.

### Why:

Closes: 

### Check off the following:

### DIFF
--- a/content/rest/about-the-rest-api/comparing-githubs-rest-api-and-graphql-api.md
+++ b/content/rest/about-the-rest-api/comparing-githubs-rest-api-and-graphql-api.md
@@ -82,3 +82,5 @@ mutation {
   }
 }
 ```
+
+The REST API also has the concept of conditional requests, which allow you to use ETAG and If-None-Modified headers to observe if the previous query has changed without incurring a rate limit query charge. For more info, see: [AUTOTITLE](/rest/using-the-rest-api/best-practices-for-using-the-rest-api#use-conditional-requests-if-appropriate)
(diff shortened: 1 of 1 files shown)


JustinGrote:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/github/docs/pull/43138#pullrequestreview-3857820602)

Sharra-writes:
@JustinGrote I need to make sure this is something we want documented, and want documented here. It's not clear how it relates to the rest of the article in terms of comparing the two APIs. If we do want it documented, it might be better in [About the REST API](https://docs.github.com/en/rest/about-the-rest-api/about-the-rest-api?apiVersion=2022-11-28).

JustinGrote:
@Sharra-writes with GraphQL there is no way to "poll" to see if something has changed without incurring rate limit. REST you can, so it's a compelling factor in making the decision between the two, hence why it should be included.

Sharra-writes:
@JustinGrote Gotcha. In that case it's going to need to be revised so that it's contrasted against GraphQL, rather than being dropped in without context. That's fine, I can figure that part out.

subatoi:
Hi @JustinGrote and thanks for your contribution and apologies for the late conclusion.

Unfortunately, we don't typically accept such contributions unless the scope has been agreed by a writer. For that reason I'd recommend opening an issue in this repo before undertaking any work.

Thanks for your interest in the GitHub docs—we hope to see you in future 🙇

Issue #42376, opened 2026-01-13
creating-an-issue doesn't talking about cross-referencing / explain how to cross-reference


jsoref:
### Code of Conduct

- [x] I have read and agree to the GitHub Docs project's [Code of Conduct](https://github.com/github/docs/blob/main/.github/CODE_OF_CONDUCT.md)

### What article on docs.github.com is affected?

https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/creating-an-issue

### What part(s) of the article would you like to see updated?

Talk about cross-referencing

### Additional information

https://docs.github.com/en/discussions/guides/best-practices-for-community-conversations-on-github#community-conversations-in-github-discussions

> For more information about opening an issue and cross-referencing a discussion, see [Creating an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue).

faizanbsc0-stack:
``import unittest
from generate_unit_tests import validate_price

class TestValidateProduct(unittest.TestCase):
  def test_price_zero(self):
    with self.assertRaises(ValueError) as context:
      validate_price(0)
    self.assertEqual(str(context.exception), "Price must be greater than 0")

  def test_price_negative(self):
    with self.assertRaises(ValueError) as context:
      validate_price(-10)
    self.assertEqual(str(context.exception), "Price must be greater than 0")

  def test_price_above_limit(self):
    with self.assertRaises(ValueError) as context:
      validate_price(1500)
    self.assertEqual(str(context.exception), "Price must be less than or equal to 1000")

  def test_price_edge_case_zero(self):
    with self.assertRaises(ValueError) as context:
      validate_price(0)
    self.assertEqual(str(context.exception), "Price must be greater than 0")

  def test_price_edge_case_max(self):
    try:
      validate_price(1000)
    except ValueError:
      self.fail("validate_price() raised ValueError unexpectedly!")

if __name__ == '__main__':
  unittest.main()

Sharra-writes:
@jsoref Yeah, I see your point. This is part of an old docs set, so no on will care too much what's done to it. I'm going to put a help wanted label on to see if someone wants to pick it up before I have time to get to it, since I'm going to have to find out more about cross-referencing, and someone else might already have that knowledge.

[...]

PR #43140, opened 2026-02-26
Clarify cross-referencing when creating an issue from a discussion

Aimaghsoodi:
## Summary
- clarify that using **Create issue from discussion** also creates a cross-reference between the discussion and the issue
- address the documentation gap described in #42376

## Notes
- docs-only change
- opened as a draft to keep review/process lightweight
- happy to close this if I should route it differently


Sharra-writes:
@Aimaghsoodi Sorry, this isn't what we're talking about in terms of cross-referencing. This is converting discussions into issues, which isn't the same thing, and which we do have documentation about.

PR #43151, opened 2026-02-27
Link to GitHub APIs in developer program documentation

schneidergithub:
Updated the integration content to link directly to GitHub's APIs.

### Why:
Updating documentation to link to the official API documentation.  Was missing from this page and saves a few clicks for devs.


Sharra-writes:
@schneidergithub We're not actually supposed to update this article, because at some point the developer program is supposed to be getting an overhaul with entirely new documentation. But I don't have a timeline on that, and this doesn't change the core of the article, so I'm going to let it go through. I just want you to know that I probably won't accept many other edits to this article since we're supposed to let it be.

Sharra-writes (approved):

PR #43533, opened 2026-03-25
Fix haproxy-nbthread ghe-config key in release notes

djdefi:
The HAProxy nbthread release note in 3.16.16, 3.17.13, 3.18.7, and 3.19.4 says `ghe-config haproxy-nbthread` but the correct command is `ghe-config core.haproxy-nbthread` (needs the `core.` prefix).

Without the prefix, admins running the command will get an error or set the wrong config key.


djdefi:
Closing — opened in error. Will resubmit through the proper internal channel.

PR #43655, opened 2026-04-03
fix(rest): properly parse body parameters when mixed with oneOf

yassine-bhn:
Fixes an issue where endpoints that declare properties at the top level alongside a oneOf field (like Create a check run) were missing those top-level properties. It also fixes oneOf elements without 'type: object'.
#4365


glitchincorperated-netizen:
All bumb ass can't take loosing energy enough haha keep it up we was just
waiting for # 3 too a world event ...wiping y'all 3rd completely out of
Africa zoning write for rebirth of black babies EVERYONE want the mixed DNA
BOW down is scary strong and LIFE changing 🚼 whaoh

Sharra-writes:
@yassine-bhn Thanks for opening a PR! Unfortunately this isn't a file that open source contributors can alter, since only our documentation is open source, not our workflows and site infrastructure. For that reason I'll have to close this.

PR #43753, opened 2026-04-09
Clarify short title of "Using custom instructions to unlock the power of Copilot code review"

divigrape9:
proposing a short title that better indicate the purpose of this section

### Why: 
Currently the short title of this tutorial is "Use custom instructions"; which doesn't indicate this tutorial is related to Code Review.

This is misleading when navigating through the tutorials from the side navigation:

Image

Closes: https://github.com/github/docs/issues/43752

I'm only proposing a change to the Short Title of this article.

image

### Check off the following:

### DIFF
--- a/content/copilot/tutorials/use-custom-instructions.md
+++ b/content/copilot/tutorials/use-custom-instructions.md
@@ -1,6 +1,6 @@
 ---
 title: Using custom instructions to unlock the power of Copilot code review
-shortTitle: Use custom instructions
+shortTitle: Custom instructions for code review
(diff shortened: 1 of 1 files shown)


Sharra-writes:
@divigrape9 The issue here is that the short title affects the URL, and it may also have a word limit that we can bypass but an open source contributor can't. I'll take your issue to the team and see what they want to do, if anything, but it will probably need to be fixed internally if we change it, since changing it will also affect redirects.

Sharra-writes:
@divigrape9 We are going to update the short title, but it does need to be done internally. We have a PR open for it, so I'm going to close this one. Thanks for flagging it!

PR #43844, opened 2026-04-16
Fixing typo - the docs variable is also populating the 'for example' string

viratsimform:
### Why:
To improve the readability - fixing typo: 'for example' is written two times.
The data variable already contain that string.

### What's being changed:

image

https://docs.github.com/en/copilot/how-tos/copilot-cli/set-up-copilot-cli/configure-copilot-cli#allowing-a-tool-for-the-first-time


Sharra-writes (approved):

Sharra-writes:
@viratsimform Thanks for the fix!

PR #43849, opened 2026-04-16
Clarify that content exclusion requires Copilot Business or Enterprise plan

samueltauil:
## Why

The [Excluding content from GitHub Copilot](https://docs.github.com/en/copilot/how-tos/configure-content-exclusion/exclude-content-from-copilot) how-to page and the [Content exclusion concepts](https://docs.github.com/en/copilot/concepts/context/content-exclusion) page do not explicitly state that repository-level content exclusion settings require the repository to belong to an **organization** with a **Copilot Business or Enterprise** plan.

While the page has a `product` frontmatter gate that renders a banner, the banner is easy to miss. The repository configuration section begins with *"You can use your repository settings…"* without clarifying that the Copilot option will not appear for repositories owned by personal accounts (even with Copilot Pro or Pro+).

This leads to confusion when users follow the instructions and cannot find the settings.

## What

**How-to page** (`exclude-content-from-copilot.md`):
- Adds a `> [!NOTE]` callout in the "Configuring content exclusion for your repository" section explicitly stating the plan and organization ownership requirement, and that personal-account repos will not show the Copilot settings option.

**Concepts page** (`content-exclusion.md`):
- Expands the "Who can configure content exclusion" section to explicitly list the plan requirement (Copilot Business or Enterprise) and clarify that personal-account plans (Copilot Free, Pro, Pro+) are not eligible.

## Validation

- All `{% data variables %}` references verified against `data/variables/copilot.yml`
- Changes follow existing doc conventions (NOTE callouts, variable references)
- No structural or formatting changes to surrounding content

### DIFF
--- a/content/copilot/concepts/context/content-exclusion.md
+++ b/content/copilot/concepts/context/content-exclusion.md
@@ -26,6 +26,8 @@ You can use content exclusion to configure {% data variables.product.prodname_co
 
 Repository administrators, organization owners, and enterprise owners can configure content exclusion.
 
+Content exclusion requires a {% data variables.copilot.copilot_business_short %} or {% data variables.copilot.copilot_enterprise_short %} plan. This feature is not available for {% data variables.product.prodname_copilot_short %} plans assigned to personal accounts (such as {% data variables.copilot.copilot_free_short %}, {% data variables.copilot.copilot_pro_short %}, or {% data variables.copilot.copilot_pro_plus_short %}). Repositories must belong to an organization with an eligible plan for the content exclusion settings to appear.
+
 {% data reusables.copilot.content-exclusions-scope %}
 
 ### Availability of content exclusion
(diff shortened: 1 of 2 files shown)


Sharra-writes:
@samueltauil Thanks for opening a PR! I see a couple of issues here. One, the formatting you're using in "Excluding content," with two notes next to each other, is explicitly against our style guide. Two, "Excluding content" is a how-to article and shouldn't be dealing with conceptual material in any depth. The box at the top really should be enough to give the reader the information they need, except that particular box starts with a long list and doesn't get to your point until the end. So on this article, we may want to do something with paring down or rearranging that box instead of adding a note.

In "Content exclusion," on the other hand, I'm inclined to err on the side of accepting your expanded explanation, because conceptual articles should be thorough. I'll just need to verify the accuracy, and make sure everything is working as intended.

I'll ask for the verification from the Copilot team for "Content exclusion," and check with the rest of the docs team to see what we can do about cleaning up the box at the top of "Excluding content."

PR #44105, opened 2026-05-06
Update "Using conditions to control job execution" Github Actions documentation

mcummings128:
Clarify usage of conditions in job execution and note limitations for shell conditions.

### Why:

Closes: TBD

The Github Actions documentation page "Using conditions to control job execution
" (see link below)

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions

Adds a small blurb that shell conditionals are not supported/evaluated in `jobs.<job_id>.if`. While immediately obvious if trying to execute a workflow that does this, this should be stated explicitly. The phrase "You can use any supported context and expression to create a conditional." is not restrictive enough.

### Check off the following:


Sharra-writes:
@mcummings128 Checking in with Actions about whether they think this is necessary or want to change any wording.

Sharra-writes:
@mcummings128 The Actions team doesn't want the documentation to discuss what *can't* be done in this article, but they did suggest linking to [this article on expressions](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions) to better describe what *can* be done. I don't know if you want to do that in this PR, but it's an option if you don't mind changing what you've got.

specialpreecha-gif:
mcummings128:patch-6

PR #44318, opened 2026-05-18
Add hyperlink to 'New SSH key' in instructions

mcfedr:
why are there no links in the docs? its so annoying


Sharra-writes:
@mcfedr I'm checking in about what Copilot is talking about with versioning. It may be why we don't have a link if it would need to be a different link depending on the version of the article.

mcfedr:
@Sharra-writes its clearly not a very good PR to be honest - its probably somewhere in that include block is where to put the link - but its just super frustrating - and its throughtout the docs - no links to the actual pages - and its particurly because if you google 'add ssh key to github' or similar, you always end up on the docs pages, but its not where you wanna be

Sharra-writes:
@mcfedr The rest of the team says we're not going to do this one, both because of the versioning problem and because our research says that too many inline links hurt the average user experience. That's why we tend to avoid them.

PR #44540, opened 2026-06-01
Reconcile contradictory Search code rate limit on the search REST page

### LINKED ISSUE #43865: REST API docs inconsistency: "About search" says 9 requests/min, "Search code" endpoint section says 10 requests/min — same page, contradictory values
### Code of Conduct

### Description

The GitHub REST API documentation page for search contains two contradictory statements about the rate limit for search requests on the same page:

**Statement 1** — In the "About search" general section:
> Unauthenticated requests are limited to **9 requests per minute**

**Statement 2** — In the "Search code" endpoint-specific section:
> Rate limit: **10 requests per minute**

These two values appear on the same documentation page (`/en/rest/search/search`) and directly contradict each other. Developers and AI agents reading this page cannot determine which rate limit to respect for `GET /search/code`.

### Steps to Reproduce

1. Navigate to: https://docs.github.com/en/rest/search/search#search-code
2. Find the "About search" section near the top of the page
3. Note the rate limit stated there (9 requests per minute)
4. Scroll to the "Search code" endpoint section
5. Note the rate limit stated there (10 requests per minute)
6. Observe that both values appear on the same page and contradict each other

### Expected Behavior

A single, consistent rate limit value for `GET /search/code` stated throughout the documentation page. If the rate limit differs between authenticated and unauthenticated requests, this should be clearly explained in both sections with the same values.

### Actual Behavior

Two different values (9/min and 10/min) appear on the same page with no explanation of the discrepancy. It is unclear which is correct, whether one applies to authenticated vs unauthenticated requests, or whether one section is outdated.

### Additional Context

- For comparison, all other search endpoints are documented at 30 requests/minute — `GET /search/code` is the only endpoint with a lower limit, making this inconsistency more impactful for developers who rely on the documented value to throttle their requests
- Page URL: https://docs.github.com/en/rest/search/search#search-code

he "About search" general section and the "Search code" endpoint-specific section both state the rate limit for GET /search/code, but with contradictory values (9 req/min vs 10 req/min respectively).

One of the following changes is needed:

Update the "About search" section to match the endpoint-specific value (10/min), OR
Update the "Search code" endpoint section to match the general section value (9/min), OR
If the difference is intentional (e.g., authenticated vs unauthenticated), add a clear explanation in both sections so the distinction is explicit rather than contradictory.

### Additional information

_No response_

viraatdas:
Fixes #43865.

The REST search docs page stated two different rate limits for `GET /search/code` on the same page. In the hand-written "About search" prose, the Search code endpoint was described as limited to 9 requests per minute, while the auto-generated "Search code" endpoint section (generated from the OpenAPI description) lists 10 requests per minute.

This aligns the hand-written prose to the auto-generated endpoint value (10 requests per minute), which is the source of truth, so the page is internally consistent. Only the prose above the autogeneration marker is changed; no generated content is touched, and no other numbers are altered.

Prepared with AI assistance.

### DIFF
--- a/content/rest/search/search.md
+++ b/content/rest/search/search.md
@@ -31,7 +31,7 @@ Unless another sort option is provided as a query parameter, results are sorted 
 {% data reusables.enterprise.rate_limit %}
 
 The REST API has a custom rate limit for searching. For authenticated requests, you can make up to
-30 requests per minute{% ifversion fpt or ghec %} for all search endpoints except for the [Search code](/rest/search/search#search-code) endpoint. The [Search code](/rest/search/search#search-code) endpoint requires you to authenticate and limits you to 9 requests per minute{% endif %}. For unauthenticated requests, the rate limit allows you to make up to 10 requests per minute.
+30 requests per minute{% ifversion fpt or ghec %} for all search endpoints except for the [Search code](/rest/search/search#search-code) endpoint. The [Search code](/rest/search/search#search-code) endpoint requires you to authenticate and limits you to 10 requests per minute{% endif %}. For unauthenticated requests, the rate limit allows you to make up to 10 requests per minute.
 
 For information about how to determine your current rate limit status, see [Rate Limit](/rest/rate-limit/rate-limit).
 


jobayerbt (commented):
#44545

Sharra-writes:
@viraatdas I don't have answers on this yet, so this PR will have to wait until I know which one is supposed to be correct and whether we're doing it internally or whether it's available for open source correction.

Sharra-writes:
@viraatdas The engineers got back to me, and the 9/minute is actually correct, so I'm going to close this PR. The portion that needs to be changed is auto-generated, and so has to be fixed internally.

PR #44596, opened 2026-06-04
Clarify "do not create an organization" warning for Copilot Business standalone setup

mattstana:
## Why

Customers setting up a dedicated enterprise account for Copilot Business standalone (no GitHub Enterprise) are repeatedly creating an organization during the trial flow because the warning in the current article is a single, unbolded paragraph buried under the trial button. This results in surprise GitHub Enterprise license assignments and billing, which then has to be walked back via Support.

## What changed

1. Promotes the "do not create an organization" warning to a proper `[!WARNING]` callout so it visually stands out from the surrounding setup instructions.
2. Adds explicit consequence language ("will bill your account for GitHub Enterprise - which you do not need for Copilot Business") so readers understand the cost impact, not just the policy.
3. States the goal-state explicitly: **0 seats of GitHub Enterprise** when setup is complete. This matches language Sales uses with customers and gives readers a self-check.
4. Tells the reader what to do instead (add users directly to the enterprise).
5. Adds one line noting this path is correct for ADO-only / non-GitHub-hosted code customers, which is the most common Copilot Business standalone scenario from the field.

## Signal from the field

Multiple Digital Sales reps and SDRs have flagged this in internal channels when customers ask why they are being billed for GitHub Enterprise they did not intend to buy. There is an internal text expander maintained by Sales that calls this step out as its own bold heading - this PR brings the public docs closer to that internal guidance.

## Preview

Before: one short paragraph easily scrolled past.
After: a `[!WARNING]` callout with bolded directive, consequence, goal-state, and the corrective action.


mattstana:
Closing - I should have opened this in the private `github/docs-internal` repo per Hubber contribution guidance, not here. No sensitive info was disclosed (the change only touches an already-public page). Reopened internally as github/docs-internal#61582.

PR #44599, opened 2026-06-04
Clarify Dependabot is exempt from IP allow list enforcement

emisanada:
## Summary

Updates the Dependabot IP allow list documentation to accurately reflect that Dependabot is a first-party GitHub App whose repository access is **exempt from IP allow list restrictions**.

## Why

The current docs state that customers "must set up a self-hosted runner or enable Dependabot for use with larger runners" when using IP allow lists. This is inaccurate for Dependabot's core operations:

- Dependabot is a privileged first-party app with explicit `ip_allowlist_exempt: true` capability
- Its repository access (reading dependency files, creating PRs) bypasses IP allow list enforcement by design
- Customers have observed this working and are confused because the docs say otherwise ([internal ref](https://github.com/github/enterprise-primitives/issues/5258))

## Changes

Rewrites `data/reusables/dependabot/ip-allow-list-dependabot.md` to:

1. **State clearly** that Dependabot's repository access is exempt from IP allow lists
2. **Remove misleading "must" language** about requiring self-hosted/larger runners for basic Dependabot functionality
3. **Keep runner guidance** for other use cases where static IPs are needed (e.g., accessing private package registries behind firewalls)

## What this does NOT cover

The interaction between `GITHUB_TOKEN` in Dependabot workflow steps and IP allow list enforcement is nuanced and not fully documented here. The Actions app has a different exemption scope (`ip_allowlist_exempt_for_internal_apis` only). This PR focuses solely on clarifying Dependabot's own access, which is unambiguously exempt.

## Affected pages

This reusable appears on:
- [Restricting network traffic to your enterprise with an IP allow list](https://docs.github.com/en/enterprise-cloud@latest/admin/configuring-settings/hardening-security-for-your-enterprise/restricting-network-traffic-to-your-enterprise-with-an-ip-allow-list#using-dependabot-with-an-ip-allow-list)


emisanada:
Closing — moving to docs-internal per DIY docs guidelines.

PR #44778, opened 2026-06-17
docs: clarify workflow_dispatch environment input

### LINKED ISSUE #44770: inputs type are undocumented / contradicting
### Code of Conduct

### What article on docs.github.com is affected? / What part(s) of the article would you like to see updated?

- workflow-syntax#onworkflow_dispatchinputsinput_idtype
  - https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputsinput_idtype
  - https://github.com/github/docs/blob/fa4d1bcf57215fe0eae198c42d3b83f8679402a1/content/actions/reference/workflows-and-actions/workflow-syntax.md?plain=1#L264-L266
- events-that-trigger-workflows#providing-inputs
  - https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#providing-inputs
  - https://github.com/github/docs/blob/fa4d1bcf57215fe0eae198c42d3b83f8679402a1/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md?plain=1#L1076-L1130
- contexts#inputs-context
  - https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#inputs-context
  - https://github.com/github/docs/blob/fa4d1bcf57215fe0eae198c42d3b83f8679402a1/content/actions/reference/workflows-and-actions/workflow-syntax.md?plain=1#L162-L164

### Additional information

Input types are undocumented / contradicting

Particularly "environment", which, as per the documentation, seems to be a special sub-type of choice automatically populating the enum with the names of the repository's environment, but this is just a guess from the only screenshot available in the documentation

Given this, there might also even be other types

ketpatil77:
Fixes #44770. Clarifies the special environment input type in GitHub Actions docs.

### DIFF
--- a/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md
+++ b/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md
@@ -1081,6 +1081,8 @@ You can configure custom-defined input properties, default input values, and req
 
 This example defines inputs called `logLevel`, `tags`, and `environment`. You pass values for these inputs to the workflow when you run it. This workflow then prints the values to the log, using the `inputs.logLevel`, `inputs.tags`, and `inputs.environment` context properties.
 
+When you use `type: environment`, GitHub shows the workflow's environment names as the available values for that input.
+
 ```yaml
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


Sharra-writes:
@ketpatil77 Hey, there! 👋 Generally we ask people not to open PRs like this until we've gotten input from the relevant team(s). I don't know that the Actions team will want to update the docs at all, let alone that they'll want this wording specifically. I'm going to close this, because otherwise I'm afraid it will get lost in the shuffle. If/when we're ready to update the docs, a `help wanted` label will be put on the issue in the question, and there will be guidance on how we would like to see the docs updated.

PR #44816, opened 2026-06-18
Add duplicate detection tip to 'Creating an issue' docs

sunnyi101:
## What this PR does

Adds a tip callout to the [Creating an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/creating-an-issue) page explaining the duplicate detection feature.

When creating an issue, GitHub now surfaces potential duplicate issues inline once the title is filled out and the body reaches 100 characters. Up to three existing issues may be suggested. The suggestions are non-blocking and don't prevent issue submission.

## Why

This feature is shipping to GA and the docs should reflect the new behavior so users understand what the suggestions are and when they appear.

## Context

- Staff ship post: https://github.com/github/planning-tracking/discussions/2780
- Docs content issue: https://github.com/github/docs-content/issues/23073


sunnyi101 (commented):
pleaser review

Sharra-writes (approved):

PR #44934, opened 2026-06-27
[Open Source 2.0] Glossary - Update open source - change 'working materials' to 'source files'

AbcSxyZ:
[*Following the article « [Open source, not just software anymore](https://ben.balter.com/2014/01/27/open-collaboration/) » by @benbalter and <https://github.com/benbalter/benbalter.github.com/pull/98> which leads to the current state of the definition in your glossary.*]

### Why:

Hello, I propose an update to the definition of open source in the glossary, replacing 'working materials' to 'source files'. This is related to a citizen research I did on the meaning of open source, published in the preprint « *[Open Source 2.0: From Open Source Software to Open Source Resources?](https://doi.org/10.5281/zenodo.20237079)* ».

It supports the idea that open source is slowly derivating from source files in other openness movements (such as « [Open Source Educational Resources](https://arxiv.org/abs/2107.14330) » in extension to OER in open education).

This update could bring more consistency to the existing definition while being based on research.

> **Open Source Resource:** a digital resource made publicly available alongside its source files.

### DIFF
--- a/data/glossaries/external.yml
+++ b/data/glossaries/external.yml
@@ -473,7 +473,7 @@
     Open source software is software that can be freely used, modified, and
     shared (in both modified and unmodified form) by anyone. Today the concept
     of "open source" is often extended beyond software, to represent a
-    philosophy of collaboration in which working materials are made available
+    philosophy of collaboration in which source files are made available
     online for anyone to fork, modify, discuss, and contribute to.
 - term: organization
   description: >-
--- a/data/glossaries/external.yml
+++ b/data/glossaries/external.yml
@@ -473,7 +473,7 @@
     Open source software is software that can be freely used, modified, and
     shared (in both modified and unmodified form) by anyone. Today the concept
     of "open source" is often extended beyond software, to represent a
-    philosophy of collaboration in which source files are made available
+    philosophy of collaboration in which source files of digital resources are made available
     online for anyone to fork, modify, discuss, and contribute to.
 - term: organization
   description: >-


Sharra-writes:
@AbcSxyZ I don't think we're going to do this for two reasons. One, this glossary is mostly for internal use and training on the terms we use a lot in everyday discussions. Two, "source files of digital resources" repeats the word "source" (open source, source files) and then "resources" also has "source" as a root word, potentially making the entire thing feel self-referential. It may be more specific and accurate from a research standpoint, but from a writing standpoint it starts to feel like it's going in circles, like this: "What is open source? It's openly sourced resources!" Not quite that bad, but getting there. "Working materials" references ideas outside of what's already been established by the term "open source," which is generally what you want when trying to help someone grasp a concept. Sorry we're not going to accept it, but thank you for the interesting research journey.

AbcSxyZ:
Hello @Sharra-writes, thanks for the reply!

Regarding internal use, how the biggest open source plateform define open source may be more significant than that. I used it in my research regarding the definition of open source so, I refer to it in many comments I do like when we debated to update the glossary of the turing way.

I'm quite confused with your explanations around working materials. Working material is something included in the scope of your open source definition, not outside it. As a result of @benbalter consideration for source files of his blog post when he wrote the defintion of open source here. Introducing working materials to explain availability of source code would be quite confusing tough.

From a writing standpoint, I agree it's a bit repetitive, and it would be awesome if people naturally consider source files with resources but it is definitely not the case... This linguistic distinction have direct technical implications, this might require other forms of consideration.

Your current definition is confused, and this can impact directly the training of your staff, this source files dimension may bring some clarity. And I guess it's not refered that much to explain open source when it's such a common knowledge, that's why you probably explain the concept as software while having a definition that goes beyond software.

The organization managing the largest open source hosting platform is inconsistent regarding the meaning of open source, just my 2 cents.

I may come back one day when(/if) the open source resources paradigm become more widespread!

PR #44992, opened 2026-06-30
docs(actions): use GiB for runner capacity tables

### LINKED ISSUE #44033: Use GiB instead of GB where binary units are intended
### Code of Conduct

Several pages use decimal units ("GB") when referring to memory or storage capacities that are measured in binary units. For technical accuracy, consistency, and to avoid ambiguity, these should use "GiB" instead of "GB" where the values represent gibibytes.

Example:

- https://github.com/github/docs/blob/5fd4cfa36b16e395961364113d1080a125e4e5a0/data/reusables/actions/supported-github-runners.md?plain=1#L22

This may also affect other pages that list runner specifications or storage capacities.

General documentation consistency update.

### Additional information

This appears to be a broader terminology and consistency issue across GitHub Docs.

happysnaker:
## Summary
- use `GiB` instead of `GB` in GitHub-hosted runner capacity tables where the values describe binary-sized memory and storage resources
- update the shared reusable tables so the terminology stays consistent across the runner reference docs
- align the larger runners reference with the same unit terminology

Closes #44033

### DIFF
--- a/content/actions/reference/runners/larger-runners.md
+++ b/content/actions/reference/runners/larger-runners.md
@@ -29,16 +29,16 @@ You can choose from several specifications for {% data variables.actions.hosted_
 
 | CPU | Memory (RAM)  | Storage (SSD) | Architecture | Operating system (OS) |
 | --- | ------------- | ------------- | ------------ | --------------------- |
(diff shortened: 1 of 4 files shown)


Sharra-writes:
@happysnaker Again, no help wanted label. If I continue to find these, I will block this account.

PR #45110, opened 2026-07-11
Typo: config.yml not _config.yml

ferdnyc:
### Why:
In the project health docs, #35289 changed a `_config.yml_` (italics) to be an inline fenced literal instead (`config.yml`), for the issue template config file name. But the leading underscore was inadvertently left in, making it `_config.yml` which is incorrect.

### Check off the following:


Sharra-writes:
@ferdnyc Looks like this is probably accurate. Don't worry about the checks that aren't passing - it's a workflow thing on our end. Hopefully it will be fixed soon.

Sharra-writes (approved):

PR #45208, opened 2026-07-19
Note that hyphens are replaced as well as spaces in seeding of `INPUT_*` environment variables in `action.yaml` reference

jonathandung:
### Check off the following:

### DIFF
--- a/content/actions/reference/workflows-and-actions/metadata-syntax.md
+++ b/content/actions/reference/workflows-and-actions/metadata-syntax.md
@@ -59,7 +59,7 @@ inputs:
     required: true
 ```
 
-When you specify an input, {% data variables.product.prodname_dotcom %} creates an environment variable for the input with the name `INPUT_<VARIABLE_NAME>`. The environment variable created converts input names to uppercase letters and replaces spaces with `_` characters.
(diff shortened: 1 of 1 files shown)


jonathandung:
Huh? I don't recall having used Copilot at all.

eddieriley85-jpg:
jonathandung:patch-1

jonathandung:
I just read the [source](https://github.com/actions/toolkit/blob/e7728b1bcda3082eca6b716f0c6e20c743b7972d/packages/core/src/core.ts#L151), and it appears that hyphens are not replaced (I might be wrong because I'm not familiar with JS). If that is indeed so, it should be made explicit.

Sharra-writes:
@jonathandung I'm not sure we need to document something that doesn't change, but I'm looking into it.

Sharra-writes:
@jonathandung Sorry, we don't really feel the need to document something that doesn't change. "Nothing else changes" is kind of implied by "spaces become underscores."

jonathandung:
But the example keys contain hyphens (`num-octocats` and `octocat-eye-color`, and I don't think environment variables can have hyphens. So how is it that hyphens are kept intact?

PR #45394, opened 2026-08-03
Modern Egyptian Arabic translation

Said12356:


Sharra-writes:
@Said12356 Unfortunately we don't support Arabic translations at this time.

PR #45432, opened 2026-08-07
docs: update code scanning tools label

breezeFur:
### Why:

Closes github/technical-content#7335

### What's being changed:

The code scanning tools banner now uses **Tools** instead of **Tool status**. This PR updates the instruction text, the image alternative text, and the screenshot so the procedure matches the current interface.

### Validation:

- `npm run lint-content -- --paths content/code-security/how-tos/find-and-fix-code-vulnerabilities/manage-your-configuration/use-the-tools-status-page-for-code-scanning.md` passed.
- `git diff --check` passed.
- The requested changed-content Vitest could not start in this Windows checkout because the repository's alias configuration resolves `@/frame/start-server` through the original non-ASCII workspace path. No test assertion ran.

### Check off the following:

- [ ] A subject matter expert has reviewed the technical accuracy of the content.
- [x] The changed content is limited to the terminology reported in the linked issue and its matching screenshot.
- [ ] All CI checks are passing.


subatoi:
Hi @breezeFur, I actually responded to @cpovirk in the original issue about this same thing, and transferred it internally. I've since realised that our automation isn't working as intended right now, so @cpovirk wouldn't have received my notification about this (hence the ping here 🙇 )

You're both correct that this needs to be fixed, but it's not trivial to do so, and more needs to be fixed than just what you've proposed here, so we'll fix it internally. I can't guarantee a timeframe but I'll mark this with the appropriate label and close it out for now, but leave my comment here visible. Thank you for your interest in our docs!

PR #45605, opened 2026-08-24
Add permissions to workflow_run example

### LINKED ISSUE #45520: workflow_run "Use the data" example omits the permissions the example itself needs
### Code of Conduct

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#using-data-from-the-triggering-workflow

The `Use the data` example in the `workflow_run` section.

The example has no `permissions:` block, and both of its steps need a permission that the restricted default does not grant.

The workflow downloads an artifact from a different workflow run, passing `run-id` and `secrets.GITHUB_TOKEN`, and then calls `github.rest.issues.createComment`. Your own reference pages say what those two calls require. Downloading an artifact is documented as needing `Actions: read`, and creating an issue comment is documented as accepting `Issues: write` or `Pull requests: write`. The restricted repository default grants read access to `contents` and `packages` only, and the workflow permissions reference says the effective token starts from that default and changes only through a `permissions:` key.

So a reader who copies this example into a repository on the restricted default gets a workflow that cannot download the artifact and cannot post the comment. On a permissive default the same example works, which is why the gap is easy to miss.

Adding the block to the example would fix it:

```yaml
permissions:
  actions: read
  issues: write
```

`pull-requests: write` is the documented alternative to `issues: write` here, not a third permission.

This matters more than a typo because the surrounding section is about handling a potentially untrusted triggering workflow, and the example already carries a security comment about not extracting into the workspace directory. An example in that context is the one readers are most likely to copy verbatim, and least likely to widen by hand.

### Additional information

I read the source on main today, 17 August 2026, at `content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md`. The example there has no `permissions:` key.

I checked six permission combinations against the two documented requirements. The current example fails on a restricted default and passes on a permissive one. The example with `actions: read` and `issues: write` passes on a restricted default. Either permission alone fails. The `pull-requests: write` variant passes. This is a reading of your documented requirements and not a live Actions run, so I am not reporting an observed error message.

I searched open and closed issues first. There are reports about `GITHUB_TOKEN` permissions in general, but none about this example.

The pages I read for the requirements:

- [How permissions are calculated for a workflow job](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#how-permissions-are-calculated-for-a-workflow-job)
- [Setting GITHUB_TOKEN permissions for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository)
- [Download an artifact](https://docs.github.com/en/rest/actions/artifacts#download-an-artifact)
- [Create an issue comment](https://docs.github.com/en/rest/issues/comments#create-an-issue-comment)

The GitHub Security Lab article on preventing pwn requests uses the same two-workflow pattern and declares `actions: read` and `pull-requests: write` explicitly, which is the shape I am suggesting here.

Three things would make this report wrong, and I would rather name them than have you find them. If a `workflow_run` job is guaranteed `Actions: read` and `Issues: write` regardless of a restricted effective default, there is nothing to fix. If the cross-run download path in this exact example does not need `Actions: read` when given `GITHUB_TOKEN`, drop that line from the suggestion. And if the restricted default grants more than `contents` and `packages` in the relevant product context, the portability argument goes away.

Arizetbest:
### Why:

Closes: #45520

The `workflow_run` "Use the data" example downloads an artifact and creates a pull request comment using `GITHUB_TOKEN`, but the example does not explicitly declare the permissions required for these operations.

Explicitly declaring the required permissions makes the example work with repositories that use restrictive default `GITHUB_TOKEN` permissions and follows the principle of least privilege.

### What's being changed:

This PR adds explicit job-level permissions to the `download` job in the `workflow_run` example:

```yaml
permissions:
  actions: read
  issues: write
```

- `actions: read` allows the workflow to access and download the artifact from the triggering workflow.
- `issues: write` allows the workflow to create the pull request comment.

No other behavior in the example is changed.

### Check off the following:

- [x] The changes in this PR meet the docs fundamentals that are required for all content.

### DIFF
--- a/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md
+++ b/content/actions/reference/workflows-and-actions/events-that-trigger-workflows.md
@@ -1248,6 +1248,9 @@ on:
 jobs:
   download:
     runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      issues: write
     steps:
       - name: 'Download artifact'
         uses: {% data reusables.actions.action-download-artifact %}


alazaouibay2004-sudo:
Arizetbest:Arizetbest-patch-1

alazaouibay2004-sudo:
مرحبا بيك في الجزائر

subatoi:
Hello, and thanks for raising this—

As I noted in https://github.com/github/docs/issues/45520#issuecomment-5407416733, whilst this does fix the problem, the issue in question wasn't already approved for a community contribution, so I'd like to give the OP the chance to raise a PR if they wish to. Otherwise, we'll go ahead and merge this. Thanks 🙇

men1986:
9950277

subatoi:
As the OP has noted [here](https://github.com/github/docs/issues/45520#issuecomment-5407875376), they would prefer to raise a PR for this themselves.

Thank you for your interest in the GitHub docs, but I'll go ahead and close this out—please feel free to keep an eye on issues that have been approved for community contributions (with a `help wanted` label) and we hope to see you in future 🙇

PR #45682, opened 2026-08-31
Update `securely-using-pull_request_target` to remove `actions/checkout` version restriction

JackPGreen:
_Originally_ the change applied to `actions/checkout` `v7` only, but it's [since been backported to older versions](https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/) without the docs being updated.


subatoi (approved):
Thank you for spotting that and raising a fix, @JackPGreen—I'll get this merged now

PR #45768, opened 2026-09-06
Update documentation link for GitHub-hosted runners

mcummings128:
Updated link to documentation for GitHub-hosted runners. The current link points to https://docs.github.com/en/enterprise-server@3.20/actions/reference/runners/github-hosted-runners#single-cpu-runners, when it should point to https://docs.github.com/en/actions/reference/runners/github-hosted-runners

### Why:

Closes: 

### Check off the following:


subatoi:
@mcummings128 Hello, I'll tighten up the anti-spam workflows but in this instance, I'm not sure why this was raised like this—if you've spotted something like this you'd suggest needs changing, I'd recommend opening an issue. Thank you

PR #45807, opened 2026-09-08
Copilot: Fix Pro+ signup attribution in setup guide

sbaidon:
## Summary

- correct the Copilot Pro+ signup link in the self-setup guide
- use `ref_plan=pro-plus` instead of attributing Pro+ clicks to `pro`

## Why

The Pro and Pro+ links currently send the same `ref_plan=pro` value. This can misattribute Pro+ intent and may select the wrong downstream plan context.

## Validation

- confirmed the branch changes only the Pro+ tracking parameter
- confirmed the corrected URL preserves `ref_plan=pro-plus` through the signed-out authentication redirect
- confirmed the Pro and Max links are unchanged


sbaidon:
Closing this PR and moving the change through the Hubber contribution workflow in the private Docs repository. The proposed one-line correction remains valid; only the contribution path is changing.

PR #45907, opened 2026-09-15
Correct language tags on five non-YAML code examples

SurefireStudios:
### Why:

Closes:

Five code blocks in [Remove access to public registries](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/remove-access-to-public-registries) are tagged `yaml copy` but contain no YAML, so the highlighter applies the wrong grammar.

| Block | Actual format |
| --- | --- |
| Yarn "Option 1" (x2) | `yarn.lock` lockfile |
| Pip "Option 1" | `pip.conf` (INI) |
| Pipenv "Option 1" | `Pipfile` (TOML) |
| Poetry "Option 1" | `pyproject.toml` (TOML) |

### What's being changed (if available, include any code snippets, screenshots, or gifs):

Retagged those five blocks from `yaml copy` to `text copy`, keeping the `copy` attribute.

I picked `text` because it is already used ~400 times in `content/` while `toml` and `ini` appear nowhere — happy to switch to `toml` for the Pipfile and pyproject.toml blocks if you prefer.

The seven genuine `dependabot.yml` examples in this article are untouched and still parse as YAML.

### Check off the following:

- [x] A subject matter expert (SME) has reviewed the technical accuracy of the content in this PR. In most cases, the author can be the SME. Open source contributions may require an SME review from GitHub staff.
- [x] The changes in this PR meet [the docs fundamentals that are required for all content](http://docs.github.com/en/contributing/writing-for-github-docs/about-githubs-documentation-fundamentals).
- [x] All CI checks are passing and the changes look good in the review environment.

PR #45908, opened 2026-09-15
Fix invalid <clear /> element in nuget.config example

SurefireStudios:
### Why:

Closes:

[Remove access to public registries](https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/remove-access-to-public-registries) writes the NuGet clear element as `< clear />`, and XML element names cannot begin with whitespace.

So the example does not parse, and NuGet rejects the `nuget.config` the article is telling readers to write.

The correct `<clear />` is already used twice in [Working with the NuGet registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry), and `< clear />` appears nowhere else in `content/`.

Removed the stray space in all three occurrences — the `xml` block plus the two prose mentions, which are wrong the same way.

The block now parses:

```
root = configuration
  child: clear {}
  child: add {'key': 'example-nuget', 'value': 'https://private_registry_url/nuget/example-nuget/index.json'}
```

Same article as #45907 but non-overlapping lines, so either can merge first.

### Check off the following:


janiceilene:
👋 @SurefireStudios thanks for the PR! I'll get this triaged for review.

mchammer01 (approved):
Nice catch! 🎯 
I appreciate that you tracked down all three occurrences.
Thanks for contributing! LGTM ✅

PR #45983, opened 2026-09-21
Use jsonc language tag for JSON examples containing comments

Prajwal-chikkur07:
### Why:

Closes:

There's no existing issue for this; it's a small technical correction to code examples.

Three code blocks in the Copilot docs are tagged with the `json` language, but their contents include comments. JSON has no comment syntax, so these examples fail to parse and can't be copied and used as-is.

The content directory already has an established convention for this case: JSON examples that carry explanatory comments are tagged `jsonc`. See, for example, the dev container examples under `content/codespaces/.../adding-a-dev-container-configuration/`.

That convention is also documented in the renderer. `src/content-render/unified/processor.ts` registers `jsonc` as an alias of `json` for syntax highlighting, with this comment:

> So by sticking to 'jsonc' where there's JSON with comments, it's technically more correct, looks good in github.com, but with this alias you get the nice syntax highlighting when viewed on our site.

So retagging these three fences makes them consistent with both the surrounding content and the stated intent, with no change to how they're highlighted on docs.github.com.

Changed the language tag from `json` to `jsonc` on three code fences. **The example content itself is unchanged** — only the three fence tags differ.

| Article | Comment in the example |
| --- | --- |
| `copilot/reference/hooks-reference.md` | `"preToolUse": [ /* hook entries */ ]` |
| `copilot/how-tos/copilot-sdk/troubleshooting/mcp-debugging.md` | `// In initialize response, check protocolVersion` |
| `copilot/reference/copilot-usage-metrics/lines-of-code-metrics.md` | three trailing `// ...` comments on `loc_*` fields |

For example, in `hooks-reference.md`:

```diff
-```json
+```jsonc
 {
   "version": 1,
   "disableAllHooks": false,
   "hooks": {
     "preToolUse": [ /* hook entries */ ]
   }
 }
```

### Check off the following:


subatoi:
This won't make any material difference, as `json` is already aliased to `jsonc`, which is noted in the text that the LLM has picked up on. We therefore won't be accepting this change

